PR #2, opened 2021-02-25
add boilerplate for the frontend

stefanotorresi:
This PR adds some preliminary boilerplate for the frontend stack, which mostly consist for now in the EOS design system built with SASS.

I still need to find a way to keep the easy installation method with `go get -u github.com/SUSE/console-for-sap`, which will be broken by this PR, due to the fact that we need to also build the frontend with SASS after having installed npm dependencies, all of which is now performed with `make`.


arbulu89 (dismissed):
I guess it is fine. I don't know still know how we are going to attach this assets serving with a potential js framework as vue in the future, but I guess you already thought on that

stefanotorresi:
We're just going to have to build the front-end assets into the directory that we embed in the backend binary. I just added a Makefile target exactly for that purpose. Only I forgot to declare it as a dependency of the `test` target, so ATM the CI is failing. I'm gonna fox that later.

arbulu89 (approved):
It looks good in all of my ignorance hehe

stefanotorresi:
@arbulu89 I know that feeling, but please, don't fret in asking me if there is something you don't understand about the code you're looking at ;)

PR #3, opened 2021-03-04
Add EOS branding layout usage

arbulu89:
EOS branding usage implementation. This includes:

- EOS styles and structure
- Layout usage implementation in `layout.go`. This module gives the option to add layout templates. It is based in the `multitemplate` module, but as it doesn't have the `parseFS` usage, I have had to create a new one based on that one. The drawback is that all the added new templates will use the layout by default, but we could easily improve this adding some new method to don't apply the layout to specific html rendering systems
**The most important part is the `layout.go` part. Have a look there to see if this design option is good**

The home endpoint is hardcoded by now. We will make this customized in the future

image
image

**Edit**. Full name of `SUSE Console for SAP Applications` added in the sidebar title (widening the sidebar a bit):
image


Makefile
```diff
@@ -29,7 +29,35 @@ webapp-deps: webapp/frontend/node_modules
 webapp/frontend/node_modules:
 	cd webapp/frontend; npm install
 
-webapp-assets: webapp/frontend/assets/stylesheets.css
-webapp/frontend/assets/stylesheets.css: webapp/frontend/node_modules
+webapp-assets: webapp-images webapp-assets-stylesheets webapp-assets-javascrips
```
MalloZup, line 32: there is a typo  `webapp-assets-javascrips`

Makefile
```diff
@@ -29,7 +29,35 @@ webapp-deps: webapp/frontend/node_modules
...
+	mkdir -p webapp/frontend/assets/js
+	cp webapp/frontend/javascripts/layout.js webapp/frontend/assets/js/layout.js
+
+javascripts-eos-ds: webapp/frontend/node_modules
+	mkdir -p webapp/frontend/assets/js/eos-ds
+	cp webapp/frontend/node_modules/eos-ds/dist/js/index.js webapp/frontend/assets/js/eos-ds/index.js
+
+webapp-images:
+	mkdir -p webapp/frontend/assets/images
+	cp -R webapp/frontend/images webapp/frontend/assets
```
MalloZup, line 63: I do think here we are introducing way to much Makefile `targets`.

I do have used Makefile in many projects, having to much targets make the logic complex and difficult to maintain.

we should strive to more simplicity and less targets here, imho

stefanotorresi: I would suggest to find a middle ground here.

I would still have a few dedicated concrete targets for the main directories inside `webapp/frontend/assets`, but only one phony target. 

I.e. I would keep the `webapp-assets` target as the only phony, which I would then point to the `webapp/frontend/assets` concrete target (since we're now cleaning that directory entirely), so that `make` can track its existence; this concrete target would then compiles all the files we need to embed in the `webapp.assetsFS` variable by invoking concrete sub targets for the `webapp/frontend/assets/js`, `webapp/frontend/assets/stylesheets` and `webapp/frontend/assets/images`  directories; I don't think we need separate sub targets for the eos ds, the layout or the override and icon styles within those directories.

Remember to always have concrete targets eventually, otherwise you're not allowing `make` to track the compiled files.

This is why initially `webapp-assets` just pointed to `webapp/frontend/assets/stylesheets.css`, but the actual compilation 
happened in the latter: `webapp-assets` is just a phony target (which I actually forgot to mark with `.PHONY`!).

arbulu89: Hi guys,
Thank you for the suggestions. I don't have almost any idea about `makefiles`. I just tried to follow the current directives.
We can do 2 things:
- You can suggest me here how we should change it, as I would learn in the process
- We let it be, and we do the changes in other PR, and I learn in the review process

arbulu89: Simplified in https://github.com/SUSE/console-for-sap-applications/pull/3/commits/1777a0440dc996233482285bc1ab1fa589aa0490

stefanotorresi (changes requested):
let's make my review in https://github.com/arbulu89/console-for-sap/pull/1

stefanotorresi (approved):

PR #5, opened 2021-03-09
WIP: Use chi instead of `go gin`

MalloZup:
see https://gist.github.com/MalloZup/aee6845fc3090696642828f7cc345685


dirkmueller (commented):
I just started implement session handling here so would like to discuss this first

stefanotorresi:
@MalloZup I'm not quite clear what would this PR achieve, because it removes the front-end scaffolding we already did. Could you produce an actual refactoring so that it produces the same exact HTML, so that we can the appreciate the difference between the two approaches without changing the end result?

MalloZup:
I didn't had time yet to do it , since I'm preparing/working on the `hawk` update and I did  this PR as demonstrative one and on the fly.

My main point for using `chi` or `mux` is that they aren't framework as `go gin ` is .

So basically we can plug a component like the router, and then we spit the choices later in future if we need  a session or whatever component.

With go gin we buy everything. 

you can see also this reddit thread:

https://www.reddit.com/r/golang/comments/dojyv2/gorillamux_or_gingonicgin/

stefanotorresi:
I appreciate your concerns, OTOH all the remarks you made so far are very generic and random personal thoughts collected from the internet; I don't see many technically compelling arguments, to be frank.

I can only see two reasonable arguments so far:
1. it's provides too many features.
2. its handler function signature is not compatible with Go stdlib `http` package.

About 1: we already knew beforehand that we actually need some of those features, so this is why we might want to buy in.
About 2: I also don't like that at all, but I know there are compatibility wrappers around that are a just a few lines of code away.

At the end of the day, I'm open to change, but it needs to be a proper refactoring matching our current implementation and future intentions, which is not just a small json API but a full fledged web application; this proposed example with single dummy template is not enough to exemplify to what extent the change of framework can help us.

Your initial argument for change seems to have spawned by the addition of the bespoke implementation of the multi-template rendering: that's not going away with `chi`, so what are we solving here exactly?

MalloZup:
Arguing "The tool picked for it is completely secondary" is imho something in software engineering something to cautious of.
And in fact, somehow you say that I pick up some random framework based on random forums,  on a previous post, and  then say the tool is secondary.

To be really objective here, I think that we should have been evaluating at least "minimally" which web-framework we would have pickedup before rushing to an implementation. 

Again `mux` /chi aren't framework, where `go gin` is, with all its logic.

Already the way we implemented  https://github.com/SUSE/console-for-sap-applications/blob/main/webapp/layout.go

this file has introduced already **technical debts*. The way we uses nested html templates is already specific to `go gin` e.g
https://github.com/SUSE/console-for-sap-applications/blob/main/webapp/layout.go#L96

If @dirkmueller  and @arbulu89 might need to see if they  like the way we are doing things  in go gin, it the team likes this approach I will let be.

I personally don't like the way we have implemented it, it is already imho an anti-pattern.

I don't have right now not the time, to dive on rendering thing, because of hawk,

 In fact since we merged `gin`, every things we will merge will increase the difficult to change the tooling/framework.

I personally don't understand the magic of `layout.go` file. IMHO maybe @arbulu89  should try to take look to the PR and see if `mux` + chi or `html/template` is way better. 

but if @dirkmueller or @arbulu89  has strong opinion on `gin` i'm ok. Now, to be honest , despite @arbulu89  has already explained me the `layout.go` logic, I still don't understand, it looks really internally to go gin.

MalloZup:
i'm closing this in favor of new one

stefanotorresi:
follow up is in #7

PR #7, opened 2021-03-12
Use a minimalist router with multiple Html template

MalloZup:
## Description:

- swap go gin with `chi` router
- simplify templates etc


go.mod
```diff
@@ -4,8 +4,9 @@ go 1.16
 
 require (
 	github.com/gin-gonic/gin v1.6.3
+	github.com/go-chi/chi/v5 v5.0.1
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/spf13/cobra v1.1.3
 	github.com/spf13/viper v1.7.1
-	github.com/stretchr/testify v1.7.0
```
dirkmueller, line 10: this is an awkward change, that indicates the line can be removed because we dropped all _test.go files? 

can we at least have one unit test back?

MalloZup: yes thx Dirk.. I have remove the test but I plan to add some unit test later on. thx!

webapp/templates/sidebar.html.tmpl
```diff
@@ -26,10 +26,10 @@
   <footer class="footer-side-menu">
     <ul class="footer-list">
       <li class="footer-list-item">
-        <i class="eos-icons" title="" data-html="true" data-toggle="tooltip" data-title="{{ escapedTemplate "license" . }}" data-trigger="hover click">find_in_page</i>
+        <i class="eos-icons" title="" data-html="true" data-toggle="tooltip" data-title="SUSE" data-trigger="hover click">find_in_page</i>
```
MalloZup, line 29: this is a shortcut i took, because of the `escapedTemplate` caused panic

webapp/templates/sidebar.html.tmpl
```diff
@@ -26,10 +26,10 @@
...
       </li>
       <li class="footer-list-item">
-        <i class="eos-icons" title="" data-html="true" data-toggle="tooltip" data-title="{{ .copyright }}" data-trigger="hover click">info</i>
+        <i class="eos-icons" title="" data-html="true" data-toggle="tooltip" data-title="2020" data-trigger="hover click">info</i>
```
MalloZup, line 32: another shortcut which I might need to take over next week. I think `copyright` is a var somewhere

go.mod
dirkmueller: Please keep one unit test :) really isn't great practive to remove it first ;)

go.mod
```diff
@@ -4,8 +4,9 @@ go 1.16
 
 require (
 	github.com/gin-gonic/gin v1.6.3
```
dirkmueller, line 6: looks like this line should be removed? we have no gin-gonic use anymore? what am I missing?

webapp/templates/footer.html.tmpl
```diff
@@ -5,4 +5,4 @@
     <li class="footer-list-item">{{ .copyright }}</li>
   </ul>
 </footer>
-{{ end }}
```
dirkmueller, line 8: NIT (not needing to be addressed): kinda prefer line breaks at the end do we have a policy around this?

cmd/webapp/webapp.go
```diff
@@ -35,15 +38,60 @@ func NewWebappCmd() *cobra.Command {
...
+}
+
+// Index data is used for the home template
+type Index struct {
+	Title string
+}
+
+func IndexHandler(w http.ResponseWriter, r *http.Request) {
+	data := Index{
+		Title: "Sapconsole",
```
dirkmueller, line 70: should be `SUSE Console for SAP Applications`

go.mod
MalloZup: yes

MalloZup: I ran go mod tidy is removed

webapp/templates/footer.html.tmpl
stefanotorresi: I guess `.tmpl` files are not covered by the style formatter, but yes, linefeed should always be ensured. Maybe we should add an [`.editorconfig`](https://editorconfig.org/) file.

arbulu89 (commented):
Personally I find this change worst than the previous one.
Now, all the content templates will need to have duplicated code always (header, sidebar, footer). In fact, I found more minimalist the last approach, where just some simple coding (less than 100 lines) we had this covered.
In fact, as it is still incomplete we cannot still compare I guess. We would still need to pass the common variables to the individual pages, and set the methods that were used before in the templates.
Besides that, the templates are loaded from local folders, and not as a File System. This is not something I can really judge, but Stefano included the FS option for some reason.

In my opinion:
- The templating solution looks worst (which is independent on chi vs gin gonic)
- I personally liked more the gin-gonic code style
- I don't see which benefit would bring chi over gin-gonic (but I would say the same in other direction, that's true).

So.... I don't know what to say. I see gin-gonic more suitable for our purpose (considering that we just want a web page displayed, we are not talking about REST APIs here). I don't see why a framework-kindof package would be worst here.

And to finish: I don't really have any strong preference between the 2 tools, for me they do exactly the same in our use case, but gin-gonic looks easier to use in general (and as far as I have seen, there is much more information in the internet about the 2nd, in stackoverflow, general documentation, etc, but I might be wrong, I have just looked for some few minutes)

webapp/templates/home.html.tmpl
```diff
@@ -1,4 +1,17 @@
...
+      {{ template "submenu" . }}
+      <div class="container">
+        <h1>{{ .Title }}</h1>
+         <p>This is the home page</p>
+      </div>
+    </section>
+    {{ template "footer"}}
+  </body>
+</html>
+{{end}}
```
stefanotorresi, line 17: This is the exact opposite of what we want: we want a generic layout where we override the "content" block, not a specific page where we duplicate all the inner block calls.

webapp/templates/sidebar.html.tmpl
stefanotorresi: We introduced that template function to render escaped HTML code inside HTML attributes.

cmd/webapp/webapp.go
```diff
@@ -35,15 +38,60 @@ func NewWebappCmd() *cobra.Command {
...
+	}
+	// you access the cached templates with the defined name, not the filename
+	err := allTemplates.ExecuteTemplate(w, "indexPage", data)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+}
```
stefanotorresi, line 78: Could you please keep all the pages handler in separate go files, like `home.go`, like it was before?

stefanotorresi (changes requested):
@MalloZup I must say I share Xabi's concernes here: all of these changes are a step backward, they remove functionality to achieve the only goal of using `chi` instead of `gin`.
It feels like you're working for the tool, rather than making the tool work for you.

cmd/webapp/webapp.go
MalloZup: https://github.com/SUSE/console-for-sap-applications/pull/7/commits/67a589510ef9822a20a6bfdf6361772e7110b46c

adressed

cmd/webapp/webapp.go
MalloZup: adressed

go.mod
MalloZup: https://github.com/SUSE/console-for-sap-applications/pull/7/commits/67a589510ef9822a20a6bfdf6361772e7110b46c done in that

webapp/templates/sidebar.html.tmpl
MalloZup: done

webapp/templates/home.html.tmpl
MalloZup: done

MalloZup:
PR reworked and ready to review

web/home.go
```diff
@@ -1,11 +1,37 @@
...
-	c.HTML(http.StatusOK, "home.html.tmpl", gin.H{})
+// Index data is used for the home template
+type Index struct {
+	Title     string
+	Copyright string
+}
+
+func IndexHandler(templates map[string]template.Template) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		data := Index{
```
arbulu89, line 17: Why do we have the `Title` and `Copyright` here in the `home` handler?
This values are used only in the layout and they shouldn't belong here.
The previous implementation added this items before, so you don't need to have them in every handler code

web/home.go
```diff
@@ -1,11 +1,37 @@
...
 
-func homeHandler(c *gin.Context) {
-	c.HTML(http.StatusOK, "home.html.tmpl", gin.H{})
+// Index data is used for the home template
+type Index struct {
+	Title     string
+	Copyright string
+}
+
+func IndexHandler(templates map[string]template.Template) http.HandlerFunc {
```
arbulu89, line 15: This should be the `HomeHandler`

web/template_manager.go
```diff
@@ -0,0 +1,71 @@
...
+			var out bytes.Buffer
+			_ = tmpl.ExecuteTemplate(&out, name, data)
+			return out.String()
+		},
+	})
+	// parse all templates
+	patterns := append([]string{base, file}, []string{blocksHtml}...)
+	tmpl = template.Must(tmpl.ParseFS(templatesFS, patterns...))
+
+	// add template to template map, consumed by handlers
```
arbulu89, line 60: Why have you added the `addTemplate` code here?
Now the method is bigger, I personally preferred having it as other method

web/template_manager.go
```diff
@@ -0,0 +1,71 @@
...
+				continue
+			}
+			addFileFromFS(templatesFS, file, templs)
+		}
+	}
+	return templs
+}
+
+// addFileFromFS parses the base template with the user
+func addFileFromFS(templatesFS fs.FS, file string, templs map[string]template.Template) {
```
arbulu89, line 42: A possible dumb question. Is there any reason why this `addFileFromFS` starts with lower case and `AddAllTemplatesFromFS` with upper case?
Does this mean that the 1st is private and the 2nd public?

arbulu89 (commented):
Some comments from my side

web/template_manager.go
MalloZup: yes this is exported in golang

web/template_manager.go
MalloZup: previously we were having 2 issues:

1) we were using `methods`  but we didn't have any state to manipulate ( https://flaviocopes.com/golang-methods-or-functions/)

2) I merged the 2 methods because  of 2 reasons:

- is better to have deeper functions ( and bigger)  rather then shorter one.
-  conceptually having 2 methods initially, is where supposed to do 2 different operations but the function we merged where just helper function. In our code, is better to don't create abstraction when we don't need them.

web/home.go
MalloZup: the previous implementation where using some go gin internal. 

I think we need to investigate if we can pass a global struct to all templates as the gin did

MalloZup:
I just post why I condensed the 4 methods in only 2. 
image

is from a philosophy of sofware design book.

In our case specific, there is no need to create 4 functions abstraction when we can just do the work with less duplication. 

or you can have also a refererence in thanos coding style:

https://github.com/thanos-io/thanos/blob/main/docs/contributing/coding-style-guide.md#defers-dont-forget-to-check-returned-errors
image

web/template_manager.go
stefanotorresi: > 1. we were using `methods`  but we didn't have any state to manipulate ( https://flaviocopes.com/golang-methods-or-functions/)

That's not true, previously we were building a struct with an interface associated with it, which abstracted both the behaviour and the data, now you're just returning data and passing that around everywhere it is needed.
I don't find any of the two approaches superior to the other, but we didn't need to pass all the templates to all the HTTP handlers before, so IMHO the new interface to the templates is less convenient than the previous one, considering we also have to check for slice keys every time. You will have to rebuild yet another abstraction that we already had in place.

> 2\. I merged the 2 methods because  of 2 reasons:
> 
> * is better to have deeper functions ( and bigger)  rather then shorter one.
> * conceptually having 2 methods initially, is where supposed to do 2 different operations but the function we merged where just helper function. In our code, is better to don't create abstraction when we don't need them.

That's actually totally the opposite of the best practices I would follow, which is to have little functions that are easy to read and understand because they only do one thing and very well, pretty much like the entire UNIX philosophy. You are a proponent of minimalism yourself, so I find this change very contradictory. 😛 

At the end of the day I share Xabi's view here: in-lining those two functions makes the code harder to read and understand, so I would keep them separated.

stefanotorresi:
> In our case specific, there is no need to create 4 functions abstraction when we can just do the work with less duplication.

those four functions were private, they were not part of the abstraction at all, the only abstraction was the Render constructor, which was indeed a very deep interface. Deep modules doesn't mean in-lining an entire module into a single function.

web/home.go
```diff
@@ -1,11 +1,37 @@
...
+				http.StatusInternalServerError)
+			return
+		}
+
+		err := tmpl.ExecuteTemplate(w, "base", data)
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+
```
stefanotorresi, line 35: I think we're missing to set the http status code and the `content-type` header, which was something that the `gin` render took care for us. How do we do this with `chi`?

web/home_test.go
```diff
@@ -1,20 +1,41 @@
...
+	expected := "SUSE Console for SAP Applications"
+	if b, err := ioutil.ReadAll(rr.Body); err != nil {
+		t.Fail()
+	} else {
+		if strings.Contains(string(b), "Error") {
+			t.Errorf("header response shouldn't return error: %s", b)
+		} else if !strings.Contains(string(b), expected) {
+			t.Errorf("header response doesn't match:\n%s", b)
+		}
+	}
```
stefanotorresi, line 39: I don't really understand why did you change this test, I don't think this should be needed at all. The test used the default http testing toolkit and `testify` assertions, could you please reinstate that?
I think you're only need to invoke `ServeHTTP` on a different struct.

web/home.go
stefanotorresi: We built the `addLayoutData` function on top of the following gin struct:
https://github.com/gin-gonic/gin/blob/master/render/html.go#L40-L45

We'd basically need to do the same with the new data structure you're introducing.

web/template_manager.go
stefanotorresi: @arbulu89 precisely. As @MalloZup mentioned, in golang everything is either "exporter" or not, from a package. You can use non-exported symbols within the same package, but they must be exported to be used in other packages; bear in mind that there is no package hierarchy, i.e. even if you can nest packages with directories the package hierarchy gets flattened, so to use a symbol in a pseudo-subpackage, you need to export it.

cmd/web/webapp.go
```diff
@@ -35,15 +33,10 @@ func NewWebappCmd() *cobra.Command {
...
 func serve(cmd *cobra.Command, args []string) {
-	engine := web.NewEngine()
-
-	s := &http.Server{
-		Addr:           fmt.Sprintf("%s:%d", host, port),
-		Handler:        engine,
-		ReadTimeout:    10 * time.Second,
-		WriteTimeout:   10 * time.Second,
-		MaxHeaderBytes: 1 << 20,
-	}
```
stefanotorresi, line 46: This configuration was here for a reason: default timeouts and header limit of the go http server are not ideal.

web/template_manager.go
MalloZup: Small functions are not easier to read. Especially here, we are creating a function for the sake of building a `container` for code which is used only 1once. It should be not even a function. A function is not a place holder for code,
 but in reality the function is just used once. There is not really anything with the unix principles. 

please read this: 

https://copyconstruct.medium.com/small-functions-considered-harmful-91035d316c29

MalloZup: https://user-images.githubusercontent.com/10886597/111984614-e18dcd80-8b0b-11eb-9773-cf92560279ac.png

web/home_test.go
MalloZup: this is more subjective taste on how to test, but I personally don't care to reintroduce what was before

web/template_manager.go
stefanotorresi: @MalloZup I have stumbled upon that article in the past and I must say I simply don't agree with most of it.
The only good take away I can find is that we should not be dogmatic, which me and Xabi hardly are: those additional functions were extracted empirically, as a refactoring of a function that became too long and thus harder to explain, which I actually had to do during a live code review session between me and @arbulu89.

You are basically rewriting the code that I already wrote and then refactored, back to its non re-factored form, which I frankly find difficult to consider as an improvement.

Since you have two colleagues expressing a personal preference here, could you please be so kind to make life easier for us all?

web/templates/base.html.tmpl
```diff
@@ -1,3 +1,4 @@
+{{define "base"}}
```
stefanotorresi, line 1: I'm not quite clear, what do we need a `base` template block for?

web/template_manager.go
MalloZup: in this context:

I don't think that 2 colleagues vs 1 one opinion, is literally a strong point to make technical choice.
However for the simplicity/realistic mind, I will bring it back the code as the other reviewers, I think for the sake of the present PR this refactoring is `secondary` but as intellectual and open learning discussion is still valid.

Although I don't agree with your points and I find unfortunate that you miss the occasion to think/discuss about it.

example for a big function from terraform: https://github.com/hashicorp/terraform/blob/main/backend/local/backend_local.go#L32-L139

@stefanotorresi why you don't agree with the article?  There are also other books  who plaid in favor of such argument. In my experience, there are many occasion where small function are suggested as good, but in many codebases they are just a PITA to maintain and understand in term of complexity. ( that is also called spaghetti code).

Refactoring is given a shape or structure to a code. We need to consider that also when a given shape is modeled on a codebase, this has a several disavantages.

So saying `refactoring` as an improvement, is wrong because refactoring can harm the cognitive perception of the code as a "pure" bigger function is much easier to understand.

Also in this context, we are building a function for code which is used only 1 in its lifetime, isn't really a function.

MalloZup: as discussed, I will bring back to the original refactoring so we can discuss other things

MalloZup:
thx @stefanotorresi @arbulu89 for review I wll try to read your comments and address the review as far it is possible with the chi/mux way of doing things as we discussed.

Regarding the Refactoring of smaller/big function, despite my personal point, I will omit such change and focus only on the parts which need an actual change.

web/template_manager.go
stefanotorresi: Let's agree to disagree, Dario.

PR #8, opened 2021-03-17
Use standard layout name web for pkg

MalloZup:
The scaffold of this project uses already `golang-standard` directory layout.
So use it for all dirs.

### Changes:

`webapp` -> `web`
# See

see https://github.com/golang-standards/project-layout#web

also is better to avoid duplicate pkg names we have 2 webapp


stefanotorresi:
Could you please remove `node_modules` from checked in files? The diff is humongous.
I think you missed to update `.gitignore` to avoid that.

MalloZup:
ok updating

MalloZup:
@stefanotorresi ok updated. Should be ok now

stefanotorresi:
> also is better to avoid duplicate pkg names we have 2 webapp

Well, one package path is `cmd/webapp` and one is just `webapp`, i.e. the final name is the same but the import path is different. Sadly this is a limitation of Go package system, and as far as I could see everybody usually resort to just alias child packages with snake_case notation in import statements.
It doesn't look quite right to me to keep `cmd/webapp` and the rest under `web`, so I would also change `cmd/webapp` to `cmd/web`. Opinions?

MalloZup:
I'm ok to rename `cmd/web` or keep it as it is right now.

MalloZup:
renamed accordingly

stefanotorresi (approved):

PR #11, opened 2021-03-30
HA Checker first iteration

stefanotorresi:
This PR is the result of my hackweek. It works, but needs a lot of validation from you folks. For the time being take it as a proof of concept.

It introduces the _Agent_ component, which integrates with Consul as a service discovery backend.
The agent provides the following functionalities:
- it registers a node on a local Consul agent, with a couple of Consul health checks (one of TTL type, one of TCP type).
- it drives the HA Checker, by leveraging the [aquasecurity/bench-common](https://github.com/aquasecurity/bench-common) project
- it exports the results of the HA Checker via a web service (sadly, it's not possible to use Consul itself to transports these, because the output payload of a Consul health check is limited to 4KB)

To demonstrate the Agent working, a new "Environment" web page has been added to the Web UI, which does nothing more than pretty-printing the data exposed the Consul API and the HA Checker web service.

To test this, a Consul server is now needed where the Web UI runs, and a Consul client is needed where the _Agent_ runs.
Both functions can be absolved by the same Consul instance when using development mode, so to run this locally you only need to get the Consul binary and run `consul agent -dev -ui -datacenter=test`.


agent/agent.go
```diff
@@ -0,0 +1,217 @@
...
+		err := a.consul.Agent().ServiceDeregister(a.cfg.InstanceName)
+		if err != nil {
+			log.Println("An error occurred while trying to deregisterConsulService the agent service with Consul:", err)
+		} else {
+			log.Println("Consul service de-registered.")
+		}
+	}()
+
+	var wg sync.WaitGroup
+	wg.Add(2)
```
arbulu89, line 103: What is this `2` ?

agent/agent.go
```diff
@@ -0,0 +1,217 @@
...
+func (a *Agent) registerConsulService() error {
+	var err error
+
+	consulService := &consul.AgentServiceRegistration{
+		ID:   a.cfg.InstanceName,
+		Name: a.cfg.ServiceName,
+		Tags: []string{"console-agent"},
+		Checks: consul.AgentServiceChecks{
+			&consul.AgentServiceCheck{
+				CheckID:  "hana_tcp_check",
```
arbulu89, line 148: What is this `hana_tcp_check` ?

examples/azure-rules.yaml
```diff
@@ -0,0 +1,604 @@
+---
```
arbulu89, line 1: Is this nomenclature based on `bench-common` tool?

arbulu89 (commented):
My initial thoughts:
1. Using `consul` looks right, but I guess other tools have been checked. This is one of the most important decisions of the whole console. I have checked it (`consul`) and it looks like it has all of the things we look for (service discovery, configuration management and network communication), and I guess that the tool is written on `go` is a plus for you guys (waiting already for go world domination... XD). In any case, double checking potential issues shipping a consul based tool is needed

2. I'm not sure on `bench-common` usage. I understand that creating something similar is a high effort, and most probably there are other solutions. But this tool is still even in beta, which brings many issues (a lot of future major changes, lack of maintenance, etc)

3. Providing the HA checks as http endpoint must be studied

4. I understand that the whole work was done together in sake of speed and to combine efforts, but I think the implementation of the HA checker and the consul usage should be completely separated
2.

stefanotorresi:
Provisional architecture diagram for reference:

```
┌──────────────────────────────────────────────────────────────────┐
│ Datacenter                                                       │
│                     ┌─────────────────────────────────────────┐  │
│ ┌──────────────┐    │SAP Environment                          │  │
│ │ Trento       │    │ ┌─────────────────────────────────────┐ │  │
│ │  ┌────────┐  │    │ │SAP Landscape                        │ │  │
│ │  │        │  │    │ │ ┌──────────────┐   ┌──────────────┐ │ │  │
│ │  │ Trento │  │    │ │ │SAP Host      │   │SAP Host      │ │ │  │
│ │  │  WebUI │  │    │ │ │ ┌─────────┐  │   │ ┌─────────┐  │ │ │  │
│ │  │        ├──┼────┼─┼─┼─┤► Trento │  │   ├─┤► Trento │  │ │ │  │
│ │  │   │    │  │    │ │ │ │   Agent │  │   │ │   Agent │  │ │ │  │
│ │  └───┼────┘  │    │ │ │ │   │     │  │   │ │   │     │  │ │ │  │
│ │      │       │    │ │ │ └───┼─────┘  │   │ └───┼─────┘  │ │ │  │
│ │  ┌───▼─────┐ │    │ │ │     │        │   │     │        │ │ │  │
│ │  │Consul   │ │    │ │ │ ┌───▼─────┐  │   │ ┌───▼─────┐  │ │ │  │
│ │  │ Agent   └─┼────┼─┼─┼─►Consul   │  │   ├─►Consul   │  │ │ │  │
│ │  │ (Server)◄─┼────┼─┼─┼─┐ Agent   │  │   ├─┐ Agent   │  │ │ │  │
│ │  │         │ │    │ │ │ │ (Client)│  │   │ │ (Client)│  │ │ │  │
│ │  └─────────┘ │    │ │ │ └─────────┘  │   │ └─────────┘  │ │ │  │
│ │              │    │ │ │              │   │              │ │ │  │
│ └──────────────┘    │ │ └──────────────┘   └──────────────┘ │ │  │
│                     │ │                                     │ │  │
│                     │ └─────────────────────────────────────┘ │  │
│                     │                                         │  │
│                     └─────────────────────────────────────────┘  │
│                                                                  │
└──────────────────────────────────────────────────────────────────┘
```

agent/agent.go
stefanotorresi: `2` is the number of items we want to wait for with the WaitGroup.
Here we start 2 goroutines we want to wait for, so we add `2` to the WG.
In each of those goroutines, below, `wg.Done()` is called: this does nothing more than decrementing the counter by one. Once the counter hits 0, the WG as a whole is considered done and `wg.Wait()` will finish blocking.

This whole mechanism is there to prevent goroutines from leaking (i.e. goroutines which are started but not ensured to actually stop before their caller terminates).

agent/webservice.go
```diff
@@ -0,0 +1,87 @@
+package agent
```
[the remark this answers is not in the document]
dirkmueller: +1

dirkmueller:
This is now in  #14 #16 and #17 . closing here.

PR #13, opened 2021-04-07
Branding Update for trento

dirkmueller:


stefanotorresi:
I was thinking about naming the CLI command `tn`, which is the official code for the province of Trento, while still naming the git module `trento`. Do you think we can do this here?

README.md
```diff
@@ -1,6 +1,6 @@
-# SUSE Console for SAP Applications v2
+# Trento
 
-A cloud-native, web application to manage OS-related tasks for SAP Applications.
+An open cloud-native, web console for managing SAP workloads on Linux.
```
stefanotorresi, line 3: this conflicts with #12, maybe we could leave these diff lines out from this PR?

dirkmueller: I have rebased this now on top of #12 after merging that one.

dirkmueller:
> I was thinking about naming the CLI command `tn`, which is the official code for the province of Trento, while still naming the Go module `trento`. Do you think we can do this here?

I'm not sure this is worth it (we don't aim for the cli command to be end user callable, it would be a container entrypoint or a systemd service or whatever else we use for launching) and "tn" already conflicts with /usr/bin/tn (telnet)

stefanotorresi:
Good points! Let's stick with `trento` then.

stefanotorresi (approved):

MalloZup (commented):
lgtm

PR #18, opened 2021-04-12
Update nodes visualization in the webapp

arbulu89:
Implement some new transition to show the nodes information. This is just a new iteration to show how the node visualization could work. It doesn't have any filtering by now (this will come later). I assume that almost the whole frontend will be refactored in the future using some other JS tools anyway. For this reason, I didn't spend much type with the styling. The idea is just to have an initial code to start working with the future grouping and filtering.

- Show the nodes as a list
- Show the checks of these nodes
- Show the HA checks for these nodes

Styles following EOS design: 
https://suse.eosdesignsystem.com/tables

Some pics:
image
image
image
image


stefanotorresi (commented):
Good stuff, Xabi.

I guess that the grouping in Environments, Landscapes, Systems, Clusters, and Hosts will come later?

Also bear in bind that so far we didn't expose Consul's `datacenter` field explicitly, but we rather used that field to identify the root "Environment" entity, which is the outmost layer in the SAP specific domain.

If you decide to expose the datacenter separation explicitly, then we need something to identify the SAP environments as well.

arbulu89:
> Good stuff, Xabi.
> 
> I guess that the grouping in Environments, Landscapes, Systems, Clusters, and Hosts will come later?
> 
Yes, the filtering comes just in the next step.

> Also bear in bind that so far we didn't expose Consul's `datacenter` field explicitly, but we rather used that field to identify the root "Environment" entity, which is the outmost layer in the SAP specific domain.
> 
> If you decide to expose the datacenter separation explicitly, then we need something to identify the SAP environments as well.

I just printed the `datacenter` because it was already there. We can give the purpose we want. If we understand the `datacenter` as the SAP `Environment` we will need to have multiple `datacenters` by default, which I don't really like. I think the SAP environment must have other meaning. We will need to discuss these things next

stefanotorresi:
> If we understand the `datacenter` as the SAP `Environment` we will need to have multiple `datacenters` by default, which I don't really like. I think the SAP environment must have other meaning.

Fair enough! Let's keep these concepts separate then.

diegoakechi:
@arbulu89 @stefanotorresi I would suggest that the datacenter really represents the datacenter =P. It is common to have metro clusters, where nodes are hosted on two different data-centers or regions for CSP, so this can be also an relevant tag/group

diegoakechi:
@arbulu89  in general looks going on the right direction. To make the visualization cleaner, I would suppress the output of command and remediation columns as they are more verbose, and only present that via clicking to expand, for example, so the user can focus first on what is failed, for example. Also, for passed checks, the remediation text does not make much sense to be presented. 

None of these improvements are needed, for now, just wanted to have it registered.

dirkmueller:
Changes look good to me. I do wonder if we want to do an /environments/:name path (e.g. is :name the uuid? can these names change?) . anyway this can also be adjusted later.

arbulu89:
> Changes look good to me. I do wonder if we want to do an /environments/:name path (e.g. is :name the uuid? can these names change?) . anyway this can also be adjusted later.

Consul api uses the name, that's why I used it. The `name` must be unique in the datacenter. Each node has an unique ID too, but it makes difficult to handle the queries.
https://www.consul.io/api-docs/health#list-checks-for-node
https://www.consul.io/api-docs/catalog#retrieve-map-of-services-for-a-node

dirkmueller:
@arbulu89 I see. okay.

dirkmueller (dismissed):

stefanotorresi (approved):

PR #19, opened 2021-04-13
Add command $ trento agent run-once <YAML FILE>

aleksei-burlakov:
It works, but it should be redone I guess. I doesn't print pretty
and it exports private fields to outside, that should not be done I
guess.

@arbulu89 , @dirkmueller could you please have a look?


aleksei-burlakov:
I have reopened it in #20

PR #20, opened 2021-04-13
Add command $ trento agent run-once <YAML FILE>

aleksei-burlakov:
It works, but it should be redone I guess. I doesn't print pretty
and it exports private fields to outside, that should not be done I
guess.

### DIFF
--- a/agent/checker.go
+++ b/agent/checker.go
@@ -2,7 +2,9 @@ package agent
 
 import (
 	"fmt"
+	"io"
 	"io/ioutil"
+	"strconv"
 
 	"github.com/aquasecurity/bench-common/check"
 	"github.com/pkg/errors"
@@ -33,6 +35,14 @@ func NewChecker(definitionsPath string) (Checker, error) {
 
 type CheckResult struct {
 	controls *check.Controls
+}
+
+func (cr *CheckResult) PrettyPrint(sw io.StringWriter) {
+	sw.WriteString("== Summary ==\n")
+	sw.WriteString(strconv.Itoa(cr.controls.Summary.Pass) + " checks PASS\n")
+	sw.WriteString(strconv.Itoa(cr.controls.Summary.Fail) + " checks FAIL\n")
+	sw.WriteString(strconv.Itoa(cr.controls.Summary.Warn) + " checks WARN\n")
+	sw.WriteString(strconv.Itoa(cr.controls.Summary.Info) + " checks INFO\n")
 }
 
 func (r CheckResult) Summary() check.Summary {
--- a/cmd/agent/agent.go
+++ b/cmd/agent/agent.go
@@ -24,6 +24,13 @@ func NewAgentCmd() *cobra.Command {
 		Short: "Command tree related to the agent component",
 	}
 
+	runOnceCmd := &cobra.Command{
+		Use:   "run-once",
+		Short: "run-once",
+		Run:   runOnce,
+		Args:  startArgsValidator,
+	}
+
 	startCmd := &cobra.Command{
 		Use:   "start path/to/definitions.yaml",
 		Short: "Start the agent",
@@ -37,7 +44,25 @@ func NewAgentCmd() *cobra.Command {
 
 	agentCmd.AddCommand(startCmd)
 
+	agentCmd.AddCommand(runOnceCmd)
+
 	return agentCmd
+}
+
+func runOnce(cmd *cobra.Command, args []string) {
+	var err error
+
+	checker, err := agent.NewChecker(args[0])
+	if err != nil {
+		log.Fatal("Failed to create a Checker instance: ", err)
+	}
+
+	res, err := checker()
+	if err != nil {
+		log.Fatal("Failed to do checks: ", err)
+	}
+
+	res.PrettyPrint(os.Stdout)
 }
 
 func start(cmd *cobra.Command, args []string) {


aleksei-burlakov:
@dirkmueller , @arbulu89 could you please have a look?

agent/agent.go
```diff
@@ -32,6 +32,8 @@ type Config struct {
 	TTL             time.Duration
 }
 
+func (a *Agent) DoCheck() (CheckResult, error) { return a.check() }
```
stefanotorresi, line 35: Hmm, I don't quite like this to be honest, it feels like some avoidable repetition.

It would make sense if we offloaded more work to this function by completely encapsulating the `Checker` here, which means we don't want to leak its internal details like the `CheckResult` struct; so we could then probably rename this method to `RunChecks`, and put the output logic invocation here, maybe by passing a `Writer` interface so that we're not hard coding side effects on the stdout.

Either this, or just make `check` public and call it from the command handler as it is now. ;)

In general, if you have a `SomeVerb` method, I find having another `DoSomeVerb` method somewhere else to be a little of an anti-pattern. You are either doing that thing, or doing something else! ;)

stefanotorresi (changes requested):
Hey Aleksei, this looks good!

I have a remark about the additional method on the `Agent` struct.

Plus we certainly need the pretty-printing added here.

cmd/agent/agent.go
```diff
@@ -37,9 +44,36 @@ func NewAgentCmd() *cobra.Command {
...
+
+	cfg.DefinitionsPath = args[0]
+	cfg.ServiceName = serviceName
+	cfg.WebPort = port
+	cfg.TTL = TTL
+
+	a, err := agent.NewWithConfig(cfg)
+	if err != nil {
+		log.Fatal("Failed to create the agent: ", err)
+	}
```
dirkmueller, line 68: this appears to be a 1:1 copy of the same section in func start. not trying to be a pain in the ass, but is there a reason why you didn't split it in a small private function that does that and returns the cfg to call in both places?

stefanotorresi: good point! Also, we don't need any of
```golang
cfg.ServiceName = serviceName
cfg.WebPort = port
cfg.TTL = TTL
```

In the `run-once` command.

stefanotorresi: Now that I think about it, we don't actually need an `Agent` instance at all for this command, you just need to instantiate the `Checker`, and run it by invoking it (the checker itself is just a function).

So the command handler could be something as simple as:
```golang
checker, err := NewChecker(args[0])
// err handling
res, err := checker()
// err handling
res.PrettyPrint(os.Stdout) // or something like that
```

With this approach, my comment above about `agent.DoChecks` would not apply anymore.

dirkmueller: well, that opens the larger question on why we tag this onto the "agent" subcommand, it really is just a separate "let me debug this quickly" command. Eventually however we need to figure out how to do priv separation between checker and agent.

stefanotorresi: Good point. For the time being, I would just put all the commands related to local execution on a target node to the `agent` group. As more sub-commands will spawn, we'll get a better idea on how to adjust that.

agent/agent.go
aleksei-burlakov: Fixed. I have moved the call inside the agent/agent/go, there is no more need to export the method.

cmd/agent/agent.go
aleksei-burlakov: I have removed the 3 commands 

```
cfg.ServiceName = serviceName
cfg.WebPort = port
cfg.TTL = TTL
```
And now it's only 3 command + error checks. I don't thinks it's worthy now to put them in a separate function, it would only unnecessary encomplicate it now.

aleksei-burlakov: > Now that I think about it, we don't actually need an `Agent` instance at all for this command, we just need to instantiate the `Checker`, and run it by invoking it (the checker itself is just a function).
> 
> So the command handler could be something as simple as:
> 
> ```go
> checker, err := NewChecker(args[0])
> // err handling
> res, err := checker()
> // err handling
> res.PrettyPrint(os.Stdout) // or something like that
> ```
> 
> With this approach, my comment above about `agent.DoChecks` would not apply anymore.

And I have moved the DoCheck call inside the agent/agent.go. But there a problem arose, I don't no how to cope with it in go. It's an unused parameter. I did it in the python style

`_, err := agent.NewWithConfig(cfg)`

with the underdash, but it seems to fail in golang.

agent/agent.go
```diff
@@ -53,6 +53,12 @@ func NewWithConfig(cfg Config) (*Agent, error) {
 		return nil, errors.Wrap(err, "could not create a Checker instance")
 	}
 
+	res, err := checker()
+	if err != nil {
+		log.Fatal("Failed to do checks: ", err)
+	}
+	res.PrettyPrint(os.Stdout)
+
```
stefanotorresi, line 61: Hmm, I think this block is misplaced. We should not execute the checker when we instantiate a new Agent! What I suggested earlier is to put this in the `runOnce` command handler, copying the `Checker` instantiation from here.

cmd/agent/agent.go
```diff
@@ -37,9 +44,27 @@ func NewAgentCmd() *cobra.Command {
...
+	if err != nil {
+		log.Fatal("Failed to create the agent configuration: ", err)
+	}
+
+	cfg.DefinitionsPath = args[0]
+
+	_, err := agent.NewWithConfig(cfg)
+	if err != nil {
+		log.Fatal("Failed to create the agent: ", err)
+	}
```
stefanotorresi, line 65: Once you address my previous comment, all of these lines will no longer be needed: we don't need to create an Agent instance, we need to create a Checker one.

agent/checker.go
```diff
@@ -35,6 +37,14 @@ type CheckResult struct {
 	controls *check.Controls
 }
 
+func (cr *CheckResult) PrettyPrint(f *os.File) {
```
stefanotorresi, line 40: Consider using a `StringWriter` interface instead of `*os.File` as a type here. In general, interfaces rather than concrete implementations make the code more pluggable and easier to test.

stefanotorresi (changes requested):
nearly there!

stefanotorresi (approved):
Perfect work, Aleksei! It only miss a little test, but since there are no examples of how to test a command handler in the project yet, feel free to skip it for now.

dirkmueller (approved):

dirkmueller:
I'm okay without unittests for this functionality (it wasn't a story, only a task, so not demanded by Product Owner)

PR #21, opened 2021-04-13
add Consul mocks via mockery and reinstate the Environments page smoke test

stefanotorresi:
This was originally part of #11.
The `make generate` task effectively requires an external dependency, `mockery`, which can be grabbed via `go install github.com/vektra/mockery/v2@latest`. This is required to keep the mocks in sync with the interfaces they target.

The introduction of such a dependency might constitute a case for starting to think about providing a Dockerfile, to perform development and build task without having to install packages on a development box.


dirkmueller:
can you just update the README.md with the `go get github.com/vektra/mockery/v2` cmd and/or the Makefile check for it?

stefanotorresi:
> can you just update the README.md with the `go get github.com/vektra/mockery/v2` cmd and/or the Makefile check for it?

done!

dirkmueller (approved):
thanks!

PR #22, opened 2021-04-14
This is the first runtime check for corosync parameters

aleksei-burlakov:


examples/azure-rules.yaml
```diff
@@ -17,6 +17,15 @@ groups:
...
           Adjust the corosync tokem timeout as recommended on the Azure best practices. https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/high-availability-guide-suse-pacemaker 
         scored: true
+      - id: 1.1.1.runtime
+        description: "Corosync Token is set to 30000, check it runtime"
+        audit: 'corosync-cmapctl | grep "runtime.config.totem.token (u32) = " | sed "s/^.*= //"'
+        tests:
+          test_items:
+            - flag: 30000
+        remediation: |
+          Adjust the corosync tokem timeout as recommended on the Azure best practices. https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/high-availability-guide-suse-pacemaker
```
dirkmueller, line 27: maybe add also "restart.."? are we executing this test only if the on-disk comparison has succeeded (passed)?

dirkmueller (approved):

dirkmueller:
not worth waiting for a fix. lets merge it.

examples/azure-rules.yaml
aleksei-burlakov: Sry, do you mean to restart the corosync service?

PR #24, opened 2021-04-15
Nodes filtering and tagging

arbulu89:
Implementation of the filtering and tagging of the systems. The major part is documented in the README what I will describe here too.

The filtering of the nodes is done using the consul nodes meta-data: 
https://www.consul.io/docs/agent/options#node_meta
https://www.consul.io/api-docs/features/filtering
This way we can use the `catalog.nodes` endpoint to get the filtered nodes.

The metadata is set in the agent node configuration file. This metadata can be a static file or it can be updated with the `consul-template` utility. This tool takes values from the kv storage to populate and update the configuration files.

The nodes data is stored with the next hierarchy:
`trento/nodes/$nodename/metadata/$item` where `$item` can be `sap-environment`, `sap-landscape` and `sap-system`. This is translated to:
```
{
  "metadata": {
    "trento-sap-environment": "env1",
    "trento-sap-landscape": "land1",
    "trento-sap-system": "sys1"
  }
}
```
The `trento-` prefix is reserved for trento specific things and makes the usage of the filters easy in the web app.
The tags showed in the filter are obrained from the kv storage too. In this path:
`trento/filters/sap-environments`, `trento/filters/sap-landscapes`, `trento/filters/sap-systems`. They must have a json list format `["env1", "env2"]`. 

We will see if these data can be obtained automatically from the systems in the future anyway.

Pics:
image
image
image

PD:
1. Don't spend that much time checking the html code as this will be changed most probably
2. The dropdown elements are done with `bootstrap-select`: https://developer.snapappointments.com/bootstrap-select/

The major drawback of using `consul-template` is that this is yet another executable. I have not figured out if it can be used as go code inside the agent yet.

Let me know what do you about the kv storage hierarchy too.


web/frontend/package-lock.json
```diff
@@ -1,242 +1,8 @@
 {
   "name": "trento",
   "version": "0.0.1",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
stefanotorresi, line 4: This indicates you're using an old version of `npm`. Could you please update to v7 (`npm update -g npm` I think), reset the lock file, and re-run npm install?

stefanotorresi (changes requested):

web/frontend/package-lock.json
arbulu89: Done in https://github.com/trento-project/trento/pull/24/commits/fd4fae72b1666dece96dcecc48e9a0b94f75f121

stefanotorresi (approved):

stefanotorresi:
LGTM!

dirkmueller (approved):

PR #25, opened 2021-04-19
Clusters view

arbulu89:
New `clusters` views. This views show the information about existing clusters and the nodes on them. Some data is hardcoded to make it more realistic (Nodes number, resources number, some resources in the cluster level).

In order to get this the value comes from the `kv` storage. This are the paths
`trento/clusters/${clusterName}/` (**clusterName is still a folder. It will store the cluster information when it is available).

Besides this, the node metadata stores the cluster which this nodes belongs to (`trento-ha-cluster`). The nodes of the cluster are obtained using this metadata filter. This metadata is obtained from the `kv` value storage (`trento/nodes/$node/metadata/ha-cluster`)

Pics:
image
image
image
image

### DIFF
--- a/web/clusters.go
+++ b/web/clusters.go
@@ -1,0 +1,111 @@
+package web
+
+import (
+	"fmt"
+	//"log"
+	"net/http"
+	"strings"
+
+	"github.com/gin-gonic/gin"
+	//consulApi "github.com/hashicorp/consul/api"
+	"github.com/pkg/errors"
+
+	"github.com/trento-project/trento/internal/consul"
+)
+
+const KV_CLUSTERS_PATH string = "trento/clusters"
+
+type Cluster struct {
+	Name string
+}
+
+type ClusterList map[string]*Cluster
+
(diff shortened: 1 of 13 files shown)


web/clusters.go
```diff
@@ -0,0 +1,95 @@
...
+	"net/http"
+	"strings"
+
+	"github.com/gin-gonic/gin"
+	"github.com/pkg/errors"
+
+	"github.com/trento-project/trento/internal/consul"
+)
+
+const KV_CLUSTERS_PATH string = "trento/clusters"
```
dirkmueller, line 15: ideally this would be a in a location that we can share with the agent (so that front and backend use the same copy of the const)

stefanotorresi: yeah, I suppose we could introduce a new package for everything related to persistence/storage, but I'm not sure we have enough to put there atm!

arbulu89: Agree!
I guess this can come in a future implementation with some kind of shared information package

dirkmueller (approved):

PR #26, opened 2021-04-19
Implement consul-template utility usage in the agent

arbulu89:
Implement in the agent the `consul-template` utility.
With this change we don't need to run `consul-template` as a separated execution making thins much easier.

It adds 2 new flags to the agent start execution to provide some information about the consul agent.

In a summary, the code detects changes in the template code every time an entry is changed in the kv storage, and it reloads the agent configuration data to get the latest information from the meta-data values.

**Disclaimer: By now, the `consul-template` only connects to the localhost consul agent. It doesn't have the option to connect to some consul agent running in a secondary network address in the current machine. This is of course doable, but as our trento agent itself has the same limitation I have decided not to add this functionality in this PR**


cmd/agent/agent.go
```diff
@@ -86,12 +93,24 @@ func start(cmd *cobra.Command, args []string) {
 		log.Fatal("Failed to create the agent: ", err)
 	}
 
+	consultemplateconfig := &agent.ConsulTemplateConfig{
+		Source:      templatePath,
+		Destination: fmt.Sprintf("%s/%s/%s", configDir, serviceName, METADATAFILE),
```
dirkmueller, line 98: NIT: This should be something like 

```path.Join(configDir, serviceName, METADATAFILE)```

File: agent/consultemplate.go [hunk 2]
```diff
@@ -0,0 +1,65 @@
...
+	consultemplateconfig "github.com/hashicorp/consul-template/config"
+	"github.com/hashicorp/consul-template/manager"
+)
+
+type ConsulTemplateConfig struct {
+	Source      string
+	Destination string
+}
+
+func StartConsulTemplate(a *Agent, c *ConsulTemplateConfig) (*manager.Runner, error) {
```
dirkmueller, line 20: to be more consistent with the other functions in agent, it could be 

```
func (a *Agent) StartConsulTemplate(c * ConsulTemplateConfig) (*manager.Runner, error) {
```

instead.

arbulu89: I'm not sure on this. This would state that this `StartConsulTemplate` method belongs to the `Agent` struct, and I created this code in a different go module to make it more explicit. It can be definitely done but to have all the pieces in order, it would require to move this methods to the `agent.go` module in my opinion (we don't want to have `agent` struct methods spread off in different files).
What do you think on this @stefanotorresi ?

stefanotorresi: The package is still the same, so there is nothing really preventing you from adding more methods to the `Agent` struct. Putting methods in an a different file is not uncommon as a Go convention, so I wouldn't worry about that. In fact, I would argue that `agent/agent.go` is too crowded already.
One thing I would actually change is to make this method private, because it's not meant to be consumed outside the Agent itself.

This also means that this method would need to be first invoked by `agent.Start()` instead of being called directly by the command handler, as it is now.

agent/consultemplate.go
```diff
@@ -0,0 +1,65 @@
...
+	// Create new thread method to check for consul-template events. If the template is rendered
+	// reload the configuration
+	go func() {
+		signals := make(chan os.Signal)
+		signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM)
+		for {
+			select {
+			case quit := <-signals:
+				log.Printf("Exiting from consul-template event listener loop: %s", quit)
+				break
```
stefanotorresi, line 45: I don't think we need to handle signals again here. This was done already by the `start()` command handler!
The `Agent` struct has a `ctx` field that you can use to control children goroutines.
This means that instead of waiting for signals in the first `case` statement here, you would instead wait for `<-a.ctx.Done()`, like it is done in `Agent.startCheckTicker()`

cmd/agent/agent.go
```diff
@@ -86,12 +93,24 @@ func start(cmd *cobra.Command, args []string) {
...
+	}
+	runner, err := agent.StartConsulTemplate(a, consultemplateconfig)
+	if err != nil {
+		log.Fatal("could not start consul template: ", err)
+	}
+
```
stefanotorresi, line 104: I think this should be encapsulated within `Agent.Start()` instead of being public API.

cmd/agent/agent.go
```diff
@@ -86,12 +93,24 @@ func start(cmd *cobra.Command, args []string) {
...
 	go func() {
 		quit := <-signals
 		log.Printf("Caught %s signal!", quit)
 
 		log.Println("Stopping the agent...")
 		a.Stop()
+
+		// Stop consul-template
+		agent.StopConsulTemplate(runner)
```
stefanotorresi, line 113: Same goes here: the CLI command handler is the only consumer of `Agent` public API, and from this standpoint nothing should have changed: we still start and stop just the `a` instance.

agent/consultemplate.go
```diff
@@ -0,0 +1,65 @@
...
+			}
+		}
+	}()
+
+	return runner, nil
+}
+
+func StopConsulTemplate(runner *manager.Runner) {
+	log.Println("Stopping consul-template")
+	runner.StopImmediately()
```
stefanotorresi, line 63: I think we can just move this call in `Agent.Stop()`

stefanotorresi (changes requested):
Hey Xabi, this is magnificent!

I would only change a few things to make the addition more cohesive with the existing subroutines handling.

If look at what we're doing in `Agent.Start()`, we already have a control structure to handle different goroutines spawned from there.

I think you can simplify the new `startConsulTemplate` method greatly by calling it within `Agent.Start()`, like we already do for `startCheckTicker()` or `webService.Start()`. Basically, we don't need to use goroutines in `startConsulTemplate`: instead we should handle these from `Agent.Start()` by adding another goroutine there which is then waited for via the existing `sync.WaitGroup` instance.

agent/consultemplate.go
stefanotorresi: I forgot to mention: as I implied with my initial review comment, this entire `for` loop should not be run into a goroutine. The only thing we need a goroutine for is `runner.Start()`. The Consul config reload cycle should instead block the entire execution, and we handle concurrency from the outside with the Context and the WaitGroup already available in `Agent.Start()`.

arbulu89:
@stefanotorresi Thanks for the review. The reason why I added everything in an independent execution was just to have it outside the agent logic, as I was not sure it should go there. Your implementation suggestions will move everything to the agent logic (which if fine).

arbulu89:
@stefanotorresi Changes done. It is true that now the consul-template code gets more attached to the agent and the go routines management.

stefanotorresi (approved):
Right on!

PR #28, opened 2021-04-19
Add initial Implementation to discover Facts of a HanaCluster

dirkmueller:


agent/agent.go
```diff
@@ -199,15 +219,36 @@ func (a *Agent) startCheckTicker() error {
...
+	ticker := time.NewTicker(a.cfg.DiscoverInterval / 2)
+	// Repeat after 15 minutes
+	if a.cfg.DiscoverInterval < time.Minute {
+		a.cfg.DiscoverInterval = 15 * time.Minute
+	}
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ticker.C:
+			for i := 0; i < len(a.discovery); i++ {
```
arbulu89, line 232: I don't really understand this.
What is the purpose of having a list of discoveries? How is this going to evolve in the future?

agent/agent.go
```diff
@@ -199,15 +219,36 @@ func (a *Agent) startCheckTicker() error {
...
+}
+
 func (a *Agent) updateConsulCheck(result CheckResult) {
 	log.Println("Updating Consul check TTL...")
 
 	var err error
 	var status string
 
-	summary := result.Summary()
+	summary := result.CheckSummary()
```
arbulu89, line 249: I preferred the old `Summary` naming, but both are fine :)

agent/discover/cluster.go
```diff
@@ -0,0 +1,91 @@
...
+	return true
+}
+
+func (cluster ClusterDiscover) KVStorePrefix() string {
+	return fmt.Sprintf(
+		"%s/%s",
+		consul_internal.KV_CLUSTERS_PATH, cluster.clusterName)
+}
+
+func NewClusterDiscover(client consul.Client) ClusterDiscoverInterface {
```
arbulu89, line 43: A dumb question. Why do we need here this `Interface` function that returns a function?
It looks pretty strange

agent/discover/cluster.go
```diff
@@ -0,0 +1,91 @@
+package discover
+
+import (
+	"fmt"
+	"log"
+	"strconv"
+
+	"github.com/ClusterLabs/ha_cluster_exporter/collector/pacemaker/cib"
```
arbulu89, line 8: I would put a comment here to raise awareness. In the future we will need to maybe replace this by specific code

agent/discover/discover.go
```diff
@@ -0,0 +1,52 @@
...
+
+	log.Print("kvPath", kvPath)
+
+	client.KV().Put(&consul.KVPair{
+		Key:   kvPath,
+		Value: []byte(cStoreValue)}, nil)
+}
+
+func CollectDiscoveries(discover DiscoverInterface, clusterDiscoveries []discoverFact, client consul.Client) error {
+	discover_session := sh.NewSession()
```
arbulu89, line 35: What is this `sh.NewSession` for? To understand the logic

arbulu89 (approved):
@dirkmueller The code looks good. There are many things I don't understand, but they don't look bad.
I think that by now we can move forward, but in the future we will need to have some more logic to create the cluster information more automatically in the KV storage.

Other side effect, is that all the nodes in the cluster will right in the storage. Maybe in the future we would need to run this only in the DC to avoid lock issues

internal/consul/kvstore.go
```diff
@@ -0,0 +1,12 @@
+package consul
+
+// ### add landscape and system to the name
+const KV_CLUSTERS_PATH string = "trento/v0/clusters"
+
+type ClusterStonithType int
+
+const (
+	ClusterStonithNone    = iota
```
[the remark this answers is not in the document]
dirkmueller: so you're saying constants should be snake case all upper case? can do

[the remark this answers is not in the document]
dirkmueller: done

agent/agent.go
dirkmueller: done

agent/discover/cluster.go
dirkmueller: comment added

agent/discover/discover.go
dirkmueller: Comment added. this code is right now not used. the idea is that we can use a declarative syntax (as an array initially, but maybe a yaml configuration file later on) with "shell command to run" and whee the stdout of the shell command is directly recorded as a value in the given KVStore path. 

I originally thought we'll need this, but then realized that the current JIRA ticket does not require this. I just wanted to keep it in because we'll most likely need it at some later point in time. 

If its confusing I can remove it for now and store it in a separate branch.

agent/discover/cluster.go
dirkmueller: I don't really know. I just copied this pattern from the Checker. there was no architecture discussion on that, it was just merged, so I assumed it is good practice.

agent/agent.go
dirkmueller: The idea is that we only run "discoveries" where it makes sense, e.g. we only collect Hana specific data on nodes that have either been tagged as such or are matching the detection that this is a hana node. 

at some point we can have a configurable list of "introspections" (Discoveries) to be executed by node, which is then filtered by "is this filter configured *and* does it match the expectation", and then only the configured collectors that apply are executed in the discover loop. 

For now there is only one.

agent/discover/discover.go
dirkmueller: I removed it to reduce the review obstacle.

stefanotorresi (commented):
Hey Dirk, this is great progress! That said, I'm a bit confused with how the various abstractions are modelled.

Putting aside for a moment that using the `Interface` suffix for interface types is not exactly idiomatic in Go, which contributes to my confusion, I'm not clear on why we have both a `Discover` function type and a `DiscoverInterface` type, and then we have an additional `ClusterDiscoverInterface` type, which I guess was supposed to be a subtype of `DiscoverInterface` (but it isn't, actually, since the latter is not composed by the former)?

This looks a bit convoluted: as far as I can see, the concrete implementations `ClusterDiscover` and `HanaClusterDiscover` implement `DiscoverInterface`, so why do we have this additional `ClusterDiscoverInterface`?

But more to the point, what is the behaviour that we want to abstract here? Is it the one defined by the `Discover` function type, or the one defined by the `DiscoverInterface` type? Why do we have two abstractions related to the same concept? It looks to me that we only consume the former one within the discovery loop: I would suggest to just drop the `Interface` prefix from the interface type, and make it a single `Discover` type that establishes what we want to consume in the Agent. The fact that we ended up with two types here could suggest that a function type was not enough, and we simply need a wider interface.

Another question about the addition of a TTL Check on Consul: why do we have such check for the discovery loop?
What would be the failure condition of such a check? As far as I can see, the `Discover` type doesn't return any error, so I assume it can't trigger a failure condition. Wouldn't everything just work identically without the TTL check? Or is this meant to only notify that the agent is up and running?

agent/discover/cluster.go
diegoakechi: Or move it to some kind of library. The go-pacemaker sounds like a good candidate and this project needs some love.

dirkmueller:
> Putting aside for a moment that using the `Interface` suffix for interface types is not exactly idiomatic in Go

changed to "-er" suffix as explained in effective Go now.

> and then we have an additional `ClusterDiscoverInterface` type, which I guess was supposed to be a subtype of `DiscoverInterface` (but it isn't, actually, since the latter is not composed by the former)?

I used to have additional functionality in ClusterDiscoverer (ClusterDiscoverInterface) but I postponed it before doing the PR, so this is a left over. removed. 

> This looks a bit convoluted: as far as I can see, the concrete implementations `ClusterDiscover` and `HanaClusterDiscover` implement `DiscoverInterface`, so why do we have this additional `ClusterDiscoverInterface`?

originally my thinking was that there are "always run Discoverers" and "conditionally run Discoverers". one would have a "should I be run? " method and the other one doesn't. but this can be simplified into only using "conditionally run" with a "return true" condition. 

> Another question about the addition of a TTL Check on Consul: why do we have such check for the discovery loop?
> What would be the failure condition of such a check? As far as I can see, the `Discover` type doesn't return any error, so I assume it can't trigger a failure condition.

I didn't implement error handling, but the idea is that, similiar to checker, it would report discovery execution failures in the services view. I don't find it useable to only report errors in the agent log, as that is really cumbersome for the user. It needs to be reported in the Web Console.

>  Wouldn't everything just work identically without the TTL check? Or is this meant to only notify that the agent is up and running?

Yes, we can also run it without the TTL check, but why?

agent/discover/cluster.go
```diff
@@ -0,0 +1,91 @@
...
+	consul "github.com/hashicorp/consul/api"
+	"github.com/pkg/errors"
+	consul_internal "github.com/trento-project/trento/internal/consul"
+)
+
+//const crm_monPath string = "/usr/sbin/crm_mon"
+//const cib_admPath string = "/usr/bin/cibadmin"
+
+const crm_monPath string = "./test/fake/crm_mon.sh"
+const cib_admPath string = "./test/fake_cibadmin.sh"
```
stefanotorresi, line 20: so how do we make this real? :)

stefanotorresi:
> I didn't implement error handling, but the idea is that, similiar to checker, it would report discovery execution failures in the services view. I don't find it useable to only report errors in the agent log, as that is really cumbersome for the user. It needs to be reported in the Web Console.

That makes sense, gotcha! I was just confused by a TTL check that would never fail. ;)

agent/discover/cluster.go
dirkmueller: sorry, removed. leftover from testing.

agent/agent.go
```diff
@@ -160,13 +178,20 @@ func (a *Agent) registerConsulService() error {
...
 		Checks: consul.AgentServiceChecks{
 			&consul.AgentServiceCheck{
 				CheckID: "ha_checks",
 				Name:    "HA config checks",
 				Notes:   "Checks whether or not the HA configuration is compliant with the provided best practices",
-				TTL:     a.cfg.TTL.String(),
+				TTL:     a.cfg.CheckerTTL.String(),
+				Status:  consul.HealthWarning,
+			},
+			&consul.AgentServiceCheck{
```
arbulu89, line 190: I'm not really sure why the discovery part goes in the Checks. To report if the discovery failed?

agent/agent.go
```diff
@@ -8,30 +8,33 @@ import (
...
 
 type Config struct {
+	CheckerTTL          time.Duration
 	WebHost             string
 	WebPort             int
 	ServiceName         string
 	InstanceName        string
 	DefinitionsPath     string
-	TTL                 time.Duration
+	DiscoverInterval    time.Duration
```
arbulu89, line 37: I guess that this eventually will become a list of `time.Duration`. If we have multiple discoveries, not all of them might need to have the same interval.
Just an observation. We can change this in the future

internal/consul/kvstore.go
```diff
@@ -0,0 +1,22 @@
...
+//
+// At some point in time this structure needs to become a versioned protocol between the
+// Trento Web Console and the Trento Agent
+
+// For now, until we tighten the structure around versioning, this file is a start
+// that records constants that need to be in sync or have compat behavior between
+// Console and Agent
+
+// ### add landscape and system to the name
+const KV_CLUSTERS_PATH string = "trento/v0/clusters"
```
arbulu89, line 14: Just to comment on this, I started changing to camel case the constants, as you started because it looks like it is more golang thing. Just a comment hehe. 
We should definitely pick one definitive option

agent/discover/cluster.go
```diff
@@ -0,0 +1,91 @@
...
+const crm_monPath string = "/usr/sbin/crm_mon"
+const cib_admPath string = "/usr/bin/cibadmin"
+
+//const crm_monPath string = "./test/fake/crm_mon.sh"
+//const cib_admPath string = "./test/fake_cibadmin.sh"
+
+// This Discover handles any Pacemaker Cluster type
+type ClusterDiscover struct {
+	cibConfig      cib.Root
+	clusterName    string
```
arbulu89, line 25: As a comment, not all clusters have a name. I don't know if this impacts somehow the current implementation

agent/discover/discover.go
```diff
@@ -0,0 +1,31 @@
...
+	// is relevant for this node. It is used as a gating condition for other
+	// functionality in this implementation
+	shouldDiscover(client consul.Client) bool
+	// Provide a Consul Key-Value Store Prefix that is used to store individual
+	// Key-Values for this Cluster
+	KVStorePrefix() string
+}
+
+// Create or Updating the given Consul Key-Value Path Store with a new value from the Agent
+func StoreDiscovery(discover Discoverer, client consul.Client, cStorePath, cStoreValue string) {
```
arbulu89, line 23: This looks a pretty generic function that will be used outside of discoveries most probably

agent/discover/hanacluster.go
```diff
@@ -0,0 +1,53 @@
...
+	return fmt.Sprintf(
+		"%s/%s",
+		consul_internal.KV_CLUSTERS_PATH, hana.cluster.clusterName)
+}
+
+// Discovery module handling specifics for SAP Hana clusters
+func NewHanaClusterDiscover(client consul.Client) Discovery {
+	return func() (Discoverer, error) {
+		r := HanaClusterDiscover{}
+		//r.cluster = NewClusterDiscover(client)(client)
```
arbulu89, line 34: Don't we need to uncomment this? Otherwise how is the cluster code initialized?

agent/discover/cluster.go
```diff
@@ -0,0 +1,91 @@
...
+		for _, primitive := range r.cibConfig.Configuration.Resources.Primitives {
+			switch primitive.Type {
+			case "external/sbd":
+				r.stonithType = consul_internal.CLUSTER_STONITH_SBD
+			case "IPaddr2":
+				if !foundVIP && primitive.Provider == "heartbeat" {
+					for _, attr := range primitive.InstanceAttributes {
+						switch attr.Name {
+						case "ip":
+							StoreDiscovery(r, client, "VIP_primary", string(attr.Value))
```
arbulu89, line 77: I wouldn't store this information this way. I think we should store all the resources entries in a single KV entrypoint as json, so we can just read everything in one singles read.
Anyway, it is fine by now. We can evolve in the future if needed.

arbulu89 (dismissed):
Hi @dirkmueller , In general it looks good. I have commented many things, but I don't think we need to change them now neccessarily.

The unique important things that I think the PR must have is the line with:
```
r.cluster = NewClusterDiscover(client)(client)
```
Shouldn't this be uncommented?

agent/discover/cluster.go
dirkmueller: the clusterName can be empty as well. it simply is supposed to show what was configured. there is no reliance on the content being non-empty afaik.

agent/agent.go
dirkmueller: Yes, it helps troubleshooting discovery failures. for now there isn't a fail that is being reported, but it would make sense to use the framework for that going forward.

agent/agent.go
dirkmueller: Thats a possibility.

internal/consul/kvstore.go
dirkmueller: well, don't really care which style we use. I changed it to all upper case based on previous review-feedback.

agent/discover/discover.go
dirkmueller: no, it uses the Discoverer Interface implementations, so it isn't useable outside discoveries.

agent/discover/cluster.go
dirkmueller: there is a size limit on kv store values, so storing everything as "json" under a single key is not working either. 

As far as I know you can fetch all key value pairs in one call from the kvstore, and filter then in your view. so that approach should be working. if it is for some reason too slow or too cumbersome we can change it.

arbulu89: Well, the limitation is `512kb`. In any case, you are right, you can all elements from a prefix in an individual call. 
Up to you. I though that it would be much easier to marshall and unmarshall things as json, but if you think that the size limitation impacts, we cannot use it.

stefanotorresi: @arbulu89 nah, let's avoid putting json inside a json. I tried that with the checker result output, and it blew up on me royally! 😁

internal/consul/kvstore.go
stefanotorresi: let's stick with `camelCase`.

https://stackoverflow.com/a/22688926/1488104

We should add this as a note in the contributing documentation.

agent/agent.go
```diff
@@ -199,6 +225,34 @@ func (a *Agent) startCheckTicker() error {
...
+		case <-ticker.C:
+			for i := 0; i < len(a.discovers); i++ {
+				_, err := a.discovers[i]()
+				if err != nil {
+					log.Println("Error while discovering: ", err)
+				}
+			}
+			a.consul.Agent().UpdateTTL("discover_cluster", "", consul.HealthPassing)
+		case <-a.ctx.Done():
+			close(a.wsResultChan)
```
stefanotorresi, line 250: I think this is a leftover from the other ticker. We shouldn't close this channel here.

stefanotorresi (changes requested):

agent/agent.go
```diff
@@ -8,30 +8,33 @@ import (
...
 	consul "github.com/hashicorp/consul/api"
 	"github.com/pkg/errors"
-	"github.com/hashicorp/consul-template/manager"
+	"github.com/trento-project/trento/agent/discover"
 )
 
 type Agent struct {
 	cfg              Config
 	check            Checker
+	discovers        []discover.Discovery
```
stefanotorresi, line 20: little typo
```suggestion
	discoveries        []discover.Discovery
```

agent/agent.go
```diff
@@ -199,6 +225,34 @@ func (a *Agent) startCheckTicker() error {
...
+	// Repeat after 15 minutes
+	if a.cfg.DiscoverInterval < time.Minute {
+		a.cfg.DiscoverInterval = 15 * time.Minute
+	}
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ticker.C:
+			for i := 0; i < len(a.discovers); i++ {
+				_, err := a.discovers[i]()
```
stefanotorresi, line 243: Same typo as above:
```suggestion
			for i := 0; i < len(a.discoveries); i++ {
				_, err := a.discoveries[i]()
```

agent/discover/hanacluster.go
```diff
@@ -0,0 +1,53 @@
...
+
+// Provide a Consul Key-Value Store Prefix that is used to store individual Key-Values for this Cluster
+func (hana HanaClusterDiscover) KVStorePrefix() string {
+	return fmt.Sprintf(
+		"%s/%s",
+		consul_internal.KV_CLUSTERS_PATH, hana.cluster.clusterName)
+}
+
+// Discovery module handling specifics for SAP Hana clusters
+func NewHanaClusterDiscover(client consul.Client) Discovery {
```
stefanotorresi, line 31: The name of this constructor is quite misleading: it would suggest that it would return a `HanaClusterDiscover` type, but it actually returns a `Discovery` function. 

By trying to understand how these two types are composed together, it appears to me that creating any `Discovery` type will need some `Discoverer` interface, so maybe we actually need a companion constructor to build the `Discovery` function by injecting a `Discoverer` in it?

All that said, I would go back to my initial suggestion to maybe collapse the `Discovery` function type into the `Discoverer` interface as a method, may this would simplify things a bit?
To be honest, I still don't quite get why we have two different types instead of one single `Discovery` interface, with something like a `Discovery.run()` method invoked by the Agent loop; am I missing something?

agent/agent.go
dirkmueller: changed that (together with the go fmt fallout)

agent/agent.go
dirkmueller: changed.

agent/agent.go
dirkmueller: removed

agent/discover/hanacluster.go
dirkmueller: I was following a guideline on the web on how to implement polymorphism using closures in Golang. I've now changed it as you requested. it is significantly more code duplication, but hopefully it'll get the PR merged then to be able to move forward.

agent/discover/hanacluster.go
dirkmueller: well, it wasn't used. now it is, so I removed the comment and it is initialized now.

internal/consul/kvstore.go
dirkmueller: added a `CONTRIBUTING.md` and referenced Effective Go in there.

stefanotorresi (changes requested):

agent/discover/discover.go
```diff
@@ -0,0 +1,52 @@
...
+// Execute one iteration of a discovery and store the result in the Consul KVStore.
+func (discover Discover) Discover() error {
+	return nil
+}
+
+// Return a Host Discover instance
+func NewDiscover(client consul.Client) Discover {
+	r := Discover{}
+	r.client = client
+	r.host, _ = os.Hostname()
```
stefanotorresi, line 50: hmm, shouldn't this be done in the `Discover` function? Not that it's likely for the hostname to change, but still, it could happen without a reboot, right?

stefanotorresi (dismissed):
Okay, let's merge it! Just take my last comments as ideas for what follows; I wasn't able to merge this PR and integrate it in a running environment, but I don't want to stall it any further, so I'll let you carry on with the next iterations.

agent/discover/cluster.go
```diff
@@ -0,0 +1,102 @@
...
+	if err != nil {
+		return err
+	}
+	cluster.stonithType = consul_internal.ClusterStonithUnknown
+
+	cibParser := cib.NewCibAdminParser(cib_admPath)
+
+	cibConfig, err := cibParser.Parse()
+	if err != nil {
+		log.Printf("Failing to parse: %s", errors.Wrap(err, "crm_mon parser error"))
```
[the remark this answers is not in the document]
dirkmueller: I added a return and fixed the error message. I agree that all the error handling logic is half baked, I need to give it another pass.

agent/discover/discover.go
dirkmueller: I duplicated it into the Discover() function as well, as it was empty so far. I need it set on construction just to make sure that storeDiscovery() calls are ending up in the right spot.

PR #30, opened 2021-04-20
azure-rules: Add check for os-release

aleksei-burlakov:
According to the https://jira.suse.com/browse/CFSA-83 check that 
- the SLES4SAP OS version is supported for the solution/scenario
- the nodes are SLES4SAP (and not something else)

### DIFF
--- a/examples/azure-rules-perf-optimized.yaml
+++ b/examples/azure-rules-perf-optimized.yaml
@@ -1,0 +1,21 @@
+---
+controls:
+version: 1.0.0
+id: 2
+description: "HA Checker for SAPHanaSR Scale-up Perf-Optimized"
+type: "master"
+groups:
+  - id: 2.1
+    description: "Check OS and package versions are supported and up-to-date"
+    checks:
+      - id: 2.1.1
+        description: "Check the base product"
+        # Assume the perf-optimized scenario and
+        # check only that the base probuct is always sles_sap.prod (case-insensitively)
+        audit: 'readlink /etc/products.d/baseproduct | sed -e "s/\(.*\)/\L\1/"'
+        tests:
+          test_items:
+            - flag: 'sles_sap.prod'
+        remediation: |
+          Install a SUSE supported OS release
+        scored: true


examples/azure-rules.yaml
```diff
@@ -656,4 +656,21 @@ groups:
           Please, check the naming convention in case the resource configuration is done correctly.
            
           More information at https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/sap-hana-high-availability#create-sap-hana-cluster-resources
+        scored: true
+  - id: 1.6
+    description: "Check OS and package versions are supported and up-to-date"
```
dirkmueller, line 661: well this is only checking "supported" not "up to date"

examples/azure-rules.yaml
```diff
@@ -656,4 +656,21 @@ groups:
...
+        tests:
+          bin_op: or
+          test_items:
+            - flag: 'PRETTY_NAME="SUSE Linux Enterprise Server for SAP Applications 12 SP4"'
+            - flag: 'PRETTY_NAME="SUSE Linux Enterprise Server for SAP Applications 15"'
+            - flag: 'PRETTY_NAME="SUSE Linux Enterprise Server for SAP Applications 15 SP1"'
+            - flag: 'PRETTY_NAME="SUSE Linux Enterprise Server for SAP Applications 15 SP2"'
+            - flag: 'PRETTY_NAME="SUSE Linux Enterprise Server for SAP Applications 15 SP3"'
+        remediation: |
+          Install a proper OS release
```
dirkmueller, line 675: "proper" is maybe a bit harsh. how about "SUSE Supported OS release" ?

diegoakechi: I think this check will fail. SLES for SAP and SLES have the same content on os-release for historical reasons and certification matters. Both state SLES on it. if we need to check the commercial product, the /etc/products.d/baseproduct file should be used instead.

Also, SLES is a certified OS for SAP, so the check should consider it too.

diegoakechi (changes requested):

examples/azure-rules.yaml
aleksei-burlakov: It's a description for the id 1.6, the 1.6.1 checks only the supported os release, there will come 1.6.2 and 1.6.3 in #31  that check that the packages are up-to-date.

examples/azure-rules.yaml
aleksei-burlakov: Thaks! I changed "proper" -> "SUSE supported". The /etc/os-release indeed should never be used for examining the base product (https://www.suse.com/support/kb/doc/?id=000019341). I check it no by the /etc/products.d/baseproduct

examples/azure-rules.yaml
```diff
@@ -656,4 +656,16 @@ groups:
...
+        scored: true
+  - id: 1.6
+    description: "Check OS and package versions are supported and up-to-date"
+    checks:
+      - id: 1.6.1
+        description: "Check the base product"
+        audit: 'ls -la /etc/products.d/baseproduct | sed "s/^.*-> //"'
+        tests:
+          test_items:
+            - flag: 'SLES_SAP.prod'
```
diegoakechi, line 668: @aleksei-burlakov  As I have commented before, SLES is also a valid option for SAP Applications, so we should also consider vanilla SLES as a valid OS. 

On the remediation text, we need something more meaningful, but we can include this later, with the support from PM.

diegoakechi (changes requested):

examples/azure-rules.yaml
diegoakechi: @lee-martin I think we should also link on the remediation text the SAP Note of supported OS.

aleksei-burlakov: As they say in https://www.suse.com/support/kb/doc/?id=000019341

_Even when the base product is SLES for SAP, the contents of the os-release file should never have been used to identify this fact._

I understand we should only look in the `/etc/products.d/baseproduct`. I check that it must be SLES_SAP.prod. If it could also be SLES.prod then as I understand I need to check the certificates instead?

lee-martin: > @aleksei-burlakov As I have commented before, SLES is also a valid option for SAP Applications, so we should also consider vanilla SLES as a valid OS.
> 
> On the remediation text, we need something more meaningful, but we can include this later, with the support from PM.

@diegoakechi @aleksei-burlakov  SLES was intentionally not a valid option here. This check is designed to check of the OS used supports the SUSE HA Solution " SAPHanaSR Scale-up Perf-Optimized" , and that is only true for SLES4SAP.
The question about which OS is valid for SAP HANA is a different discussion and would be a different requirement. For now we are focusing on the SUSE HA context only. Hence I would not consider Vanilla SLES since this does not support SAPHanaSR.

lee-martin: > As they say in https://www.suse.com/support/kb/doc/?id=000019341
> 
> _Even when the base product is SLES for SAP, the contents of the os-release file should never have been used to identify this fact._
> 
> I understand we should only look in the `/etc/products.d/baseproduct`. I check that it must be SLES_SAP.prod. If it could also be SLES.prod then as I understand I need to check the certificates instead?

@aleksei-burlakov  You need a way to distinguish between SLES and SLES4SAP; As I commented above, the intention of this check is to be sure the OS supports the HA solution SAPHanaSR  .

lee-martin: > @lee-martin I think we should also link on the remediation text the SAP Note of supported OS.

@diegoakechi I'm not really keen on SUSE making statements about what SAP supports, at some point we could add a text and link pointing customers to the relevant SAP notes, but for now I'd focus on the SUSE aspects.

diegoakechi: @lee-martin ok, so please propose the right terminology and text for the check itself and it's remediation, because it is really hard for a user understand what is being checked here.

examples/azure-rules.yaml
dirkmueller: ok got it. I was already confused why these are two PRs rather than one iwth multiple commits. fine by me.

examples/azure-rules.yaml
```diff
@@ -656,4 +656,18 @@ groups:
           Please, check the naming convention in case the resource configuration is done correctly.
            
           More information at https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/sap-hana-high-availability#create-sap-hana-cluster-resources
+        scored: true
+  - id: 1.6
+    description: "Check OS and package versions are supported and up-to-date"
+    checks:
+      - id: 1.6.1
+        description: "Check the base product"
+        audit: 'ls -la /etc/products.d/baseproduct | sed "s/^.*-> //"'
```
dirkmueller, line 665: this is scary code, better do something like 

`readlink /etc/products.d/baseproduct`

examples/azure-rules.yaml
aleksei-burlakov: > > As they say in https://www.suse.com/support/kb/doc/?id=000019341
> > _Even when the base product is SLES for SAP, the contents of the os-release file should never have been used to identify this fact._
> > I understand we should only look in the `/etc/products.d/baseproduct`. I check that it must be SLES_SAP.prod. If it could also be SLES.prod then as I understand I need to check the certificates instead?
> 
> @aleksei-burlakov You need a way to distinguish between SLES and SLES4SAP; As I commented above, the intention of this check is to be sure the OS supports the HA solution SAPHanaSR .

@lee-martin Do I understand it correctly, if the base product is the SLES_SAP.prod, it automatically supports SAPHanaSR, and no further checks needed. If it's only SLES.prog, we also need to check for support of the SAPHanaSR? If not why, then in which way I should distinguish them?

stefanotorresi: clarified for @aleksei-burlakov that this check is specifically about the "HANA Scale-Out Performance Optimized" Scenario, so it must go in a different ruleset than the "Azure Generic" one.

aleksei-burlakov: > > > As they say in https://www.suse.com/support/kb/doc/?id=000019341
> > > _Even when the base product is SLES for SAP, the contents of the os-release file should never have been used to identify this fact._
> > > I understand we should only look in the `/etc/products.d/baseproduct`. I check that it must be SLES_SAP.prod. If it could also be SLES.prod then as I understand I need to check the certificates instead?
> > 
> > 
> > @aleksei-burlakov You need a way to distinguish between SLES and SLES4SAP; As I commented above, the intention of this check is to be sure the OS supports the HA solution SAPHanaSR .
> 
> @lee-martin Do I understand it correctly, if the base product is the SLES_SAP.prod, it automatically supports SAPHanaSR, and no further checks needed. If it's only SLES.prog, we also need to check for support of the SAPHanaSR? If not why, then in which way I should distinguish them?

@lee-martin, @diegoakechi let's make it simple and assume that this yaml file is for performance-optimized scenario and check that the base-product is strictly _sles_sap.prod_ (case insensitively).

aleksei-burlakov: @here It's a small PR and it breaks the next #31, let's merge it as is and we will add other checks later.

stefanotorresi: hmm, @aleksei-burlakov this is not what we just agreed. I left you with the indication to create a separate rule set specific for the "HANA ScaleOut Perfomance Optimized" scenario. Are there other rules we already implemented that are specific to that scenario alone, and are not generic?

aleksei-burlakov: > hmm, @aleksei-burlakov this is not what we just agreed. I left you with the indication to create a separate rule set specific for the "HANA ScaleOut Perfomance Optimized" scenario. Are there other rules we already implemented that are specific to that scenario alone, and are not generic?

I thought I would do it already another PR. I have amended this PR and created a new yaml file that checks only for SAPHanaSR Scale-up Perf-Optimized scenario.

examples/azure-rules.yaml
dirkmueller: Thanks!

examples/azure-rules-perf-optimized.yaml
```diff
@@ -0,0 +1,21 @@
...
+type: "master"
+groups:
+  - id: 2.1
+    description: "Check OS and package versions are supported and up-to-date"
+    checks:
+      - id: 2.1.1
+        description: "Check the base product"
+        # Assume the perf-optimized scenario and
+        # check only that the base probuct is always sles_sap.prod (case-insensitively)
+        audit: 'readlink /etc/products.d/baseproduct | sed -e "s/\(.*\)/\L\1/"'
```
dirkmueller, line 15: I think the `| sed ...` is  unnecessary here.

dirkmueller (dismissed):
its okay for me, the points from diego about SLES and versioning checks are still valid, but we can do that in the other pull request. I suggest to merge this for now..

examples/azure-rules-perf-optimized.yaml
aleksei-burlakov: I was told that the base product could be in both lower- and upper-case.

dirkmueller: we're reading the target of the symlink here, and we assume case sensitive filesystems. the check by itself is not very precise anyway, as we're just checking for filename not filecontents (and in the products.d specification, filename doesn't matter, file content is all that matters). 

then we need to get back to checking the action product version and name again, like suggested earlier.

dirkmueller (approved):
lets move forward here..

PR #31, opened 2021-04-20
Check package versions

aleksei-burlakov:
This PR as a continuation of #30 it adds checks for package versions of pacemaker, corosync, and sbd.

### DIFF
--- a/examples/azure-rules-perf-optimized.yaml
+++ b/examples/azure-rules-perf-optimized.yaml
@@ -19,3 +19,74 @@ groups:
         remediation: |
           Install a SUSE supported OS release
         scored: true
+      - id: 2.1.2
+        description: "Check the pacemaker version IS"
+        audit: 'rpm -qv pacemaker | sed -e "s/pacemaker-\(.*\)+.*/\1/"'
+        tests:
+          bin_op: or
+          test_items:
+            - flag: '2.0.3'
+        remediation: |
+          Update pacemaker packages
+        scored: true
+      - id: 2.1.2.exclude
+        description: "Check the pacemaker version IS NOT"
+        audit: 'rpm -q --qf "%{VERSION}\n" pacemaker'
+        tests:
+          bin_op: and
+          test_items:
+          - flag: '2.0.3+20200511.2b248d828'
+            compare:
+              op: noteq
+              value: '2.0.3+20200511.2b248d828'
+            set: true
+        remediation: |
+          Update pacemaker packages
+        scored: true
+      - id: 2.1.3
+        description: "Check the corosync version IS"
+        audit: 'rpm -qv corosync'
+        tests:
+          bin_op: or
+          test_items:
+            - flag: 'corosync-2.4.5-5.8.x86_64'
+        remediation: |
+          Update corosync packages
+        scored: true
+      - id: 2.1.4
+        description: "Check the sbd version IS"
+        audit: 'rpm -qv sbd | sed -e "s/sbd-\(.*\)+.*/\1/"'
+        tests:
+          bin_op: or
+          test_items:
+            - flag: '1.4.0'
+        remediation: |
+          Update sbd packages
+        scored: true
+      - id: 2.1.4.exclude
+        description: "Check the sbd version IS NOT"
+        audit: 'rpm -q --qf "%{VERSION}\n" sbd'
+        tests:
+          bin_op: and
+          test_items:
+          - flag: '1.4.0+20190326.c38c5e6'
+            compare:
+              op: noteq
+              value: '1.4.0+20190326.c38c5e6'
+            set: true
+        remediation: |
+          Update sbd packages
+        scored: true
+      - id: 2.1.5
+        description: "Check python version"
+        audit: 'python3 --version'
+        tests:
+          bin_op: or
+          test_items:
+            - flag: 'Python 3.6.5'
+            - flag: 'Python 3.6.6'
+            - flag: 'Python 3.6.7'
+            - flag: 'Python 3.6.8'
+        remediation: |
+          Update python packages
+        scored: true


examples/azure-rules.yaml
```diff
@@ -656,4 +656,51 @@ groups:
...
+            - flag: 'PRETTY_NAME="SUSE Linux Enterprise Server for SAP Applications 15"'
+            - flag: 'PRETTY_NAME="SUSE Linux Enterprise Server for SAP Applications 15 SP1"'
+            - flag: 'PRETTY_NAME="SUSE Linux Enterprise Server for SAP Applications 15 SP2"'
+            - flag: 'PRETTY_NAME="SUSE Linux Enterprise Server for SAP Applications 15 SP3"'
+        remediation: |
+          Install a proper OS release
+        scored: true
+      - id: 1.6.2
+        description: "Check pacemaker version"
+        audit: 'rpm -qv pacemaker | sed -e "s/pacemaker-\(.*\)+.*/\1/"'
```
dirkmueller, line 679: you could do something like 

`rpm -q --qf "%{VERSION}\n" pacemaker`

instead

examples/azure-rules.yaml
```diff
@@ -656,4 +656,51 @@ groups:
           Please, check the naming convention in case the resource configuration is done correctly.
            
           More information at https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/sap-hana-high-availability#create-sap-hana-cluster-resources
+        scored: true
+  - id: 1.6
+    description: "Check OS and package versions are supported and up-to-date"
+    checks:
+      - id: 1.6.1
+        description: "Check os release"
+        audit: 'cat /etc/os-release'
```
dirkmueller, line 665: how does that work? does this then only match this in *any* line or in the first line?

can we do something lik e

`. /etc/os-release ; echo $NAME-$VERSION_ID`

examples/azure-rules.yaml
```diff
@@ -656,4 +656,51 @@ groups:
...
+        audit: 'rpm -qv pacemaker | sed -e "s/pacemaker-\(.*\)+.*/\1/"'
+        tests:
+          bin_op: or
+          test_items:
+            - flag: '2.0.3'
+        remediation: |
+          Update the pacemaker
+        scored: true
+      - id: 1.6.3
+        description: "Check crosync version"
```
dirkmueller, line 688: typo: corosync

examples/azure-rules.yaml
```diff
@@ -656,4 +656,51 @@ groups:
...
+        tests:
+          bin_op: or
+          test_items:
+            - flag: 'corosync-2.4.5-5.8.x86_64'
+        remediation: |
+          Update the corosync
+        scored: true
+      - id: 1.6.4
+        description: "Check sbd version"
+        audit: 'rpm -qv sbd | sed -e "s/sbd-\(.*\)+.*/\1/"'
```
dirkmueller, line 699: `rpm -q --qf "%{VERSION}\n" sbd`

examples/azure-rules.yaml
aleksei-burlakov: Thanks! I don't know if the `rpm -q --qf "%{VERSION}\n"` already too specific because it also returns the commit hex. I replaced the rules anyway.

examples/azure-rules.yaml
aleksei-burlakov: It didn't work for me when I did `source /etc/release && echo $PRETTY_NAME`, so I did it in the uglier way. There is an `bin_op: or`, so it checks if any of the flags is in the /etc/os-release.

examples/azure-rules.yaml
aleksei-burlakov: Thanks, fixed.

examples/azure-rules.yaml
```diff
@@ -656,4 +656,51 @@ groups:
...
+        remediation: |
+          Install a proper OS release
+        scored: true
+      - id: 1.6.2
+        description: "Check pacemaker version"
+        audit: 'rpm -q --qf "%{VERSION}\n" pacemaker'
+        tests:
+          bin_op: or
+          test_items:
+            - flag: '2.0.3+20200511.2b248d828'
```
diegoakechi, line 683: @aleksei-burlakov This looks a bit too restrictive. we should check some minimal version, not a specific one. As soon some maintenance release occurs, the check will start to fail.

diegoakechi (changes requested):

examples/azure-rules.yaml
diegoakechi: Same for all the packages version checks.

aleksei-burlakov: @diegoakechi  there is a bin_op: or option, so we could simply add a new flag each time a new commit comes. Or I could bring it back to check only the '2.0.3' substring. What do you think?

diegoakechi: @aleksei-burlakov  changing this for every maintenance is error-prone and will be a burden. these checks should only occur to guarantee that system is not using a version known of causing problems. For example, check the blue box documented on the best practices: 

https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/sap-hana-high-availability#create-sap-hana-cluster-resources

examples/azure-rules.yaml
```diff
@@ -656,4 +656,51 @@ groups:
...
+        remediation: |
+          Update the pacemaker
+        scored: true
+      - id: 1.6.3
+        description: "Check crosync version"
+        audit: 'rpm -qv corosync'
+        tests:
+          bin_op: or
+          test_items:
+            - flag: 'corosync-2.4.5-5.8.x86_64'
```
dirkmueller, line 693: is there a "not in" combination? that would be more stable for the future, by simply listing what are *outdated* versions that we do not want to see. otherwise the next mainteannce update will break this check..

examples/azure-rules.yaml
dirkmueller: it reappeared

examples/azure-rules.yaml
```diff
@@ -656,4 +656,51 @@ groups:
...
+        remediation: |
+          Update the corosync
```
dirkmueller, line 695: for me "Update corosync packages" reads a bit nicer.

dirkmueller:
so I can see that it is valid to have some "boilerplate" OS  checks, but maybe we could focus on known bad versions that we just want to blacklist for now? maybe there are specific kernel versions that we know are broken in some form of fashion and we just check for those being *not* *installed* ?

aleksei-burlakov:
> so I can see that it is valid to have some "boilerplate" OS checks, but maybe we could focus on known bad versions that we just want to blacklist for now? maybe there are specific kernel versions that we know are broken in some form of fashion and we just check for those being _not_ _installed_ ?

This is a good idea. I will add a rule, that the versions must not be xyz.

examples/azure-rules.yaml
aleksei-burlakov: Done.

examples/azure-rules.yaml
aleksei-burlakov: I have added excluding rules for the pacemaker and sbd, but they exclude now valid releases, pacemaker-2.0.3+20200511.2b248d828 and sbd-1.4.0+20190326.c38c5e6.

aleksei-burlakov:
I have moved the rules to another file examples/azure-rules-perf-optimized.yaml introduced in #30
This commit also includes the https://github.com/trento-project/trento/pull/30/commits/284f4c41a1d6beb6589c3811d59923e7429edc40 , so that when the #30 is merged, this PR would have to be rebased. If those version checks are rather generic I will bring them back into the examples/azure-rules.yaml

examples/azure-rules-perf-optimized.yaml
```diff
@@ -0,0 +1,92 @@
...
+          - flag: '2.0.3+20200511.2b248d828'
+            compare:
+              op: noteq
+              value: '2.0.3+20200511.2b248d828'
+            set: true
+        remediation: |
+          Update pacemaker packages
+        scored: true
+      - id: 2.1.3
+        description: "Check the crosync version IS"
```
dirkmueller, line 47: typo: corosync

aleksei-burlakov: Thanks, fixed.

dirkmueller:
@aleksei-burlakov  please rebase.

aleksei-burlakov:
> @aleksei-burlakov please rebase.

Done!

stefanotorresi:
Reviewed live with aleksei, merging.

PR #32, opened 2021-04-22
Rename current environments to hosts

arbulu89:
Rename of `environments` to `nodes.

The unique technical change is that I have removed the `datacenters` usage as environment as it didn't make much sense after deciding that they don't match 1 to 1. I have removed the datacenter column visualization from the node table too as it doesn't give any valuable information.

### DIFF
--- a/web/app.go
+++ b/web/app.go
@@ -52,9 +52,9 @@ func NewAppWithDeps(host string, port int, deps Dependencies) (*App, error) {
 	engine.Use(ErrorHandler)
 	engine.StaticFS("/static", http.FS(assetsFS))
 	engine.GET("/", HomeHandler)
-	engine.GET("/environments", NewEnvironmentsListHandler(deps.consul))
-	engine.GET("/environments/:name", NewEnvironmentHandler(deps.consul))
-	engine.GET("/environments/:name/checks/:checkid", NewCheckHandler(deps.consul))
+	engine.GET("/nodes", NewNodesListHandler(deps.consul))
+	engine.GET("/nodes/:name", NewNodeHandler(deps.consul))
+	engine.GET("/nodes/:name/checks/:checkid", NewCheckHandler(deps.consul))
 	engine.GET("/clusters", NewClustersListHandler(deps.consul))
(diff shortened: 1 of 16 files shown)


stefanotorresi (approved):
excellent!

dirkmueller (commented):
according to https://confluence.suse.com/display/SAP/SUSE+Console+for+SAP+-+User+Stories#SUSEConsoleforSAPUserStories-Terminology the correct terminology is "host" not "node". why is node more correct?

lee-martin:
@dirkmueller  @arbulu89 
> according to https://confluence.suse.com/display/SAP/SUSE+Console+for+SAP+-+User+Stories#SUSEConsoleforSAPUserStories-Terminology the correct terminology is "host" not "node". why is node more correct?

I think the word "node" just came up spontaneously in the Trento review this week, you are indeed correct that "host" would be the better word to use, so actually we should "Rename current environments to hosts".

We also need to consider that this screen will at some point show non-clustered hosts, and "node" would sound strange in that meaning.  Also in SAP there ca be multiple "java nodes" on a single OS (host).

At the end of the day, this is just a word and we'll collect user feedback over time. Any choice will be a compromise, so lets just be consistent with ourselves and stick to "host" for now.  We've also been having a discussion around what a "system" is as well.

dirkmueller (changes requested):
Please do another search&replace to "Hosts".

arbulu89:
> according to https://confluence.suse.com/display/SAP/SUSE+Console+for+SAP+-+User+Stories#SUSEConsoleforSAPUserStories-Terminology the correct terminology is "host" not "node". why is node more correct?

You are right @dirkmueller . I renamed to `Nodes` based on some discussion we had in a meeting, and because consul uses this nomenclature. We can rename to `Hosts` if @stefanotorresi @diegoakechi

stefanotorresi:
Yes, `Host` is the actual term to use, my bad!

stefanotorresi (approved):

dirkmueller (approved):

PR #33, opened 2021-04-23
feature: agent: several yaml files can be specified

aleksei-burlakov:
Previously there must have been exactly one yaml file. Now it should be
at least one.

The question to duscuss is how to update the consul check (see agent/agent.go:195).
It's current implementation is that it's been sequentially updated for each yaml file.

### DIFF
--- a/agent/checker.go
+++ b/agent/checker.go
@@ -4,7 +4,9 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"log"
 	"strconv"
+	"strings"
 
 	"github.com/aquasecurity/bench-common/check"
 	"github.com/pkg/errors"
@@ -12,15 +14,46 @@ import (
 
 type Checker func() (CheckResult, error)
 
-func NewChecker(definitionsPath string) (Checker, error) {
-	data, err := ioutil.ReadFile(definitionsPath)
-	if err != nil {
-		return nil, errors.Wrap(err, "could not read definitions file")
+/* Remove
+  '---
+   controls:
+	version: ...
+	id: ...
+	description: ...
+	type: ...
+	groups:
+'
+that comes before '    - id:...'
+You need to do it to append the yaml to another yaml file
+*/
+func trim_yaml_header(yaml_text string) string {
+	pattern := "\ngroups:"
+	index := strings.Index(yaml_text, pattern)
+	if index == -1 {
+		log.Fatal("the yaml file doesn't contain the 'id:' field and is possibly broken")
 	}
+	index += len(pattern)
+	return yaml_text[index:]
+}
 
+func NewChecker(definitionsPaths []string) (Checker, error) {
+	var data [][]byte
+	for _, path := range definitionsPaths {
+		datum, err := ioutil.ReadFile(path)
+		if err != nil {
+			return nil, errors.Wrap(err, "could not read definitions file")
+		}
+		data = append(data, datum)
+	}
 	return func() (CheckResult, error) {
-		result := CheckResult{}
-		controls, err := check.NewControls(data, nil)
+		var result CheckResult
+		first_yaml_file := string(data[0])
+		// Appent all next files to the first
+		for _, datum := range data[1:] {
+			first_yaml_file += trim_yaml_header(string(datum))
+		}
+
+		controls, err := check.NewControls([]byte(first_yaml_file), nil)
 		if err != nil {
 			return result, errors.Wrap(err, "could not parse definitions file")
 		}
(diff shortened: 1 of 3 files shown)


stefanotorresi (changes requested):
@aleksei-burlakov there is a design oversight in this implementation, which I now realise it's not as easy as we thought:

We cannot simply scroll through the results and push them out one by one, otherwise the previous will always be immediately  rewritten and lost.

This happens in two different ways: 
- when we send results to the webservice (`a.wsResultChan <- res`)
- when we update the TTL on the HA check (`a.updateConsulCheck(res)`) 

Both of these operations are intended to be atomic, and cannot be incremental:

- The way the `wsResultChan` is consumed by the webservice is that it just takes the last result from the channel and displays it; we want to instead display them all into a single structure.
- The way the consul check works is that the status of the last TTL update is stored.

The operation we're trying to do is not a queue processing, but rather a demultiplexing of many different outputs into a single one.

To do this, you can either combine the yaml files into a single one when parsing them, or change the `CheckResult` struct so that you can combine multiple instances of the `Controls` struct into a single `CheckResult` instance.

aleksei-burlakov:
> @aleksei-burlakov there is a design oversight in this implementation, which I now realise it's not as easy as we thought:
> 
> We cannot simply scroll through the results and push them out one by one, otherwise the previous will always be immediately rewritten and lost.
> 
> This happens in two different ways:
> 
> * when we send results to the webservice (`a.wsResultChan <- res`)
> * when we update the TTL on the HA check (`a.updateConsulCheck(res)`)
> 
> Both of these operations are intended to be atomic, and cannot be incremental:
> 
> * The way the `wsResultChan` is consumed by the webservice is that it just takes the last result from the channel and displays it; we want to instead display them all, via a single JSON payload.
> * The way the consul check works is that the status of the last TTL update is stored. So we want to update the TTL once and report the status as failed/warning/success by combinning all the checker results together.
> 
> In other words, the operation we're trying to do is not a queue processing, but rather a demultiplexing of many different outputs into a single one.
> 
> To do this, you can either combine the yaml files into a single one when parsing them, or change the `CheckResult` struct so that you can combine the result of multiple instances of the `Controls` struct into a single `CheckResult` instance.

Ok, it's getting more interesting now)

aleksei-burlakov:
> @aleksei-burlakov there is a design oversight in this implementation, which I now realise it's not as easy as we thought:
> 
> We cannot simply scroll through the results and push them out one by one, otherwise the previous will always be immediately rewritten and lost.
> 
> This happens in two different ways:
> 
> * when we send results to the webservice (`a.wsResultChan <- res`)
> * when we update the TTL on the HA check (`a.updateConsulCheck(res)`)
> 
> Both of these operations are intended to be atomic, and cannot be incremental:
> 
> * The way the `wsResultChan` is consumed by the webservice is that it just takes the last result from the channel and displays it; we want to instead display them all, via a single JSON payload.
> * The way the consul check works is that the status of the last TTL update is stored. So we want to update the TTL once and report the status as failed/warning/success by combinning all the checker results together.
> 
> In other words, the operation we're trying to do is not a queue processing, but rather a demultiplexing of many different outputs into a single one.
> 
> To do this, you can either combine the yaml files into a single one when parsing them, or change the `CheckResult` struct so that you can combine the result of multiple instances of the `Controls` struct into a single `CheckResult` instance.

It combines several yamls into one now. 👌

stefanotorresi:
@aleksei-burlakov all right! What happens if two checks have the same id, though?

agent/checker.go
```diff
@@ -4,31 +4,61 @@ import (
...
+		log.Fatal("the yaml file doesn't contain the 'id:' field and is possibly broken")
 	}
+	index += len(pattern)
+	return yaml_text[index:]
+}
 
+func NewChecker(definitionsPaths []string) (Checker, error) {
+	var data [][]byte
+	for _, path := range definitionsPaths {
+		datum, err := ioutil.ReadFile(path)
```
dirkmueller, line 42: `datum` is not a perfect name, can we use something more descriptive?

agent/checker.go
```diff
@@ -4,31 +4,61 @@ import (
...
+		if err != nil {
+			return nil, errors.Wrap(err, "could not read definitions file")
```
dirkmueller, line 44: the error could include details about which file we had an issue with.

agent/checker.go
```diff
@@ -4,31 +4,61 @@ import (
...
+		data = append(data, datum)
+	}
 	return func() (CheckResult, error) {
-		result := CheckResult{}
-		controls, err := check.NewControls(data, nil)
+		var result CheckResult
+		first_yaml_file := string(data[0])
+		// Appent all next files to the first
+		for _, datum := range data[1:] {
+			first_yaml_file += trim_yaml_header(string(datum))
```
dirkmueller, line 53: interesting, so this is merging them textually into one yaml and then parse them in one go. what is the trade off over https://stackoverflow.com/questions/51947879/read-and-merge-two-yaml-files ?

is there a way to make the bench tool parse a list of strings/yamls instead of just one? I wasn't able to clearly find an answer on that in a cursory look at the code.

aleksei-burlakov:
> @aleksei-burlakov all right! What happens if two checks have the same id, though?

I havn't seen any collision.

dirkmueller:
> > @aleksei-burlakov all right! What happens if two checks have the same id, though?> 
> I havn't seen any collision.

right, but I think the wanted behavior is that a later (or an earlier file) is able to overwrite a check when it specifies the same id? which means you need to merge semantically not just append the input files together.

aleksei-burlakov:
> > > @aleksei-burlakov all right! What happens if two checks have the same id, though?>
> > > I havn't seen any collision.
> 
> right, but I think the wanted behavior is that a later (or an earlier file) is able to overwrite a check when it specifies the same id? which means you need to merge semantically not just append the input files together.

I was thinking of some kind of an object-oriented approach (overriding+ingheritance), but we decided with Stefano it would be too difficult in the beginning.

agent/checker.go
aleksei-burlakov: AFAI understand the https://stackoverflow.com/questions/51947879/read-and-merge-two-yaml-files merges files so that the properties from the next file overwrite those from the former. We have many description: properties and we don't want them to be overwritten.

I was unsuccessfully trying to parse only a half of a yaml file. It's possible the unparsed part will simply be lost. Though I could do something that does somewhat, but let's decide on Thursday about the requirements for the expressiveness of our cheks.

agent/checker.go
aleksei-burlakov: Fixed.

agent/checker.go
aleksei-burlakov: Done.

dirkmueller:
@aleksei-burlakov  please rebase

aleksei-burlakov:
> @aleksei-burlakov please rebase

Done!

aleksei-burlakov:
@stefanotorresi I've made the pretty-print just like in the bench-common.

dirkmueller (approved):
wfm

stefanotorresi (changes requested):
`make test` is failing with the following:
```
cmd/agent/agent.go:73:5: res.CheckPrettyPrint undefined (type agent.CheckResult has no field or method CheckPrettyPrint)
```

aleksei-burlakov:
I have rebased it it should work now.

dirkmueller (approved):

stefanotorresi (approved):
manually tested ok! We'll need to give this code some love in the future, though.

PR #34, opened 2021-04-26
New environments view

arbulu89:
Implementation of the new `Environments` entry point. This adds a new element in the sidebar that gives the option to move along the environments/landscapes/sap systems in hierarchy mode. Each of the elements summarizes the status of the whole entity.
Besides that, the individual elements show the group they belong too.

The usage of the KV storage is explained in the `README` file

Some pics:
image
image
image
image
image

FYI: @lee-martin

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,16 +96,17 @@ Notice that a new entry must exists for every node.
 - `config-dir`: Consul agent configuration files directory. It must be the same used by the consul agent. The `trento` agent creates a new folder with the node name where the trento meta-data configuration file is stored (e.g. `consul.d/node1/trento-config.json`).
 - `consul-template`: Template used to populate the trento meta-data configuration file (by default [meta-data file][./examples/trento-config.json] is used).
 
-### Filtering the nodes in the wep app
+### Grouping and filtering the nodes in the wep app
 
-The web app provides the option to filter the systems using the previously commented reserved tags. To achieve this, the tags must be stored in the KV storage.
+The app provides the option to see the environment composed by the nodes and filter the systems using the previously commented reserved tags. To achieve this, the tags must be stored in the KV storage.
 Use the next path:
-- `trento/filters/sap-environments`
-- `trento/filters/sap-landscapes`
-- `trento/filters/sap-systems`
+- `trento/environments/$yourenv/`
+- `trento/environments/$yourenv/landscapes/$yourland/`
+- `trento/environments/$yourenv/landscapes/$yourland/sapsystems/$yoursapsy`
 
-Each of them must have a json list format. As example: `["land1", "land2"]`.
-These entries will be available in the filters on the `/environments` page.
+Keep in mind that the created environments, landscapes and sap systems are directories themselves, and there can be multiple of them.
+The possibility to have multiple landscapes with the same name in different environments (and the same for SAP systems) is possible.
+Be aware that the nodes meta-data tags are not strictly linked to these names, they are soft relations (this means that only the string matches, there is no any real relationship between them).
 
 ## Development
 
--- a/internal/consul/consul.go
+++ b/internal/consul/consul.go
@@ -21,6 +21,7 @@ type Catalog interface {
 type KV interface {
 	Get(key string, q *consulApi.QueryOptions) (*consulApi.KVPair, *consulApi.QueryMeta, error)
 	List(prefix string, q *consulApi.QueryOptions) (consulApi.KVPairs, *consulApi.QueryMeta, error)
+	Keys(prefix, separator string, q *consulApi.QueryOptions) ([]string, *consulApi.QueryMeta, error)
 }
 
 type Health interface {
--- a/internal/consul/mocks/KV.go
+++ b/internal/consul/mocks/KV.go
@@ -45,6 +45,38 @@ func (_m *KV) Get(key string, q *api.QueryOptions) (*api.KVPair, *api.QueryMeta,
 	return r0, r1, r2
 }
 
+// Keys provides a mock function with given fields: prefix, separator, q
+func (_m *KV) Keys(prefix string, separator string, q *api.QueryOptions) ([]string, *api.QueryMeta, error) {
+	ret := _m.Called(prefix, separator, q)
+
+	var r0 []string
+	if rf, ok := ret.Get(0).(func(string, string, *api.QueryOptions) []string); ok {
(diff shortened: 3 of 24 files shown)


stefanotorresi:
I'm not quite sure about this. Shouldn't we go in steps? Why are we directly pointing them to environments, if we yet have to present them two smaller groupings?

arbulu89:
> I'm not quite sure about this. Shouldn't we go in steps? Why are we directly pointing them to environments, if we yet have to present them two smaller groupings?

I don't understand your question. The user story I was working on says:
```
As a User, one can view the cluster (incl. its resources and check results) in the context of how it fits into the overall SAP systems, landscapes and environments, so that the user can see and drill down to his systems from an SAP perspective.
```
So this implementation gives the option to find all the environments, and all the landscapes that belong to them, all the SAP systems that belong to each specific landscape.

Once you click in environments, this shows all the environments. If you click in some of the listed environments, this will lead you to the landscape list of this environment, and when you click in a specific SAP system, the SAP systems belonging to this landscape.
So, essentially this is a tree view (but you need to move along the pages back and forth).

stefanotorresi:
Sorry, I realise wasn't very clear: shouldn't we have a sidebar navigation item for each level of the various groupings? We have Hosts, Clusters and Environments now, but we're skipping direct access to Systems and Landscapes; why is that?

arbulu89:
> Sorry, I realise wasn't very clear: shouldn't we have a sidebar navigation item for each level of the various groupings? We have Hosts, Clusters and Environments now, but we're skipping direct access to Systems and Landscapes; why is that?

Yes... This is this way because right now we cannot move to a Landscape without specifying an environment. At the end, the task was to create a tree kind of hierarchy, and all landscape belong to an environment. So I cannot create a page of all the landscapes.

I could of course show all the landscapes and sap systems. But this would break the strict tree hierarchy, because we would show multiple landscapes/sap systems that are not filtered.

I see 2 potential solutions:
1. I just show all the landscapes (and the same for SAP systems), and we still continue having the `landscape must belong to a environment`. This way, if I click in the landscapes sidebar option it would show all the landscapes, without filtering with the environment (but the relation would still exist)
2. We completely remove this hard dependency. This requires major changes

Do we want to change this? @lee-martin @stefanotorresi 
After Stefano's comment, I like the 1st option.

stefanotorresi:
Yes, option 1 is exactly what I was hinting at: showing only one level of the tree hierarchy doesn't break the hierarchy itself, it's just a different view.
Drilling down from the Environments will still be possible, but sometimes users could have only one environment, or one landscape, and having dedicated pages for each level would allow them to skip directly to the level they are most interested in.

This is also for consistency with the bottom layers, Hosts and Clusters, which can be browsed independently already.

README.md
```diff
@@ -96,16 +96,17 @@ Notice that a new entry must exists for every node.
...
-- `trento/filters/sap-systems`
+- `trento/environments/$yourenv/`
+- `trento/environments/$yourenv/landscapes/$yourland/`
+- `trento/environments/$yourenv/landscapes/$yourland/sapsystems/$yoursapsy`
 
-Each of them must have a json list format. As example: `["land1", "land2"]`.
-These entries will be available in the filters on the `/environments` page.
+Keep in mind that the created environments, landscapes and sap systems are directories themselves, and there can be multiple of them.
+The possibility to have multiple landscapes with the same name in different environments (and the same for SAP systems) is possible.
+Be aware that the nodes meta-data tags are not strictly linked to these names, they are soft relations (this means that only the string matches, there is no any real relationship between them).
```
dirkmueller, line 109: this whole section should in some internal documentation in docs/ not in the README imho. README is end-user information, this is more contribure/architecture/design documentation.

dirkmueller:
I'm feeling an inconsistency between "landscapes"  and "sapsystems" - either all should have a sap prefix or none (I suggest to remove it everywhere). 

Other than that I'm also confused about why environments are implemented the way they are, I suggest to go with Option 1, or just stop at landscapes for now and worry about environments later.

README.md
arbulu89: Well, this goes here to make things easy and allow the user to use the groupings, without setting these options manually the filters don't exist, so it is an end user option to create them or not. Eventually, when this is done by the console this will disappear from the README file.

I don't think it is that important right now where to put the information. I just wanted to show you guys how you can use them.

arbulu89:
> I'm feeling an inconsistency between "landscapes" and "sapsystems" - either all should have a sap prefix or none (I suggest to remove it everywhere).
> 

Ok, I can remove the `sap` prefix from them.

> Other than that I'm also confused about why environments are implemented the way they are, I suggest to go with Option 1, or just stop at landscapes for now and worry about environments later.

Why are you confused? Could you elaborate more? I have implemented them this way because I found it the best, but If you have other proposals I'm open to discuss them

stefanotorresi:
> I'm feeling an inconsistency between "landscapes" and "sapsystems" - either all should have a sap prefix or none (I suggest to remove it everywhere).

I agree that we need some consistency, but I would add the SAP prefix in all these three categories, so that we clarify that it's specific terminology (e.g. "Systems" might also have generic connotations)

arbulu89:
**Disclaimer**: I will do the needed renaming in a next PR. This is already big enough to change all the elements.

@stefanotorresi @dirkmueller @lee-martin New changes on https://github.com/trento-project/trento/pull/34/commits/baf0133374ae1fa3451475810457423b16197e6f

Now, we have 3 new sidebar elements: `Environments`, `Landscapes` and `SAP Systems`. Each of them work individually, but you can move along them going in `Environment->Landscape->SAP system` direction.

Besides that, I have implemented minimal detail pages for each of them. By now they don't have almost any information, but I find it needed to make a reasonable tree view

Here the pictures:
image
image
image

Detail pages (they filter the subsequente elements and show the name by now):
image
image
image

stefanotorresi (approved):
Looking very good from here!

stefanotorresi:
does this have mitigations for #38 ? I only see some `c.Error` calls but I'm not clear whether we're handling 404s correctly or not.

arbulu89:
> does this have mitigations for #38 ? I only see some `c.Error` calls but I'm not clear whether we're handling 404s correctly or not.
> 
> (you can handle this in a follow-up PR, btw)

No, I didn't implement any error handling by now (neither for the hosts/clusters/envs, etc).
But yes, this is something we need to take care of

PR #36, opened 2021-04-27
Check correct config of Python Hook SAPHanaSR

aleksei-burlakov:
Ref: section 8.1 Implementing the Python Hook SAPHanaSR
https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-sr-guide-PerfOpt-15/

### DIFF
--- a/examples/azure-rules.yaml
+++ b/examples/azure-rules.yaml
@@ -753,7 +753,7 @@ groups:
           More information at https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/sap-hana-high-availability#create-sap-hana-cluster-resources
         scored: true
       - id: 1.5.7
-        description: "Check /hana/share/myHooks/global.ini"
+        description: "Check SAPHanaSR configuration"
         audit: |
          sid=$(crm configure show | grep -m1 SID= | sed -e "s/.*SID=\(...\).*/\1/")
          cat /hana/shared/$sid/global/hdb/custom/config/global.ini
--- a/examples/azure-rules.yaml
+++ b/examples/azure-rules.yaml
@@ -752,6 +752,28 @@ groups:
 
           More information at https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/sap-hana-high-availability#create-sap-hana-cluster-resources
         scored: true
+      - id: 1.5.7
+        description: "Check /hana/share/myHooks/global.ini"
+        audit: |
+         sid=$(crm configure show | grep -m1 SID= | sed -e "s/.*SID=\(...\).*/\1/")
+         cat /hana/shared/$sid/global/hdb/custom/config/global.ini
+         sed -n -e '/ha_dr_provider_SAPHanaSR\]/,/\[/ p; /trace\]/,/\[/ p' /hana/shared/$sid/global/hdb/custom/config/global.ini
+         path=$(cat /hana/shared/$sid/global/hdb/custom/config/global.ini | grep "path =" | sed "s/.*= //")
+         if [ -s "$path/SAPHanaSR.py" ]; then echo "HookFound"; else echo "HookNotFound"; fi
+        tests:
+          test_items:
+          - flag: "provider = SAPHanaSR"
+          - flag: "path = /usr/share/SAPHanaSR"
+          - flag: "execution_order = 1"
+          - flag: "ha_dr_saphanasr = info"
+          - flag: "HookFound"
+        remediation: |
+          ## Remediation
+          Set the recommended values in global.ini.
+
+          ## References
+          - https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/sap-hana-high-availability#implement-the-python-system-replication-hook-saphanasr
+        scored: true
   - id: 1.6
     description: "Check that <sidadm> sudo user is set up in /etc/sudoers"
     checks:


examples/azure-rules.yaml
```diff
@@ -656,4 +656,17 @@ groups:
...
+        audit: 'cat /hana/share/myHooks/global.ini'
+        tests:
+          test_items:
+          - flag: "provider = SAPHanaSR"
+          - flag: "path = /usr/share/SAPHanaSR"
+          - flag: "execution_order = 1"
+          - flag: "ha_dr_saphanasr = info"
+        remediation: |
+          Set the recommended values in global.ini. Ref: section 8.1 Implementing the Python Hook SAPHanaSR
+          https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-sr-guide-PerfOpt-15/
```
dirkmueller, line 671: is this really an "azure check"? shouldn't we refer to https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/sap-hana-high-availability#implement-the-python-system-replication-hook-saphanasr then instead?

aleksei-burlakov (commented):
I have changed the URL in the remedy

stefanotorresi (approved):

diegoakechi:
@aleksei-burlakov what happens with this check if global.ini has more lines or the hook configuration is not on the beginning of the file?

Global.ini is a generic config file for HANA and it is not uncommon to have more lines for other hooks, params, etc...

examples/azure-rules.yaml
```diff
@@ -656,4 +656,17 @@ groups:
           Please, check the naming convention in case the resource configuration is done correctly.
            
           More information at https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/sap-hana-high-availability#create-sap-hana-cluster-resources
-        scored: true
\ No newline at end of file
+        scored: true
+      - id: 1.5.7
+        description: "Check /hana/share/myHooks/global.ini"
+        audit: 'cat /hana/share/myHooks/global.ini'
```
diegoakechi, line 662: Actually, this is not the location where global.ini resides, but where the python hook will be, so this check will always fail.

diegoakechi (changes requested):

aleksei-burlakov:
> @aleksei-burlakov what happens with this check if global.ini has more lines or the hook configuration is not on the beginning of the file?
> 
> Global.ini is a generic config file for HANA and it is not uncommon to have more lines for other hooks, params, etc...

This will work. Those flags check that there are those flags among the whole file.

examples/azure-rules.yaml
aleksei-burlakov: Thanks, fixed. It searches the global.ini in /hana/shared/<SID>/global/hdb/custom/config/ now.

examples/azure-rules.yaml
```diff
@@ -656,4 +656,17 @@ groups:
...
+        scored: true
+      - id: 1.5.7
+        description: "Check /hana/share/myHooks/global.ini"
+        audit: 'cat $(find /hana/shared -name global.ini | grep "global/hdb/custom/config/global.ini" | head -n1)'
+        tests:
+          test_items:
+          - flag: "provider = SAPHanaSR"
+          - flag: "path = /usr/share/SAPHanaSR"
+          - flag: "execution_order = 1"
+          - flag: "ha_dr_saphanasr = info"
```
stefanotorresi, line 668: I'm not entirely clear how does this check work.
Why are we first finding multiple config files and then scanning only the first one?

Also, how do multiple `test_items` are combined? Does it mean that it looks for all of them?

aleksei-burlakov: > I'm not entirely clear how does this check work.
> Why are we first finding multiple config files and then scanning only the first one?

The file is stored in /hana/shared/<SID>/global/hdb/custom/config/ but we cannot parametrise the SID, so it takes for the first SID found

> Also, how do multiple `test_items` are combined? Does it mean that it looks for all of them?

It checks that the file has all those strings "provider = SAPHanaSR" AND "path = /usr/share/SAPHanaSR" AND ...

stefanotorresi: A simpler approach that could work is to repeat the check of ALL the SIDs, but you need to ask Lee if that would work. I'm not sure about how the shared storage might be used in the context of multiple systems.

aleksei-burlakov: @lee-martin we need your advice.

lee-martin:
@aleksei-burlakov @stefanotorresi @dirkmueller @arbulu89  (I mentioned you all since the topic of finding the SID is relevant in many places)

Actually, since we are only interested in the HANA which pacemaker is taking care of, why not reverse the logic of checks. Specifically, assume the SID and Instance Number in pacemaker to be correct and use that SID & Instance number as the reference baseline for all checks.  If the checks fail, then then the cluster has a problem and needs attention. 


Also remember we have several checks requiring the SID, so it could make sense to reuse that finding, or at least write the collection logic once and reuse I suppose. The frontend also needs to find the SID, so suggest tackling this logic together with frontend.


One can find SIDs in /usr/sap/sapservices  or you could also look at /usr/sap/SID - but any of these could have multiple SIDs, and we only want to check the one which pacemaker controls. :/  Hence the idea above to assume the SID in pacemaker to be correct and work from there. I think the end result should be the same?

aleksei-burlakov:
> @aleksei-burlakov @stefanotorresi @dirkmueller @arbulu89 (I mentioned you all since the topic of finding the SID is relevant in many places)
> 
> Actually, since we are only interested in the HANA which pacemaker is taking care of, why not reverse the logic of checks. Specifically, assume the SID and Instance Number in pacemaker to be correct and use that SID & Instance number as the reference baseline for all checks. If the checks fail, then then the cluster has a problem and needs attention.
> 
> Also remember we have several checks requiring the SID, so it could make sense to reuse that finding, or at least write the collection logic once and reuse I suppose. The frontend also needs to find the SID, so suggest tackling this logic together with frontend.
> 
> One can find SIDs in /usr/sap/sapservices or you could also look at /usr/sap/SID - but any of these could have multiple SIDs, and we only want to check the one which pacemaker controls. :/ Hence the idea above to assume the SID in pacemaker to be correct and work from there. I think the end result should be the same?

This should work. I have in mind however to make the trento not a completely reading-only application, but let it to write a file with environment variables before running the checks. What do you think @lee-martin @stefanotorresi ?

stefanotorresi:
Hey @lee-martin, @aleksei-burlakov,

> Also remember we have several checks requiring the SID, so it could make sense to reuse that finding, or at least write the collection logic once and reuse I suppose.

Finding out the SID is a bash one-liner querying the Pacemaker CIB, so I don't think we have a compelling reason for the added complexity of reuse here.

> The frontend also needs to find the SID, so suggest tackling this logic together with frontend.

Lee, mate, let me reiterate again a _core_ design aspect of the project: the HA checker is **completely segregated** from the frontend, they have nothing to do with each other! This is a deliberate design choice that we have to live with for now.
The discovery of facts, consumed by the frontend, cannot be reused by the checker component until we completely redesign it.

> let it to write a file with environment variables before running the checks.

Hmm, nope, I don't fancy this approach at all. One of the aspects making the current implementation of the checks simple is precisely that they don't have state. This is one of the key limitations of the current HA checker design and there is no point in trying to circumvent this aspect with quirky things like trying to make bash scripts stateful.

We need to embrace this aspect for the time being, so if two checks need the same fact, I would advise to just copy-paste the bash line that produces that fact, because doing otherwise will add complexity without any meaningful gain.

lee-martin:
> Hey @lee-martin, @aleksei-burlakov,
@stefanotorresi  @aleksei-burlakov 

> 
> > Also remember we have several checks requiring the SID, so it could make sense to reuse that finding, or at least write the collection logic once and reuse I suppose.
> 
> Finding out the SID is a bash one-liner querying the Pacemaker CIB, so I don't think we have a compelling reason for the added complexity of reuse here.

My point is that we have a repeated need to query the same SID thing from the CIB, I thought in programming it was typical to be able to access the same function from multiple places, that way you also only need to maintain that functionality in one place. Just a thought to make maintenance easier.

> 
> > The frontend also needs to find the SID, so suggest tackling this logic together with frontend.
> 
> Lee, mate, let me reiterate again a _core_ design aspect of the project: the HA checker is **completely segregated** from the frontend, they have nothing to do with each other! This is a deliberate design choice that we have to live with, for now.
> The discovery of facts, consumed by the frontend, cannot be reused by the checker component until we completely redesign it.

Sure, I understood in the meantime that frontend and checker implementations are separate and cannot reuse each others data, but that was not my point. Both frontend and checker have some common problems to solve, like finding SID. So I was suggesting to solve the common problem, both checker and frontend developers could cooperate in finding the "logic" of how to go about best finding the SID, and then (as you say) the implementations are separate. It would seem to make sense that Trento uses a consistent logic for finding SID, even if they don't share the same binary/data directly. A bit like invent the wheel only once, but then reuse the same concept, make sense? 

> 
> > let it to write a file with environment variables before running the checks.
> 
> Hmm, nope, I don't fancy this approach at all. One of the aspects of the current implementation making checks very simple is precisely that they don't have state. This is one of the key limitations of the current HA checker design and there is no point in trying to circumvent this aspect with quirky things like trying to make bash scripts stateful.
> 
> We need to embrace this aspect for the time being, so if two checks need the same fact, I would advise to just copy-paste the bash line that produces that fact, because doing otherwise will add complexity without any meaningful gain.

Embracing copy & paste sounds like a need emerging for different approach, even if not immediately. We can discuss with @dirkmueller

lee-martin:
@stefanotorresi 

May I propose the following approach for the host local check (please give me you're feedback/ideas of course)

    Get the SID of the HANA RDBMS from the pacemaker CIB
    Using that information go to the specific global.ini for the HANA system  our cluster is managing (we should not assume /hana/shared as a base path since a user might not use the default, going via the SIDadm user should be reliable to get you to the right place though)
    Now inside the global.ini we need to check that the entries are correct.
    NOTE: that the "path" is user definable
        so for that part we need to extract the path
        then check if the python hook script is there
        then also check if the hook script is the same as the one inside the SAPHanaSR rpm (/usr/share/SAPHanaSR/SAPHanaSR.py) - if different we should warn
        and finally check the path is RW (although the guide does not mention for which user, but I would assume SIDadm for now)

What do you think?

stefanotorresi:
> 2\. going via the SIDadm user should be reliable to get you to the right place though

What does it mean "going via the SIDadm user"? You mean the home directory of that user?

> * then also check if the hook script is the same as the one inside the SAPHanaSR rpm (/usr/share/SAPHanaSR/SAPHanaSR.py), if different we should warn

I'm not sure I understand the need for this: what makes this file special? Can't we just assume that it's the same? After all, we don't do such consistency check for the bazillion of other files in the system, which are also shipped by our RPMs. :)

> * and finally check the path is RW
Does this refer to the path of the hook file or of the directory it is contained within?
Also, why are we requiring that a system directory/file needs to be writable? Is it the hook actually requiring to write on its current directory when it is executed?
I suspect the indication of the best practice document is a bit ambiguous and potentially at fault: it mentions that the hook might be optionally installed in a custom user directory, and of course that would need to be writable. I don't think we need to check for this detail, to qualify the installation as correct, though.

lee-martin:
> > 2. going via the SIDadm user should be reliable to get you to the right place though
> 
> What does it mean "going via the SIDadm user"? You mean the home directory of that user?

If you "su -" to the SIDadm user, then the profile of that user leads you to the correct path of the global.ini. Since you cannot assume "/hana/shared" is the top level path this was just an idea of a reliable way to find the global.ini. IIRC you can use the command/alias "cdhdb" as SIDadm, I don't remember precise details of the top of my head. If you have a HANA system we can take a look.

> 
> > * then also check if the hook script is the same as the one inside the SAPHanaSR rpm (/usr/share/SAPHanaSR/SAPHanaSR.py), if different we should warn
> 
> I'm not sure I understand the need for this: what makes this file special? Can't we just assume that it's the same? 

Well the user is allowed to COPY the file from our rpm and put in a path of his choice. Here there is potential that the user forgot to copy a new version of the python file after updating the rpm, or potential for user modification. Either way, this is something to check and warn about IMO.

> After all, we don't do such consistency check for the bazillion of other files in the system,
>  which are also shipped by our RPMs. :)

Well in this case the user is allowed to copy the file from the RPM, that's my concern as mentioned above.
What examples are on your mind? Maybe something we should also check?

> 
> > * and finally check the path is RW
> 
> Does this refer to the path of the hook file or of the directory it is contained within?

I am quoting what we document, I agree that is not clear. No reason we cannot in parallel open a doc bug to query this (and we know the authors too).

> Also, why are we requiring that a system directory/file needs to be writable? Is it the hook actually requiring to write on its current directory when it is executed?

Maybe it creates some temporary file, I do not know. I just mailed the SAP HA experts, lets see.

> I suspect the indication of the best practice document is a bit ambiguous and potentially at fault: it mentions that the hook might be optionally installed in a custom user directory, and of course that would need to be writable.

Might be. If Trento has the side effect of improving our documentation and/or other code, then that itself is a success. Root cause fixes are always better.

>  I don't think we need to check for this detail, to qualify the installation as correct, though.

We do no know this. If a setup guide tells the reader to do something very explicitly then there is usually a good reason. If our HA experts decide the step is not necessary then of course we can open a doc bug and skip the check, but as it stands the documentation says "set rw" permissions. Good to ask though.

aleksei-burlakov:
@lee-martin @stefanotorresi  I have amended the PR. It takes the SID from `cib.xml` now, but it doesn't do the ("su -" to the SIDadm). Shell I continue and add the "su -" option to get the path, or it's already enough how it is. The example in https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-sr-guide-PerfOpt-15/#_set_up_sap_hana_hadr_providers takes the `global.ini` exactly from `/hana/shared/SID/global/hdb/custom/config/`

stefanotorresi (changes requested):
@aleksei-burlakov as @lee-martin mentioned, we cannot assume that the top level dir is `/hana/shared`, so you need to get the home directory of the `${sid}adm` user somehow; here is a SO thread for inspiration: https://superuser.com/questions/484277/get-home-directory-by-username.

lee-martin:
@aleksei-burlakov  & @aleksei-burlakov Ah, hold on, I did not say we need the homedir of SIDadm. I was just dumping roughly my memory of how I find the global.ini , but I guess I need to provide more details - and we need a real test system up to verify & test these things.

Ok, so I did some research:

- The SIDadm user has lots of aliases to make the SAP admins life easier (just like Trento wants to do). SAP Note "2813612 - HANA alias commands for sidadm user" gives a nice list of these, you can find a public copy of that SAP note made by someone at https://itsiti.com/sap-hana-alias-commands-for-sidadm-user/
- The interesting one for use here is the alias "cdcoc", which takes the user directly to the custom config (coc?) directory, which is where our global.ini is.
- And it gets more interesting, the SIDadm shell defines the "cdcoc" alias as "cdcoc=cd /usr/sap/[SID]/SYS/global/hdb/custom/config". So, since we know the SID from the CIB (@aleksei-burlakov  can we assume the cib.xml to be the truth and what is actually live/running in pacemaker?), we now know the path to the global.ini without the need to assume the base path on HANA. 
- In the background, SAP creates symlinks out of the /usr/sap/SID path structure into the real paths like the /hana/shared example you started with. The important thing is you should be able to rely on /usr/sap/SID on all installations.


BTW: Dont forget to update the remediation into those sections we discussed and attribute the Setup guide as the reference for these things.

BTW v2: these python checks are HANA specific, so not in azure-rules.yaml, right?

stefanotorresi:
Okay, so I get that we can use `/usr/sap/${sid}/SYS/global/hdb/custom/config/global.ini` as a final, reliable path. Can we carry on with this?

stefanotorresi:
> Okay, so I get that we can use `/usr/sap/${sid}/SYS/global/hdb/custom/config/global.ini` as a final, reliable path. Can we carry on with this?

@lee-martin answered YES and lighted it up!

lee-martin:
> Okay, so I get that we can use `/usr/sap/${sid}/SYS/global/hdb/custom/config/global.ini` as a final, reliable path. Can we carry on with this?

YES

aleksei-burlakov:
@stefanotorresi  As agreed in the CFSA-88, this check became more specific. It doesn't check the flags among the whole global.ini file but only in two sections: ha_dr_provider_SAPHanaSR and trace.
Besides, it checks that the hook exists in the specified path.

examples/azure-rules.yaml
```diff
@@ -752,6 +752,28 @@ groups:
 
           More information at https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/sap-hana-high-availability#create-sap-hana-cluster-resources
         scored: true
+      - id: 1.5.7
+        description: "Check /hana/share/myHooks/global.ini"
```
stefanotorresi, line 756:
```suggestion
        description: "Check SAPHanaSR configuration"
```

stefanotorresi (approved):

PR #39, opened 2021-04-28
Document the trento nomenclature

dirkmueller:
This adds some references to how Trento UI uses the various nomenclature.

### DIFF
--- a/docs/trento-architecture.md
+++ b/docs/trento-architecture.md
@@ -81,7 +81,6 @@
     |                                                                                                                                                            |
     +------------------------------------------------------------------------------------------------------------------------------------------------------------+
 
-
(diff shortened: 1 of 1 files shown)


stefanotorresi (approved):
LGTM!

stefanotorresi (changes requested):
Actually, could we put this in a dedicated markdown file? :)

dirkmueller:
@stefanotorresi I had it originally in a separate file, but the nomenclature makes total sense as a explanation of the architecture picture. do you agree?

stefanotorresi (approved):

stefanotorresi:
fair enough!!

PR #45, opened 2021-04-30
Unify the consul api usage in web and agent

arbulu89:
As the tittle says, unify the consul api usage.

I found that the agent side and web side are using a different consul agent. The first is using directly the official consul api client, but the web side is using the internal one (for unit test and mocking reasons). 
This implementation just change the usage of it, to use the internal consul api in the agent as well.

Without this, we cannot share some methods later on in the `web` and `agents` sides if they need the consul client, as they have different specification.

Besides this, It doesn't change anything on the functional side.


stefanotorresi (commented):
Looks good! The only thing I would make sure of is that the import alias is always the same everywhere.
Worth to have a look and make them all consistent, while we're at it.
I think Dirk used `consulInternal` for our own wrapper and just `consul` for the upstream one? I'm not suggesting to do that, I'm fine with `consul` and `consulApi` respectively, but just to ensure that we're using always the same pair of aliases.

dirkmueller (approved):
LGTM

PR #47, opened 2021-05-03
Create background documentation for Trento

lee-martin:
As a first step I'm uploading a copy of the SUSE Best Practice for SAP HANA scale-up which I will annotate with references to our checks (or CFSA tickets requesting those checks). This way we can see why checks have been implemented, as well as be able to attribute them the the original source on suse.com


stefanotorresi:
Hey Lee, this is good effort, but I'm afraid HTML files are not really usable in the context of project documentation: from the contributor standpoint, the normal user expectations is to just open the file itself (directly within GitHub or in an text editor after having cloned the sources) and read it. To this end, Markdown is generally used as the lingua-franca of documentations on Git, because it's basically plain text without markup.

There are also a couple of other problems:
- the `htmlpreview.github.io` link doesn't seem to be working, probably because it needs a permalink of the document.
- there is little value in referencing links to SUSE's private Jira board in a publicly available document.

What I would suggest to do is just to elaborate the `background.md` document by referencing the public SUSE document, instead of trying to do the other way around.

stefanotorresi (changes requested):
as per my comment above and discussed face-to-face.

dirkmueller:
you can find the sources in https://github.com/SUSE/suse-best-practices/blob/main/adoc/SLES4SAP-hana-scaleOut-PerfOpt-12.adoc - unfortunately it is asciidoc.

stefanotorresi:
I'm closing this for now, since I think we ended up addressing this in some internal Confluence page.

PR #48, opened 2021-05-03
Add CONSUL_ADDR to override default lo addr

rtorrero:
This PR allows to use an environment variable to override the default `127.0.0.1` address that is used to connect to the consult HTTP API. The `127.0.0.1` default assumes that the `consul` agent is running on the same machine, which is not necessarily true.

### DIFF
--- a/internal/consul/consul.go
+++ b/internal/consul/consul.go
@@ -1,6 +1,8 @@
 package consul
 
 import (
+	"os"
+
 	consulApi "github.com/hashicorp/consul/api"
 )
 
@@ -39,7 +41,15 @@ type Health interface {
 }
 
 func DefaultClient() (Client, error) {
-	w, err := consulApi.NewClient(consulApi.DefaultConfig())
+	config := consulApi.DefaultConfig()
+	addrVal, addrPresent := os.LookupEnv("CONSUL_ADDR")
+	if addrPresent {
+		config.Address = addrVal
+	} else {
+		config.Address = "127.0.0.1:8500"
+	}
+
+	w, err := consulApi.NewClient(config)
 	if err != nil {
 		return nil, err
 	}


arbulu89 (commented):
Hi @rtorrero ,

Is this needed? I think that the value is used even with this patch.
I use the `export CONSUL_HTTP_ADDR=127.0.0.3:8500` since the beginning to attach our trento agent to the consul agent without any issue.

Other thing would we if we could pass the variable from the as a flag to trento, but I don't know if this is really useful if the variable export already works

rtorrero:
@arbulu89 I might be wrong but testing with `export CONSUL_HTTP_ADDR` didn't work for me. I think this adjusts the address that consul will bind to, but `trento` will still attempt to connect to `127.0.0.1` (The default address the go consul API uses)

In any case, this is indeed valuable; would it make sense that instead of using a different env var such as `CONSUL_ADDR`, we also use the default `CONSUL_HTTP_ADDR` and we adjust our client config based on this?

stefanotorresi:
> Other thing would we if we could pass the variable from the as a flag to trento, but I don't know if this is really useful if the variable export already works

I would argue against introducing new flags just for internal Consul details. Since Consul can leverage environment variables, I think we should use those.

Since you folks are not sure if the changes here proposed actually do what they're supposed to do, I would suggest to cover `DefaultClient()` with unit tests. ;)

internal/consul/consul.go
```diff
@@ -39,7 +41,15 @@ type Health interface {
...
 
 func DefaultClient() (Client, error) {
-	w, err := consulApi.NewClient(consulApi.DefaultConfig())
+	config := consulApi.DefaultConfig()
+	addrVal, addrPresent := os.LookupEnv("CONSUL_ADDR")
+	if addrPresent {
+		config.Address = addrVal
+	} else {
+		config.Address = "127.0.0.1:8500"
+	}
```
stefanotorresi, line 50: I don't think we need this `else` statement, since that's actually the default value, as shown here:
https://github.com/hashicorp/consul/blob/8d6cbe72813779f60ac24d119cc311c21003f4ce/api/api.go#L396

stefanotorresi (changes requested):

arbulu89:
> @arbulu89 I might be wrong but testing with `export CONSUL_HTTP_ADDR` didn't work for me. I think this adjusts the address that consul will bind to, but `trento` will still attempt to connect to `127.0.0.1` (The default address the go consul API uses)
> 
> In any case, this is indeed valuable; would it make sense that instead of using a different env var such as `CONSUL_ADDR`, we also use the default `CONSUL_HTTP_ADDR` and we adjust our client config based on this?

@rtorrero How did you run your test?
I simply do `export CONSUL_HTTP_ADDR=127.0.0.2:8500` and without any change when `trento` is started I can see this on the logs:
```
021/05/04 13:09:21 [DEBUG] (runner) final config: {"Consul":{**"Address":"127.0.0.2:8500"**,"Namespace":"","Auth":{"Enabled":false,"Username":"","Password":""},"Retry":{"Attempts":12,"Backoff":250000000,"MaxBackoff":60000000000,"Enabled":true},"SSL":{"CaCert":"","CaPath":"","Cert":"","Enabled":false,"Key":"","ServerName":"","Verify":true},"Token":"","Transport":{"DialKeepAlive":30000000000,"DialTimeout":30000000000,"DisableKeepAlives":false,"IdleConnTimeout":90000000000,"MaxIdle
```

We are using the same exact code to create the client that `consul` executable does, so it makes sense that both use the same env variables.

internal/consul/consul.go
rtorrero: Totally right, I'll remove it.

rtorrero:
> > @arbulu89 I might be wrong but testing with `export CONSUL_HTTP_ADDR` didn't work for me. I think this adjusts the address that consul will bind to, but `trento` will still attempt to connect to `127.0.0.1` (The default address the go consul API uses)
> > In any case, this is indeed valuable; would it make sense that instead of using a different env var such as `CONSUL_ADDR`, we also use the default `CONSUL_HTTP_ADDR` and we adjust our client config based on this?
> 
> @rtorrero How did you run your test?
> I simply do `export CONSUL_HTTP_ADDR=127.0.0.2:8500` and without any change when `trento` is started I can see this on the logs:
> 
> ```
> 021/05/04 13:09:21 [DEBUG] (runner) final config: {"Consul":{"Address":"127.0.0.2:8500","Namespace":"","Auth":{"Enabled":false,"Username":"","Password":""},"Retry":{"Attempts":12,"Backoff":250000000,"MaxBackoff":60000000000,"Enabled":true},"SSL":{"CaCert":"","CaPath":"","Cert":"","Enabled":false,"Key":"","ServerName":"","Verify":true},"Token":"","Transport":{"DialKeepAlive":30000000000,"DialTimeout":30000000000,"DisableKeepAlives":false,"IdleConnTimeout":90000000000,"MaxIdle
> ```
> 
> Check the `"Address":"127.0.0.2:8500"` part
> 
> We are using the same exact code to create the client that `consul` executable does, so it makes sense that both use the same env variables.

My test was done using the containers. I did try today using the `export CONSUL_HTTP_ADDR=...` on some VM I created (instead of running it in the docker-compose.yml) and it seems to have worked. Can we leave this open for a few days until I figure out what I was doing wrong? 

In any case i'd say that you are probably right and `CONSUL_HTTP_ADDR` is honored by the consul go module (and not only by the consul binary as I initially thought)

rtorrero:
After further tests, I've concluded that @arbulu89 is indeed right and by exporting the `CONSUL_HTTP_ADDR` env var, trento also attempts to connect to consult agents using the value specified in that var.

Thanks for the reviews and sorry for the noise!

PR #49, opened 2021-05-03
Add Dockerfile and docker-compose.yml

rtorrero:
This PR should make it easier during development to quickly set up a simple docker-based cluster with `consul` & `trento` and also ensure that all developers are making use of the same versions of our dependencies.

~~**Note that it is using the environment variable from https://github.com/trento-project/trento/pull/48**, so that PR is a requirement for the docker-compose.yml.~~
Update: not needed anymore

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,6 +230,39 @@ You can install it with `go install github.com/vektra/mockery/v2@latest`.
 > **Note**  
 > Be sure to add the `mockery` binary to your `$PATH` environment variable so that `make` can find it.
 
+
+## Docker
+
+To assist in testing & developing `trento`, we have added a [Dockerfile](Dockerfile) 
+that will automatically fetch all the required compile-time dependencies to build
+the binary and finally a container image with the fresh binary in it.
+
+We also provide a [docker-compose.yml](docker-compose.yml) file that allows to
+deploy other required services to run alongside `trento` by fetching 
+the images from the [dockerhub](https://hub.docker.com/) registry and running 
+the containers in your `docker` instance.
+
+To only build the docker image:
+```shell
+git clone https://github.com/trento-project/trento.git
+cd trento
+docker build -t trento ./
+```
+
+If you want to build & start `trento` and it's dependencies, you only need `docker-compose`:
+```shell
+docker-compose up
+```
+
+The application should be reachable on the port that is defined in the 
+`docker-compose.yml` file (8080 by default).
+
+> **Note**
+> Take into account that `trento` requires an agent instance that is running on
+> the OS (*not* inside a container) so while it is possible to hack the `docker-compose.yml`
+> file to also run a `trento` agent, it makes little sense as most of the checks
+> require direct access to host files.
+
 # Support
 
 As the project is currently in its early stages, we suggest that any question or
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,40 @@
+version: '3'
+
+services:
+
+  consul-server-1:
+    image: consul:latest
+    networks:
+      - trento
+    ports:
+      - "8400:8400"
+      - "8500:8500"
+      - "8600:8600"
+      - "8600:8600/udp"
+    command: "agent -server -bootstrap-expect 3 -ui -client 0.0.0.0"
+
+  consul-server-2:
+    image: consul:latest
+    networks:
+      - trento
+    command: "agent -server -retry-join consul-server-1 -client 0.0.0.0"
+
+  consul-server-3:
+    image: consul:latest
+    networks:
+      - trento
+    command: "agent -server -retry-join consul-server-1 -client 0.0.0.0"
+
+  trento-web:
+    build: .
+    image: trento:latest
+    networks:
+      - trento
+    ports:
+      - "8080:8080"
+    environment:
+      - CONSUL_HTTP_ADDR=consul-server-1:8500
+    command: "web serve -p 8080"
+
+networks:
+  trento:
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,11 @@
+FROM opensuse/leap:15.2
+MAINTAINER Rubén Torrero Marijnissen <rtorreromarijnissen@suse.com>
+
+COPY . /app
+
+RUN zypper -n in go1.16 nodejs14 make
+RUN cd /app && make build
+
+EXPOSE 8080/tcp
+
+ENTRYPOINT ["/app/trento"]


arbulu89:
@rtorrero Thanks!

I think the docker container would be useless for the agent side, as this needs to run several commands in the host itself to get the data of the cluster.
It could be useful though for the server side.

I don't know if this would somehow change the dockerfile implementation

Xabi

rtorrero:
> @rtorrero Thanks!
> 
> I think the docker container would be useless for the agent side, as this needs to run several commands in the host itself to get the data of the cluster.
> It could be useful though for the server side.
> 
> I don't know if this would somehow change the dockerfile implementation
> 
> Xabi

Yes, you are totally right, I've removed it from the docker-compose.yml file. The Dockerfile itself is not affected as we are currently using the same binary for the agent and the webserver.

Note that depending on what we decide in https://github.com/trento-project/trento/pull/48 in terms of which env var we should use, this PR might still change.

README.md
```diff
@@ -230,6 +230,32 @@ You can install it with `go install github.com/vektra/mockery/v2@latest`.
...
+too.
+
+To build the image:
+```shell
+git clone https://github.com/trento-project/trento.git
+cd trento
+docker build -t trento ./
+```
+
+The only parameter that needs to be defined is
```
arbulu89, line 248: @rtorrero Is there something missing here? It looks like the phrase is not finished

docker-compose.yml
```diff
@@ -0,0 +1,46 @@
+version: '3'
+
+services:
+
+  consul-agent-1: &consul-agent
```
arbulu89, line 5: I don't really get it. What are all of these `agent/servers` doing?
Do we need consul agents here>
As far as I understand having 3 consul agents working as servers is what we need, right?

arbulu89 (changes requested):
@rtorrero I think the README has one incomplete phrase.
Besides that, I don't really get the docker compose entries. What is the docker-compose creation there?

docker-compose.yml
```diff
@@ -0,0 +1,46 @@
...
+  consul-server-bootstrap:
+    <<: *consul-agent
+    ports:
+      - "8400:8400"
+      - "8500:8500"
+      - "8600:8600"
+      - "8600:8600/udp"
+    command: "agent -server -bootstrap-expect 3 -ui -client 0.0.0.0"
+
+  trento-webserve:
```
stefanotorresi, line 33: (nit) how about:
```suggestion
  trento-web:
```

docker-compose.yml
```diff
@@ -0,0 +1,46 @@
...
+    image: trento:latest
+    networks:
+      - trento
+    ports:
+      - "8080:8080"
+    expose:
+      - "8080"
```
stefanotorresi, line 40: I don't think we need this here, because it's in the Dockerfile already. What does it do?

docker-compose.yml
```diff
@@ -0,0 +1,46 @@
...
+    <<: *consul-agent
+    ports:
+      - "8400:8400"
+      - "8500:8500"
+      - "8600:8600"
+      - "8600:8600/udp"
+    command: "agent -server -bootstrap-expect 3 -ui -client 0.0.0.0"
+
+  trento-webserve:
+    image: trento:latest
```
stefanotorresi, line 34: We might actually want to use `build: .`, instead of `image` with something that needs to be created separately with `docker build`, so that `docker-compose up` will do the building for us. :)

docker-compose.yml
```diff
@@ -0,0 +1,46 @@
...
+    <<: *consul-server
+
+  consul-server-bootstrap:
+    <<: *consul-agent
+    ports:
+      - "8400:8400"
+      - "8500:8500"
+      - "8600:8600"
+      - "8600:8600/udp"
+    command: "agent -server -bootstrap-expect 3 -ui -client 0.0.0.0"
```
stefanotorresi, line 31: To be honest, I find the usage of yaml anchors a little confusing. How would you feel about making everything a little more explicit and repeating the commands instead?

Also, personal nitpick, I would rename `consul-server-bootstrap` to just `consul-server-1`, but that's just me. 😛

stefanotorresi (changes requested):

docker-compose.yml
rtorrero: Yes, it's not really needed as we are exposing the port in the Dockerfile for the container image. It's a leftover from a override I have in my setup, as I have port 8080 already used in my dev machine and use this to switch it to 8082.

I'll remove it, thanks!

docker-compose.yml
rtorrero: Totally right, as the docker-compose.yml is just for internal usage it might as well reference the local Dockerfile to build the image. We could maybe switch it to static image names at some point in the future if we decide to upload the container in some registry.

docker-compose.yml
rtorrero: I personally don't have a strong preference here. I'd usually lean towards repeating less, but tbh the anchors here are indeed not very easy to read specially if one is not familiar with it. I'll change it.

docker-compose.yml
rtorrero: Indeed, it's pointless even in my usecase as I could simply change the port mapping in the Dockerfile :)

README.md
rtorrero: Yes! Thanks for pointing that out, I've removed it as what I wanted to put there no longer makes sense

docker-compose.yml
rtorrero: Will change, thanks!

docker-compose.yml
rtorrero: It's more of an example than a real requirement, I took it from a demo `docker-compose.yml` file on their repo -> https://github.com/hashicorp/consul/blob/master/demo/docker-compose-cluster/docker-compose.yml

Do you have a suggestion on the layout of the services?

stefanotorresi: Hmm, now that I think about it, we might actually skip the clients entirely, and use the compose file to run just the server side. What do you think?

rtorrero: Sure, at least for now until we figure out if embedding the consul agents in trento is viable. Done!

docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+
+  trento-web:
+    build: .
+    networks:
+      - trento
+    ports:
+      - "8080:8080"
+    environment:
+      - CONSUL_HTTP_ADDR=consul-server-1:8500
+    command: "/app/trento web serve -p 8080"
```
arbulu89, line 36: I think that in order to have the syntax, this last `command` should be `web serve -p 8080`, where `/app/trento` is already defined in the trento dockerfile

docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+  consul-server-1:
+    image: consul:latest
+    networks:
+      - trento
+    ports:
+      - "8400:8400"
+      - "8500:8500"
+      - "8600:8600"
+      - "8600:8600/udp"
+    command: "agent -server -bootstrap-expect 3 -ui -client 0.0.0.0"
```
arbulu89, line 14: In my personal execution I use the `-bind` parameter, but I'm not sure if it is needed.

arbulu89 (approved):
Hi @rtorrero 

It looks good from my side. I have added some minor changes, but they are just small things to consider about.
Besides that, my last 2 cents:

- We need to make sure that the `trento` web is able to access the agents in `8700` port to get the checks output. Maybe it already does, but just to be aware of this thing.
- Consul web is available and `trento` web is able to connect to the consul

As long as 2 tests are green the docker env is good to me

docker-compose.yml
stefanotorresi: You make a good point, and you made me realise another change that we need to make!
Having `command: web serve -p 8080` would only work if `/app/trento` is set as `ENTRYPOINT` in the Dockerfile, instead of it being a `CMD`. So basically we should swap `CMD` with `ENTRYPOINT` there, and leave `CMD` empty.
This would result in `docker run trento` running the default application command (which is `help`, I think), which is exactly what users would expect!

docker-compose.yml
stefanotorresi: I think it's not needed in this case because the container only has one network interface.

docker-compose.yml
rtorrero: Yes, indeed it's cleaner to use ENTRYPOINT in the Dockerfile and only the arguments in the `command` part of the compose. Thanks!

docker-compose.yml
rtorrero: Yes, I'm guessing it's binding on `lo` and the port redirection is taking care of allowing it to be reachable outside the container. Regardless, that part has been tested and works so I wouldn't add anything here.

stefanotorresi (approved):
Great effort, Ruben!

PR #51, opened 2021-05-05
Sap systems discovery

arbulu89:
**Disclaimer**: Still on draft because I didn't create the UT. I want first to validate the logic

Implementation of the SAP systems discovery.
Includes:
- SAP systems discovery using sapcontrol (`internal/sapsystems/sapsystems.go`)
- SAP systems interaction with the KV storage (`internal/sapsystems_kvstore.go`)
- New function in `kv.go`. `PutStr` and `PutMap`
- Visualization of the SAP systems and the SAP system in the environments tree

Tech details:
- The marshalling and unmarshalling of the structs to store/load the data in the KV storage is done with [mapstructure](https://github.com/mitchellh/mapstructure). This package is great to load Maps to structs and viceversa. It makes the code super simple. This package has a `MIT` license, so we should be fine.
- The SAP systems data is stored under `trento/v0/hosts/$host/sapsystems` with the `SAPSystem` struct format from `internal/sapsystems/sapsystems.go`. The structs coming from `sapcontrol` don't have the `mapstructure` syntax, so they just return the struct name as they are
- The SAP systems data is refreshed every discovery. For that, it first deletes the current data to load the new one
- If the SAP system doesn't belong to any environment/landscape, it is stored in the `environments/ungrouped/landscapes/ungrouped/sapsystems` path. As all of the `sapsystems` should belong to a env/land, they belong to this reserved groups. The name can be changed to something else
- Non grouped sapsystemd (don't belong to an env/land) have a different visualization (shown in the pictures)

Limitations:
- If a sapsystem is removed (machine removed, SAP uninstalled, etc), the SAP system will be still there in the environments tree. We will need to have the option to show the inactive sap systems (they don't have any host) to be removed (most probably by the user)
- The sapsystems are stored with their name (`PRD` for instance). This string is not unique. I have implemented an option to get a unique ID of the SAP system, and this ID should be used to store and visualize the hosts. This will be implemented in the next changes.
- The Keys of the SAP systems are changed with any lock. So we can have a scenario (it is strange, but it can happen) that we read the values during a write, and the data might be incomplete. Consul KV has a lock option for that, but it is not implemented yet.

Pictures:
Ungrouped SAP system.
image
image
image

Grouped SAP system (env1/land1): This is done manually adding `PRD` in `trento/v0/environmens/env1/landscapes/land1/sapsystems/PRD/` and changing the metadata values in `trento/v0/hosts/$host/metatada`

image
image
image

FYI: @lee-martin


agent/discover/sapsystem.go
```diff
@@ -0,0 +1,51 @@
...
+func (discover SAPSystemsDiscover) Discover() error {
+	systems, err := sapsystem.NewSAPSystemsList()
+
+	if err != nil {
+		return err
+	}
+
+	discover.SAPSystems = systems
+	for _, s := range discover.SAPSystems {
+		err := s.Store(discover.host.client)
```
stefanotorresi, line 37: I'm a bit confused by this method. Aren't we supposed to use `storeDiscovery`? Why is that a no-op and we are calling another method on the SAPSystem struct instead?

Maybe we should drop `storeDiscovery` from the interface altogether?

internal/sapsystem/sapsystem.go
```diff
@@ -0,0 +1,165 @@
...
+
+	"github.com/SUSE/sap_host_exporter/lib/sapcontrol"
+	"github.com/spf13/viper"
+
+	"github.com/trento-project/trento/internal"
+)
+
+const sapInstallationPath string = "/usr/sap"
+const sapIdentifierPattern string = "^[A-Z][A-Z0-9]{2}$" // PRD, HA1, etc
+const sapInstancePattern string = "^[A-Z]*([0-9]{2})$"   // HDB00, ASCS00, ERS10, etc
```
gereonvey, line 21: You probably want that to be "^[A-Z]+([0-9]{2})$", i.e. **at least one** character in front of 2 digits.

stefanotorresi: nice catch, Gereon!

agent/discover/cluster.go
```diff
@@ -15,19 +15,25 @@ import (
 
 const crm_monPath string = "/usr/sbin/crm_mon"
 const cib_admPath string = "/usr/sbin/cibadmin"
+const ClusterDiscoverId string = "discover_cluster"
```
rtorrero, line 18: Just a very dumb remark, but should we go with this syntax for declaring const or the "block" one used e.g. in https://github.com/trento-project/trento/pull/51/files#diff-c79c0069c6158bdd3d204da61ea8495ac31e344c40050f7e6e8ec298426fe998R23 ?

agent/discover/sapsystem.go
arbulu89: Yes. As my code does the KV storing outside the discovery I don't need to use it.

agent/discover/cluster.go
arbulu89: Both are fine. The 2nd might be cleaner. There might be much more of those in the code. 
I would bet to use an additional PR in the future to fix them if we want to do son

agent/discover/sapsystem.go
stefanotorresi: Ok. I'm opening a follow-up issue on this.

agent/discover/discover.go
```diff
@@ -11,6 +11,7 @@ import (
 type Discovery func() (Discoverer, error)
 
 type Discoverer interface {
+	GetId() string
```
dirkmueller, line 14: please add a comment/description on what the Id is supposed to be.

agent/discover/cluster.go
```diff
@@ -15,19 +15,25 @@ import (
...
+	Id             string
 	host           Discover
 	cibConfig      cib.Root
 	clusterName    string
 	stonithEnabled bool
 	stonithType    consul.ClusterStonithType
 }
 
+func (cluster ClusterDiscover) GetId() string {
+	return cluster.Id
```
dirkmueller, line 34: so its not a uuid, what is it then?

agent/discover/sapsystem.go
dirkmueller: The idea of using storeDiscovery was that all implementations use a shared prefix on where they store the artifacts. that means we do not need to deduplicate key-values (e.g. if two implementations try to store "id", they can do so without overwriting each other). 

why are we deviating from that here?

arbulu89: Because it is much easier to store whole `struct`s, that will have this `prefix` in their code. Keep in mind that the SAP system has dozens of values that we need to store. It is not a good idea to store them individually. That's why we create the struct with the whole obtained data, we convert it to map, and we store it. And this is done in the `Store` method for each of the individuals fields (this is the SAPSystem, but we would have the same for Cluster).  
This way, we abstract the KV usage in the discovery loop, this is now done by the new code.

agent/discover/discover.go
```diff
@@ -24,10 +25,15 @@ type Discoverer interface {
 }
 
 type Discover struct {
+	Id     string
```
stefanotorresi, line 28: nit: since we have the `GetId()` method, I think we can make this private.

stefanotorresi:
@arbulu89 the logic appears mostly sound to me, feel free to remove the draft status.

internal/consul/kv.go
```diff
@@ -112,3 +132,37 @@ func (k *kv) ListMap(prefix, offset string) (map[string]interface{}, error) {
...
+func (k *kv) PutStr(prefix string, value string) error {
+	_, err := k.Put(&consulApi.KVPair{
+		Key:   prefix,
+		Value: []byte(value)}, nil)
+
+	if err != nil {
+		return errors.Wrap(err, "Error storing a new value in the KV storage")
+	}
+
+	//log.Printf("Value %s properly stored at %s", value, prefix)
```
stefanotorresi, line 166: please make sure not to forget this commented line :)

internal/sapsystem/sapsystem.go
```diff
@@ -0,0 +1,165 @@
...
+			i, err := findInstances(path.Join(sapInstallationPath, f.Name()))
+			if err != nil {
+				log.Print(err.Error())
+			}
+			instances = append(instances, i[:]...)
+		}
+	}
+
+	return instances, nil
+}
```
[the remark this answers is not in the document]
arbulu89: This is done in purpose (at least in this particular case). When we have an error in the `findInstances` function, I didn't want to stop the whole finding process, so it just prints the error message but continues the process. 
It will just return an error if no SAP system is found. Otherwise, it will just return an empty list of instances within a SAP systems (I don't know if this happens, but maybe the user remove some instance, or change the user rights, or anything like that).

internal/consul/kv.go
arbulu89: As we don't have a good logging system right now, with different verbosity levels, I just commented this, because it prints hundreds of lines every discovery. It would be a `debug` log in a final version.

agent/discover/discover.go
arbulu89: Yes...
In any case, this id is only used a the consul `checkId`, and I'm not sure if we should integrate the discovery mechanisms as a consul check

agent/discover/cluster.go
arbulu89: This is the `consul` CheckId. As we have more discovery entries with the SAP discovery, the id must belong to the discovery mechanism itself.

Anyway, as commented in other point, I'm not sure if we should include the discovery mechanisms in the consul checks

internal/sapsystem/sapsystem.go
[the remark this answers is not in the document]
stefanotorresi: hmm, I see; so, the only actual error condition in this function is an I/O error while reading the installation path?

arbulu89:
> @arbulu89 the logic appears mostly sound to me, feel free to remove the draft status.

I can do that, but this creates the risk someone merging the PR without creating the UT (that I didn't do before validating the code logic)

stefanotorresi:
nono, I meant after you add the test 😁

internal/sapsystem/sapsystem.go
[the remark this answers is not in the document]
stefanotorresi: Hmm, yes, I would actually add the `continue` statement, now that you mention it: we should never rely on the return value of a function in presence of an error. It is also more communicative of the intent to continue looping even in presence of an error.

arbulu89:
@stefanotorresi I have added some UT to the project (there are many parts not covered yet, but we would need to have some coverage analysis for that).

I have done some changes to make the code testable, adding `afero` among them to have a more flexible File system implementation: https://github.com/spf13/afero

agent/discover/discover.go
```diff
@@ -11,6 +11,8 @@ import (
 type Discovery func() (Discoverer, error)
 
 type Discoverer interface {
+	// Get Consul checkId string. Each discoverer has an unique id for that
+	GetId() string
```
stefanotorresi, line 15: @dirkmueller it's a free text field, there is no format associated to it.

@arbulu89 we could comment as follows:
```suggestion
	// Returns an arbitrary unique string identifier of the discovery, so that we can associate it to a Consul check ID
	GetId() string
```

stefanotorresi (approved):
Brilliant work!

A few non blocking notes for future reference:
- As you pointed out yourself, we need to parse the `Pid` field of `sapcontrol.OSProcess`, which are of type `int32`, and currently ignored by the switch in `KV.PutMap`.
- We might start considering copying the entire `sapcontrol` package into the project, because we already have a case for changing it (i.e. adding `mapstructure` annotations to the data structs coming from the SAPControl webservice)
- Having one Consul check for each discovery might be overkill.

stefanotorresi (approved):

dirkmueller (approved):

PR #54, opened 2021-05-06
check SAPHanaSR >= 0.153, os-release >= 15.1

aleksei-burlakov:
additional checs for CFSA-83.
Ref: https://jira.suse.com/browse/CFSA-83?jql=key%20in%20(CFSA-83)

### DIFF
--- a/examples/azure-rules-perf-optimized.yaml
+++ b/examples/azure-rules-perf-optimized.yaml
@@ -20,6 +20,23 @@ groups:
           Install a SUSE supported OS release
         scored: true
       - id: 2.1.2
+        description: "Check os release"
+        audit: 'echo VERSION_ID=$(cat /etc/os-release | grep VERSION_ID= | sed "s/[^0-9]//g")'
+        tests:
+          test_items:
+            - flag: 'VERSION_ID'
+              compare:
+                op: gte
+                value: 151
+        remediation: |
+          ## Abstract
+          You need at least SUSE Linux Enterprise Server for SAP Applications 15 SP1 or newer
+          ## Remediation
+          Please install or upgrade to a supported OS version
+          ## Reference
+          https://documentation.suse.com/en-us/sbp/all/single-html/SLES4SAP-hana-sr-guide-PerfOpt-15/
+        scored: true
+      - id: 2.1.3
         description: "Check the pacemaker version IS"
         audit: 'rpm -qv pacemaker | sed -e "s/pacemaker-\(.*\)+.*/\1/"'
         tests:
@@ -29,7 +46,7 @@ groups:
         remediation: |
           Update pacemaker packages
         scored: true
-      - id: 2.1.2.exclude
+      - id: 2.1.3.exclude
         description: "Check the pacemaker version IS NOT"
         audit: 'rpm -q --qf "%{VERSION}\n" pacemaker'
         tests:
@@ -43,7 +60,7 @@ groups:
         remediation: |
           Update pacemaker packages
         scored: true
-      - id: 2.1.3
+      - id: 2.1.4
         description: "Check the corosync version IS"
         audit: 'rpm -qv corosync'
         tests:
@@ -53,7 +70,7 @@ groups:
         remediation: |
           Update corosync packages
         scored: true
-      - id: 2.1.4
+      - id: 2.1.5
         description: "Check the sbd version IS"
         audit: 'rpm -qv sbd | sed -e "s/sbd-\(.*\)+.*/\1/"'
         tests:
@@ -63,7 +80,7 @@ groups:
         remediation: |
           Update sbd packages
         scored: true
-      - id: 2.1.4.exclude
+      - id: 2.1.5.exclude
         description: "Check the sbd version IS NOT"
         audit: 'rpm -q --qf "%{VERSION}\n" sbd'
         tests:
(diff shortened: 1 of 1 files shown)


stefanotorresi:
@aleksei-burlakov don't bother referencing jira from here, just do the opposite. :)
This is a follow-up of #31 right?

examples/azure-rules-perf-optimized.yaml
```diff
@@ -20,6 +20,23 @@ groups:
...
+            - flag: 'PRETTY_NAME="SUSE Linux Enterprise Server for SAP Applications 15 SP1"'
+            - flag: 'PRETTY_NAME="SUSE Linux Enterprise Server for SAP Applications 15 SP2"'
+            - flag: 'PRETTY_NAME="SUSE Linux Enterprise Server for SAP Applications 15 SP3"'
+        remediation: |
+          ## Abstract
+          You need at least SLES for SAP 15.1
+          ## Remediation
+          Install at least SLES for SAP 15.1
+          ## Reference
+          T.B.A.
```
stefanotorresi, line 37: @lee-martin could you point us at where we document these version constraints?

examples/azure-rules-perf-optimized.yaml
```diff
@@ -77,7 +94,24 @@ groups:
...
+              compare:
+                op: gte
+                value: 153
+        remediation: |
+          ## Abstract
+          You need at least SAPHanaSR 0.153
+          ## Remediation
+          Install at least SAPHanaSR 0.153
+          ## Reference
+          T.B.A.
```
stefanotorresi, line 112: @lee-martin same as above.

aleksei-burlakov:
> @aleksei-burlakov don't bother referencing jira from here, just do the opposite. :)
> This is a follow-up of #31 right?

Yes, it's the follows up for both #30 and #31.

examples/azure-rules-perf-optimized.yaml
lee-martin: @stefanotorresi 

I have several points to raise here:

1.  While external contributes cannot see our Jira, https://jira.suse.com/browse/CFSA-83 does have the references for both of these checks (min OS version and HanaSR version) - admittedly I only added those details on May 3rd.
Anyway, please refer to section 5.1.2 https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-sr-guide-PerfOpt-15/   )

2. The check is "at least SLES for SAP 15.1" ,  but as far as I understand the implementation it only looks for SP1, SP2 and SP3. What about versions after that, SP4? Next major version perhaps?
Wouldn't it be easier just to check "VERSION_ID" gte 15.1 ?

3. Since SLES4SAP is a SLES at heart, it uses the SLES PRETTY_NAME, e.g. PRETTY_NAME="SUSE Linux Enterprise Server 15 SP1".  Which test systems were giving you a PRETTY_NAME with the SAP Applications names? That would be a bug.

4. In the Abstract we should use the full product name.
Could be good if the abstract were to quote the source reference, in this case it's a simple sentence.

5. Can we add a date to when the references was checked. As Angela pointed out recently thing change over time, so we should record when the reference was last checked.

6. I notice the test IDs just changed due to this PR. Is there some way we could keep test id's static so that we can refer to them consistently in discussions.

stefanotorresi: > 5\. Can we add a date to when the references was checked. As Angela pointed out recently thing change over time, so we should record when the reference was last checked.

This is not an aspect specific for this particular check, so I would avoid addressing this right now.
Besides, since these rule sets are living documents within a version control system, I think trying to keep track of the date of the reference would be just a lot of overhead.

stefanotorresi: > 6\. I notice the test IDs just changed due to this PR. Is there some way we could keep test id's static so that we can refer to them consistently in discussions.

As per our face-to-face discussion, this is bound to happen for the time being, and I don't think we can address this right now.

aleksei-burlakov:
> @aleksei-burlakov let me summarize what needs to be done here:
> 
> * change the check logic so that it doesn't rely on the PRETTY_NAME, which won't give any indication of the system actually being SLES4SAP instead of SLES. Also, the check logic must allow for any future SLES4SAP 15 SP version.
> * use `You need at least SUSE Linux Enterprise Server for SAP Applications 15 SP1 or newer` as abstract.
> * use `Please install or upgrade to a supported OS version` as remediation.
> * put a link to https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-sr-guide-PerfOpt-15 in the references

It's clear with the last 3 points, but not with the first. It may not perfectly match the PRETTY_NAME, but I still can extract either if's version 15 or 12 and which service pack is there. Or is there a better alternative than os-release?

aleksei-burlakov:
> @aleksei-burlakov let me summarize what needs to be done here:
> 
> * change the check logic so that it doesn't rely on the PRETTY_NAME, which won't give any indication of the system actually being SLES4SAP instead of SLES. Also, the check logic must allow for any future SLES4SAP 15 SP version.

@stefanotorresi , as @lee-martin proposed, I used the VERSION_ID instead

> * use `You need at least SUSE Linux Enterprise Server for SAP Applications 15 SP1 or newer` as abstract.

Fixed.

> * use `Please install or upgrade to a supported OS version` as remediation.

Fixed.

> * put a link to https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-sr-guide-PerfOpt-15 in the references

Fixed.

examples/azure-rules-perf-optimized.yaml
```diff
@@ -20,6 +20,23 @@ groups:
...
         scored: true
       - id: 2.1.2
+        description: "Check os release"
+        audit: 'echo VERSION_ID=$(cat /etc/os-release | grep VERSION_ID= | sed "s/[^0-9]//g")'
+        tests:
+          test_items:
+            - flag: 'VERSION_ID'
+              compare:
+                op: gte
+                value: 151
```
stefanotorresi, line 30: @aleksei-burlakov this is on the right track for the version number, but we need to add something that verifies that this is SLES4SAP and not another distribution!

examples/azure-rules-perf-optimized.yaml
stefanotorresi, line 112: Missing reference:
```suggestion
          ## Reference
          https://documentation.suse.com/en-us/sbp/all/single-html/SLES4SAP-hana-sr-guide-PerfOpt-15/#cha.hana-sr.scenario
```

stefanotorresi (changes requested):
Getting there!

examples/azure-rules-perf-optimized.yaml
aleksei-burlakov: This check was added in #30 where we check the product. This PR is only aimed at the version.

stefanotorresi: Ohh, okay!

stefanotorresi (approved):

PR #55, opened 2021-05-06
warn if there are not 3 sbd devices

aleksei-burlakov:
It was decided during the HA CHecker features review on 05.05.2021 that
we warn the not only when there are <3 sbd devices,
but also if they are > 3.

### DIFF
--- a/examples/azure-rules.yaml
+++ b/examples/azure-rules.yaml
@@ -1,4 +1,4 @@
----
+
 controls:
 version: 1.0.0
 id: 1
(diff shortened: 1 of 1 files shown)


examples/azure-rules.yaml
```diff
@@ -282,7 +282,7 @@ groups:
           test_items:
             - flag: "SBD_COUNT"
               compare:
-                op: gte
+                op: eq
                 value: 2
```
dirkmueller, line 286: value should be raised to 3

examples/azure-rules.yaml
```diff
@@ -282,7 +282,7 @@ groups:
...
         remediation: |
           It is recommended to configure up to 3 SBD devices for production environments.
```
dirkmueller, line 288: remediation needs to be adjusted (3 not up to 3)

aleksei-burlakov (commented):
Fixed

dirkmueller (approved):

PR #58, opened 2021-05-18
Unify code to the internal folder

arbulu89:
3 major changes here:

- Move the code logic from `web` to the internal packages, to have a shared point for the code, that later is used in the `web` and the `agent`. It doesn't have almost any new feature addition more than rewritting the code and adding many UTs.
- Improve the used methods in `internal/consul/kv.go`. Now, our `ListMap` and `PutMap` methods are much better. They can handle more data types (slices, integers, booleans, etc) and it is much more tested.
- Additionally, I have added the `SAP layout` to the `SAP system view` (`sapcontrol -function GetSystemInstanceList` output basically). It looks like this:
image

### DIFF
--- a/internal/environments/environments.go
+++ b/internal/environments/environments.go
@@ -1,0 +1,32 @@
+package environments
+
+import (
+  "github.com/trento-project/trento/internal/consul"
+)
+
+type Environment struct {
+	Name       string                `mapstructure:"name,omitempty"`
+	Landscapes map[string]*Landscape `mapstructure:"landscapes,omitempty"`
+}
+
+type Landscape struct {
+	Name       string                `mapstructure:"name,omitempty"`
+	SAPSystems map[string]*SAPSystem `mapstructure:"sapsystems,omitempty"`
+}
+
+type SAPSystem struct {
+	Name  string `mapstructure:"name,omitempty"`
+}
+
+func NewEnvironment(env, land, sys string) Environment {
+  newSys := &SAPSystem{Name: sys}
+  systemsMap := make(map[string]*SAPSystem)
+  systemsMap[sys] = newSys
+
+  newLand := &Landscape{Name: consul.KvUngrouped, SAPSystems: systemsMap}
+  landsMap := make(map[string]*Landscape)
+  landsMap[land] = newLand
+
+  newEnv := Environment{Name: env, Landscapes: landsMap}
+  return newEnv
+}
--- a/internal/environments/environments.go
+++ b/internal/environments/environments.go
@@ -1,8 +1,8 @@
 package environments
 
 import (
-  "github.com/trento-project/trento/internal/consul"
-  "github.com/trento-project/trento/internal/hosts"
+	"github.com/trento-project/trento/internal/consul"
+	"github.com/trento-project/trento/internal/hosts"
 )
 
 type Environment struct {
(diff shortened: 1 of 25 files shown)


internal/environments/environments.go
```diff
@@ -0,0 +1,110 @@
...
+		HealthMap: make(map[string]string),
+	}
+
+	for _, host := range s.Hosts {
+		h := host.Health()
+		health = health.updateHealth(host.Name(), h)
+	}
+
+	return health
+}
```
stefanotorresi, line 110: There is an extreme degree of repetition here. Have you thought about generalizing getting the name of an entity and their health status into some interfaces?

Also, what happens if there is a clash between host names, system names, and landscape names? We are using the same map for all of these three, and I think this might potentially cause problems.

internal/environments/environments.go
```diff
@@ -0,0 +1,110 @@
...
+	return ungrouped(e.Name)
+}
+
+func (l *Landscape) Ungrouped() bool {
+	return ungrouped(l.Name)
+}
+
+func (s *SAPSystem) Ungrouped() bool {
+	return ungrouped(s.Name)
+}
```
stefanotorresi, line 51: I'm not quite clear why do we need this `ungrouped` fake group. Can we do without it somehow?
Maybe we could just assign every node to a `default` environment and landscape, and treat it like any other actual value. I think there is no need to have a special one with conditional around it: we just need to assign one by default.
This would also have the advantage that in the absence of user input, we can always show something in the environments and landscape lists anyway.

internal/environments/metadata.go
```diff
@@ -0,0 +1,13 @@
...
+	Cluster     string `mapstructure:"ha-cluster,omitempty"`
+	Environment string `mapstructure:"sap-environment,omitempty"`
+	Landscape   string `mapstructure:"sap-landscape,omitempty"`
+	SAPSystem   string `mapstructure:"sap-system,omitempty"`
+}
+
+func NewMetadata() Metadata {
+	var metadata = Metadata{}
+	return metadata
+}
```
stefanotorresi, line 13: This function doesn't do any initialization logic: calling `new(Metadata)` or assigning `Metadata{}` to a variable would be identical. Maybe we can just remove it?

internal/environments/metadata.go
```diff
@@ -0,0 +1,13 @@
+package environments
+
+type Metadata struct {
+	Cluster     string `mapstructure:"ha-cluster,omitempty"`
+	Environment string `mapstructure:"sap-environment,omitempty"`
+	Landscape   string `mapstructure:"sap-landscape,omitempty"`
+	SAPSystem   string `mapstructure:"sap-system,omitempty"`
+}
```
stefanotorresi, line 8: Isn't this node metadata, rather than environment?

stefanotorresi (changes requested):
Quite a lot to go through here, it's very substantial!
Well done so far, but there is room for a couple of improvements.

internal/environments/metadata.go
arbulu89: Ok, yes, it makes sense. I will move it to the `hosts` package (if there is not any strange import issue at least)

internal/environments/environments.go
arbulu89: This is a business decision I guess. I'm fine with any of the options. @lee-martin What do you think?
As a context, now new discovered `SAP systems` are added to a environmend and landscape named `ungrouped`, which means that they don't belong to any particular env/land. I use this to have a different treatment in the frontend. 
If you prefer a default Env/Land it is fine

internal/environments/environments.go
arbulu89: Actually, I just copy/pasted already existing code, so I didn't even have a look. Anyway, I did myself this code in the past. 
I will try to make it more reusable

internal/environments/environments.go
stefanotorresi: @lee-martin are you okay with displaying Systems under a `default` Environment and Landscape, in the absence of user input?

internal/environments/environments.go
arbulu89: About the last comment about a potential clash, you are right. I didn't consider this scenario. I will need to think about it, maybe having some kind of nested map or similar. Let's see what I can think of

internal/environments/environments.go
lee-martin: @arbulu89  & @stefanotorresi  oh wow, this is quite difficult, I can also see both points of view here as well. I would go with @stefanotorresi and the idea to put hosts into some default environment and landscape here, since I agree that the empty pages would indeed be confusing.  What would be the name of the default environment and landscapebe? just call it "default" ?
In any event, we need to see and use it, then we can iterate and improve.

stefanotorresi: Yes, my idea is to just call it `default`, and then in the future we give users the ability to rename it and/or create new ones and reassign systems to them.

arbulu89: Ok, let's do it out of this PR, as it would require changes in many files (specially frontend templates) and It would grow up much more the PR

internal/environments/environments.go
arbulu89: Hey @stefanotorresi ,
I have done some changes in https://github.com/trento-project/trento/pull/58/commits/8932b53821e42ee50311ac2007c5663691622565

internal/environments/environments.go
stefanotorresi: All right! I will add a follow-up issue for this.

stefanotorresi (approved):

Issue #50, opened 2021-05-05
Use markdown for checks remediation text


stefanotorresi:
This is the current status of the checks visualization:

![Screenshot from 2021-04-29 15-09-50](https://user-images.githubusercontent.com/2952427/117124937-383c2780-ad99-11eb-9b58-dcc49d9bf808.png)

We need to make the remediation text (the content of the fourth column in the picture above) more usable.

Since the source of the remediation text comes from a text field in a yaml file, we can use Markdown to enrich it with some details. For the time being, we can start with the SBD checks as an example; we'll then apply the same formatting to the remaining checks, progressively.

Contextually, we also need to do the following:
- establish a common Markdown template that all the checks should follow, so that the remediation text always have all similar and predictable structure.
- update the web ui so that we can render this markdown to HTML in the checks page. The content of the remediation can be extensive, so we need to accomodate for this with a different layout: we previously used [Bootstrap Collapse](https://getbootstrap.com/docs/4.6/components/collapse/) to allow the user to click on a check row showing just the id, description and status, and then the remediation would appear below. Feel free to use any other method, and bring the UI/UX team in making a decision here.

Here is an example of Markdown template for the remediation structure:
```markdown
## Abstract
// One or two paragraphs from the upstream source document that establishes the best practice we're advising users to follow.

## Remediation
// A step by step guide to remediate a check failure, if at all possible

## References
// a list of links to upstream documentation
- link
- link 2
```

PR #62, opened 2021-05-21
Use markdown for checks remediation text

aleksei-burlakov:
This PR is for solving #50. It's only the first iteration.
This commit parses md --> html.

It was prepared during a paring programming session together with
Rubén Torrero Marijnissen <rtorreromarijnissen@suse.com>

### DIFF
--- a/web/layout.go
+++ b/web/layout.go
@@ -8,6 +8,10 @@ import (
 	"path/filepath"
 
 	"github.com/gin-gonic/gin/render"
+
+	"github.com/gomarkdown/markdown"
+	"github.com/gomarkdown/markdown/html"
+	"github.com/gomarkdown/markdown/parser"
 )
 
 // LayoutRender wraps user templates into a root one which has it's own data and a bunch of inner blocks
@@ -92,6 +96,19 @@ func (r *LayoutRender) addFileFromFS(templatesFS fs.FS, file string) {
 		},
 		"sum": func(a int, b int) int {
 			return a + b
+		},
+		"markup": func(md string) template.HTML {
+			extensions := parser.CommonExtensions | parser.AutoHeadingIDs
+			p := parser.NewWithExtensions(extensions)
+
+			htmlFlags := html.CommonFlags | html.HrefTargetBlank
+			opts := html.RendererOptions{Flags: htmlFlags}
+			renderer := html.NewRenderer(opts)
+
+			// html := markdown.ToHTML([]byte(md), nil, nil) // could be as simple as this
+			html := markdown.ToHTML([]byte(md), p, renderer)
+			result := template.HTML(html)
+			return result
 		},
 	})
 	patterns := append([]string{r.root, file}, r.blocks...)
--- a/web/templates/ha_checks.html.tmpl
+++ b/web/templates/ha_checks.html.tmpl
@@ -21,7 +21,7 @@
                     <td>{{ $Section }}</td>
                     <td>{{ .ID }}</td>
                     <td>{{ .Description }}</td>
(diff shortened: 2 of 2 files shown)


arbulu89 (commented):
@aleksei-burlakov It looks good!
Could you add a picture to show us how it looks like? Maybe with some real MD example in the check file.

rtorrero:
Thanks Aleksei! It's looking good.

I made a quick test with this PR adding code-block marks in `azure-rules.yml` for the shell commands in the remediation text and it definitely looks much better though I see something weird with multi-line code blocks, which appear in a very unreadable yellow / golden color. Single line code-blocks seem to readable.

rtorrero:
For a next iteration we need to look into making the longest remediation messages collapsed by default with maybe bootstrap collapse as @stefanotorresi suggested in issue#50

web/layout.go
```diff
@@ -93,6 +97,19 @@ func (r *LayoutRender) addFileFromFS(templatesFS fs.FS, file string) {
 		"sum": func(a int, b int) int {
 			return a + b
 		},
+		"markup": func(md string) template.HTML {
```
stefanotorresi, line 100: I guess the name of the function should be `markdown`? :)

web/layout.go
```diff
@@ -93,6 +97,19 @@ func (r *LayoutRender) addFileFromFS(templatesFS fs.FS, file string) {
...
+			extensions := parser.CommonExtensions | parser.AutoHeadingIDs
+			p := parser.NewWithExtensions(extensions)
+
+			htmlFlags := html.CommonFlags | html.HrefTargetBlank
+			opts := html.RendererOptions{Flags: htmlFlags}
+			renderer := html.NewRenderer(opts)
+
+			// html := markdown.ToHTML([]byte(md), nil, nil) // could be as simple as this
```
stefanotorresi, line 108: I don't think we need to leave this commented line behind.

web/templates/ha_checks.html.tmpl
```diff
@@ -21,7 +21,7 @@
                     <td>{{ $Section }}</td>
                     <td>{{ .ID }}</td>
                     <td>{{ .Description }}</td>
-                    <td>{{ index .TestInfo 0 }}</td>
+                    <td>{{ markup ( index .TestInfo 0 ) }}</td>
```
stefanotorresi, line 24: see above suggestion about renaming the template function

web/layout.go
```diff
@@ -93,6 +97,19 @@ func (r *LayoutRender) addFileFromFS(templatesFS fs.FS, file string) {
...
+			html := markdown.ToHTML([]byte(md), p, renderer)
+			result := template.HTML(html)
+			return result
```
stefanotorresi, line 111: The `html` variable here collides with the imported `github.com/gomarkdown/markdown/html` package.
Maybe rename it as follows?
```suggestion
			rendered := markdown.ToHTML([]byte(md), p, renderer)
			result := template.HTML(rendered)
			return result
```

stefanotorresi (changes requested):
Good job guys! We just need a few tweaks.

Other than the ones below, you need to update the `go.mod` and `go.sum` files after having run `go get`, otherwise the module is not declared in the project dependencies! (see the tests failing due to this)

I would also like to see a unit test of the template function; to that end, I would suggest to move the entire anonymous function to a private one in the same `layout.go` file, maybe named `func markdownToHtml(md string) template.HTML`, so that we can test it in isolation.

stefanotorresi:
Since you folks are out the whole week I took the liberty to do the changes and merge this manually.

PR #63, opened 2021-05-24
Implement build CI step

stefanotorresi:
This PR does the following:

- adds a `cross-compiled` make target to build binaries in all architectures
- renames the `test` job to `build`
- integrates it with GH actions build artifacts
- adds the `upload-release-assets` job

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,20 +1,34 @@
-on: [push, pull_request]
+on:
+  push:
+  pull_request:
+  release:
+    types: [ published, created, edited ]
 
 jobs:
-  test:
+  build:
     runs-on: ubuntu-20.04
     steps:
+    - uses: actions/checkout@v2
     - uses: actions/setup-go@v2
       with:
         go-version: ^1.16
-      id: go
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yaml
```diff
@@ -23,3 +37,28 @@ jobs:
...
+
+  upload-release-assets:
+    needs: build
+    if: github.event.release
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/download-artifact@v2
+        with:
+          name: trento-binaries
+      - uses: AButler/upload-release-assets@v2.0
```
arbulu89, line 61: Is this `AButler` `AB` in capital letters in purpose? Just in case if its a typo (most probably not, but better to double check hehe)

arbulu89 (approved):
Looks good. If you tested in your branch and works, green light from me

.github/workflows/ci.yaml
stefanotorresi: Yep, it's capitalized! https://github.com/AButler/upload-release-assets
I guess it would work regardless, but hey,

PR #64, opened 2021-05-24
cluster discovery and visualization

arbulu89:
Cluster visualization changes.
Major features:

- Cluster discovery and storage in the KV. It will store everything coming from `cibAdmin` and `crm_mon`.
- Cluster data visualization
- Remove many not used methods from the `discovery` interface

Pending to do:
- The clusters are identified by the cluster name, which is not unique and even mandatory. We will need to find a unique id (`csync` auth key?)
- The clusters health status in the main `clusters` page is hardcoded to `passing` by now.
- There is not any lock mechanism, so time to time the clusters page data is shown with empty data (just the moment where the agent is writing and the web loading). This will be fixed in a next PR. The code is ready, but I preferred to have a smaller PR.

Some pics:
image
image


web/templates/cluster.html.tmpl
```diff
@@ -3,43 +3,88 @@
     <h6><a href="/clusters">Clusters</a> > {{ .Cluster.Name }}</h6>
     <h1>Cluster details</h1>
     <dl class="inline">
-      <dt class="inline">Name</dt>
-      <dd class="inline">{{ .Cluster.Name }}</dd>
       <dt class="inline">Hosts number</dt>
-      <dd class="inline">{{ len .Hosts  }}</a></dd>
+      <dd class="inline">{{ .Cluster.Crmmon.Summary.Nodes.Number }}</a></dd>
```
stefanotorresi, line 7: Hanging close tag here!
```suggestion
      <dd class="inline">{{ .Cluster.Crmmon.Summary.Nodes.Number }}</dd>
```

web/templates/cluster.html.tmpl
```diff
@@ -3,43 +3,88 @@
...
       <dt class="inline">Resource number</dt>
-      <dd class="inline">4</a></dd>
+      <dd class="inline">{{ .Cluster.Crmmon.Summary.Resources.Number }}</a></dd>
```
stefanotorresi, line 9: Here too:
```suggestion
      <dd class="inline">{{ .Cluster.Crmmon.Summary.Resources.Number }}</dd>
```

stefanotorresi (changes requested):
Looks mostly good, other than two minor typo in the HTML template.

If there is something that I really don't like atm, is the fact that we use the `Discover` verb instead of the `Discovery` noun for implementations of the `Discoverer` interface.
`Discover` was originally intended to be used for a function type, and now it's just the name of the main method of `Discoverer`.
This is very subjective, but I would rather use `Discovery` instead of `Discoverer`, and rename everything with the `Discover` suffix to `Discovery`,  except the `Discover()` method of the interface.
IOW, `Discover` should never be an object, an interface, or a struct. Having something like `discover.Discover()` is really ugly. :-[

That said, this is mostly a naming issue, we can solve this in another issue.

arbulu89:
@stefanotorresi Fixes done.
I have renamed the `discover` package to `discovery` (a noun looks better), and within this, `discoverer` to `discovery`, and the rest of the structs of this type follow the same convention.
Change in this commit: https://github.com/trento-project/trento/pull/64/commits/b4666fbf4760bb25fc7f960a8fce231a96e91696

stefanotorresi (approved):
🔝 !!

PR #65, opened 2021-05-26
Agent tweaks

### LINKED ISSUE #35: Embed the consul configuration template and provide default input related to the consul-template daemon
Right now, if one brings the `trento` binary in a node, they also need to bring the `examples/trento-config.tpl` file with it.

This is not ideal: we should remove the `examples/trento-config.tpl` as a default value, because this path only exists in the sources. Instead of this, we should provide the content of this file by default, embedding its content somehow (either via a string template or a Go embedded FS).

In general, we need to hide as much as possible of Consul details to the user, because we don't want to this cognitive load to leak.

Even the configuration directory should probably become something specific to our project instead of Consul. (e.g. in the future we would put all the things consul related into something like `~/.config/trento/consul`)

stefanotorresi:
- Renames Consul checks to make them more consistent, and group them all under an hard-coded `trento-agent` service name.
- Removes the `--service-name` flag, as there is no actual use for it.
- Updates README.md accordingly, plus add a note about Consul `-dev` mode, which doesn't work due to our need to run `consul reload` to reload node metadata , which in dev mode clears all the registered health checks.
- Better logging in the discovery TTL loop.
- Removes the `--consul-template` agent flag and make the template embedded. (fixes #35)
- Rename the `--config-dir` flag to `--consul-config-dir`.
- Extract agent ticker loops into a common `repeat()` function that always runs the first tick immediately.
- Simplify Agent goroutines error handling


stefanotorresi:
@trento-project/developers FYI. Can't wait for next week, sorry. 😛

PR #60, opened 2021-05-19
Add deploy.sh script helper

rtorrero:
A small simple script to assist in deploying trento for dev/testing purposes. Enjoy!

PR #66, opened 2021-05-26
Allow to override TARGET_USER

rtorrero:
A follow up to #60 to allow to override the default "root" user which is something that will very likely be needed.

### DIFF
--- a/hack/deploy.sh
+++ b/hack/deploy.sh
@@ -1,14 +1,14 @@
 #!/bin/bash
 TARGET_HOST=$1
 CONSUL_HOST=$2
-ACTION=$3
-TRENTO_BIN="trento"
-CONSUL_BIN="consul"
-TRENTO_PATH="/srv/trento/"
-CONSUL_PATH="/srv/consul/"
-EXAMPLES_PATH="$TRENTO_PATH/examples"
-TARGET_USER="root"
-CONSUL_DATA_DIR="consul-agent-data"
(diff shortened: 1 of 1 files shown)


stefanotorresi:
@rtorrero you can do this much more succinctly by using `:-`
E.g.
```bash
TARGET_USER=${TARGET_USER:-"root"}
```

see https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Shell-Parameter-Expansion for more details

stefanotorresi (changes requested):
@rtorrero you can do this much more succinctly by using `:-`
E.g.
```bash
TARGET_USER=${TARGET_USER:-"root"}
```

see https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Shell-Parameter-Expansion for more details

stefanotorresi:
Oh, btw, I forgot to mention that if we're not using root then we need to use `sudo` when starting `trento agent`!

rtorrero:
Yes, using parameter expansion seems much more elegant here. I've also added a $ADD_PARAMS with `sudo` when $TARGET_USER is not root. I belive that's the only place where it's required, right?

Let me know if it looks better and I'll squash it before merge

stefanotorresi (dismissed):

stefanotorresi:
Perfect! You can rebase and merge.

stefanotorresi (approved):

PR #70, opened 2021-05-28
add sap systems to default environment and landscape in absence of one

### LINKED ISSUE #61: Remove the `ungrouped` fake group and replace it with an actual `default`.
As discussed in in https://github.com/trento-project/trento/pull/58#discussion_r635065364:

We want to assign SAP Systems to a `default` Landscape and Environment OOTB, so that we provide users a baseline that makes the list pages work without their explicit input.

stefanotorresi:
This PR mainly closes #61 and perform various related changes, plus some things I just stumbled upon.

- The `sapsystem` discovery loop now creates `ḍefault` Environment and Landscapes in case there is none to associate to the discovered systems.
- The constructors of the structs in the `environments` package are now separated.
- A `GetSID()` method has been added to the `sapsystem.SAPSystem` struct.
- To better distinguish from the above, `SAPSystem.getSapSystemId()` is no longer a method, it's now a private function named `getUniqueId(sid string)`.
- Some HTTP handler constructors were renamed because they were all over the place.
- The HTTP error handling logic is now in a dedicated `errors.go` file under the `web` package.
- The `ErrorHandler` middleware now handles a custom type of error `HttpError`, that allows to use different templates for different errors.
- The `/sapsystems/:sid` handler now handles 404 errors instead of causing nil pointer dereferences.
- `LayoutRender` now panics earlier with a meaningful message when Gin tries to render an non-existing template.
- All the custom logic applied to `ungrouped` things has been removed.
- The `/sapsystems/:sid` handler now passes all the relevant full structs to the template, instead of a mix of structs and strings.
- Various coding style changes my IDE applied automagically (e.g. unnecessary type hints on map/array children).
- New `make test-coverage` target.

P.S. by working a little bit with the various `internal` packages, I must say it's not exactly easy to understand why some things are the way they are; we need a live chat to point out a few things and start working on making the model a little cleaner.


web/environments.go
```diff
@@ -38,9 +37,9 @@ func NewEnvironmentListHandler(client consul.Client) gin.HandlerFunc {
 	}
 }
 
-func NewLandscapesListHandler(client consul.Client) gin.HandlerFunc {
+func NewLandscapeListHandler(client consul.Client) gin.HandlerFunc {
```
arbulu89, line 40: It is not exactly about this line, but we still have `NewEnvironmentsListHandler` above. The follow the same convention, it should be `NewEnvironmentListHandler` or this one `NewLandscapesListHandler, isn't it?

Below we have `NewSAPSystemsListHandler` with a `s` too

arbulu89 (approved):
@stefanotorresi Thanks! Looks good.

By the way, happy you realized that the code needs some clarifications and updates. As commented multiple times, we are investing much more time in planning and refinement meetings than in technical reviews and meetings hehe

arbulu89 (approved):
Nice @stefanotorresi

PR #72, opened 2021-06-01
sbd discovery

arbulu89:
SBD data discovery. It is used within the `cluster` discovery loop.
The data is stored in `trento/v0/clusters/$cluster/sbd`.
The major part of the logic is in `internal/cluster/sbd.go`

The stored data:
- `/etc/sysconfing/sbd` conent
- sbd dump command output
- sbd list command output

It brings some new views in the cluster view too:
image
image
image


test/sbd_config
```diff
@@ -0,0 +1,115 @@
+## Type: string
```
rtorrero, line 1: I'm guessing these are here for testing purposes? If so, I'd strip all comments from the original file and add some brief ones explaining what use case this config is testing

internal/cluster/sbd.go
```diff
@@ -0,0 +1,210 @@
...
+	defer sbdConfFile.Close()
+
+	sbdConfigRaw, err := ioutil.ReadAll(sbdConfFile)
+
+	if err != nil {
+		return nil, fmt.Errorf("could not read sbd config file %s", err)
+	}
+
+	// Loop through sbd list output and find for matches
+	r := regexp.MustCompile(`(\S+)\s*=(\S+)`)
```
arbulu89, line 99: The next regexp should remove the lines starting with a comment:
`^\s*(\w+)\s=\s(\S+)$`

arbulu89:
@stefanotorresi I have changed the code for:
1. Improve the regular expression to cover all the use cases
2. Load partial information in case of failure (this might need some changes in the future, as I don't exactly know all the use cases, but now it sends all the available data). I bet for let it as it is until we have some real users feedback

test/sbd_config
arbulu89: Hi @rtorrero ,
I have decided the let the test files as they are. Using exact real use cases help to find strange errors.
About the description, I think that the file names are enough meaningful

web/templates/cluster.html.tmpl
```diff
@@ -7,13 +7,36 @@
...
+        <div class="alert-title">Fencing resource not found</div>
+        There is not any fencing resource configured. A fencing resource is mandatory for a correct functioning of the cluster.</br>Find more information <a href="https://documentation.suse.com/sle-ha/15-SP1/single-html/SLE-HA-guide/#sec-ha-requirements-hw" target="_blank">here</a>.
+      </div>
+    </div>
+    {{- end }}
+    {{- if .Cluster.IsFencingSBD }}
+    {{ template "sbd" .Cluster.SBD }}
     <hr/>
     <p class='clearfix'/>
+    {{- end }}
```
stefanotorresi, line 39: Just a suggestion: I would put the entire SBD section after the Nodes one, which I think is the most important one.

stefanotorresi (changes requested):

aleksei-burlakov:
@arbulu89 you are possibly already keeping it in mind, if not, then it must not necessarily be a disk-based sbd, it could also be a qdevice or a diskless sbd when 3 nodes.

web/templates/cluster.html.tmpl
arbulu89: Done: https://github.com/trento-project/trento/pull/72/commits/adbc734cbac7bf9168c7e68e0904f8a6f21eb731

internal/cluster/sbd.go
```diff
@@ -0,0 +1,220 @@
...
+	sbdDumpStr := string(sbdDump)
+
+	dump.Header = assignPatternResult(sbdDumpStr, `Header version *: (.*)`)[1]
+	dump.Uuid = assignPatternResult(sbdDumpStr, `UUID *: (.*)`)[1]
+	dump.Slots, _ = strconv.Atoi(assignPatternResult(sbdDumpStr, `Number of slots *: (.*)`)[1])
+	dump.SectorSize, _ = strconv.Atoi(assignPatternResult(sbdDumpStr, `Sector size *: (.*)`)[1])
+	dump.TimeoutWatchdog, _ = strconv.Atoi(assignPatternResult(sbdDumpStr, `Timeout \(watchdog\) *: (.*)`)[1])
+	dump.TimeoutAllocate, _ = strconv.Atoi(assignPatternResult(sbdDumpStr, `Timeout \(allocate\) *: (.*)`)[1])
+	dump.TimeoutLoop, _ = strconv.Atoi(assignPatternResult(sbdDumpStr, `Timeout \(loop\) *: (.*)`)[1])
+	dump.TimeoutMsgwait, _ = strconv.Atoi(assignPatternResult(sbdDumpStr, `Timeout \(msgwait\) *: (.*)`)[1])
```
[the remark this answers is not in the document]
arbulu89: Done: https://github.com/trento-project/trento/pull/72/commits/5ac7f046d46e3c596b659508f530d80653269415

stefanotorresi (approved):

stefanotorresi:
Have some 🐝

PR #74, opened 2021-06-02
merge all checks in 1.6 into one 1.6.1

aleksei-burlakov:


arbulu89 (commented):
@aleksei-burlakov I'm a bit out of the conversation, but this check doesn't look for the basic sudoers configuration, it looks actually the basic and the advanced together.

The users will have whether basic or advanced option, not both (it doesn't make much sense at least, as the basic is a combination of the others).

So:
- The description should be changes, as this doesn't apply to the basic configuration
- If we want to have the advanced option, I would remove the first `cat/grep` entry, and only let the remaining 5

PD: In our deployment, we only enable the basic option in fact, so this check would raise a warning.

aleksei-burlakov:
> @aleksei-burlakov I'm a bit out of the conversation, but this check doesn't look for the basic sudoers configuration, it looks actually the basic and the advanced together.
> 
> The users will have whether basic or advanced option, not both (it doesn't make much sense at least, as the basic is a combination of the others).
> 
> So:
> 
> * The description should be changes, as this doesn't apply to the basic configuration
> * If we want to have the advanced option, I would remove the first `cat/grep` entry, and only let the remaining 5
> 
> PD: In our deployment, we only enable the basic option in fact, so this check would raise a warning.

Thank you for reviewing, it's indeed the wrong description, this check is not only for basic sudoers

examples/azure-rules.yaml
```diff
@@ -756,63 +756,25 @@ groups:
...
+          cat /etc/sudoers | grep -E "^Cmnd_Alias +SFAIL_SITEA += +/usr/sbin/crm_attribute" | grep -E "\-n +hana_[a-z][a-z0-9]{2}_site_srHook_.+" | grep -E "\-v +SFAIL" | grep -E "\-t +crm_config" | grep -E "\-s +SAPHanaSR" | sed "s/.*/true3/"
+          cat /etc/sudoers | grep -E "^Cmnd_Alias +SOK_SITEB += +/usr/sbin/crm_attribute" | grep -E "\-n +hana_[a-z][a-z0-9]{2}_site_srHook_.+" | grep -E "\-v +SOK" | grep -E "\-t +crm_config" | grep -E "\-s +SAPHanaSR" | sed "s/.*/true4/"
+          cat /etc/sudoers | grep -E "^Cmnd_Alias +SFAIL_SITEB += +/usr/sbin/crm_attribute" | grep -E "\-n +hana_[a-z][a-z0-9]{2}_site_srHook_.+" | grep -E "\-v +SFAIL" | grep -E "\-t +crm_config" | grep -E "\-s +SAPHanaSR" | sed "s/.*/true5/"
+          cat /etc/sudoers | grep -E "[a-z][a-z0-9]{2}adm +ALL=\(ALL\) +NOPASSWD:" | sed "s/.*: //" | grep SOK_SITEA | grep SFAIL_SITEA | grep SOK_SITEB | grep SFAIL_SITEB | sed "s/.*/true6/"
         tests:
           test_items:
-          - flag: "true"
+          - flag: "true1"
```
dirkmueller, line 769: I think this is missing a "set: true" to ensure that missing the output will not pass the check. 

see https://github.com/aquasecurity/bench-common/blob/main/docs/README.md#check

dirkmueller (changes requested):
see inline comment. when I'm wrong on that reading of the documentation, please don't hesitate to tell me :)

examples/azure-rules.yaml
aleksei-burlakov: @dirkmueller  What they mean in https://github.com/aquasecurity/bench-common/blob/main/docs/README.md#check is that they check a specific flag that it equals (eq) to some value like `--flag=value`. In this code I check that those flags exist at all and they exist alltogether. My check is somewhat like here https://github.com/trento-project/trento/blob/main/examples/azure-rules.yaml#L446-L453
I haven't specified that `bin_op: and` because it's the default.

stefanotorresi:
I'm not sure we're actually implementing the required logic here.
A valid configuration has to _either_ pass what previously was in 1.6.1 _or_ pass all the other ones. It looks to me that we are just merging all the sudoers config lines with an `and` operation instead, while it should be kind of `(1.6.1) or (1.6.2 and 1.6.3 and 1.6.4 and 1.6.5 and 1.6.6)`; I don't see this `or` operation in the bash script, or am I missing something?

aleksei-burlakov:
> I'm not sure we're actually implementing the required logic here.
> A valid configuration has to _either_ pass what previously was in 1.6.1 _or_ pass all the other ones. It looks to me that we are just merging all the sudoers config lines with an `and` operation instead, while it should be `(1.6.1) or (1.6.2 and 1.6.3 and 1.6.4 and 1.6.5 and 1.6.6)`; I don't see this `or` operation in the bash script, or am I missing something?

Good catch, now it doesn't require them both to be configured.

examples/azure-rules.yaml
dirkmueller: ah, I missed the `bin_op: and` being default, thank you for the explanation!

aleksei-burlakov:
> > I'm not sure we're actually implementing the required logic here.
> > A valid configuration has to _either_ pass what previously was in 1.6.1 _or_ pass all the other ones. It looks to me that we are just merging all the sudoers config lines with an `and` operation instead, while it should be `(1.6.1) or (1.6.2 and 1.6.3 and 1.6.4 and 1.6.5 and 1.6.6)`; I don't see this `or` operation in the bash script, or am I missing something?
> 
> Good catch, now it doesn't require them both to be configured.

And special apologies to @arbulu89 for my ignorance.

examples/azure-rules.yaml
```diff
@@ -756,63 +756,28 @@ groups:
...
+            exit 0
+          fi
+          cat /etc/sudoers | grep -E "^Cmnd_Alias +SOK_SITEA += +/usr/sbin/crm_attribute" | grep -E "\-n +hana_[a-z][a-z0-9]{2}_site_srHook_.+" | grep -E "\-v +SOK" | grep -E "\-t +crm_config" | grep -E "\-s +SAPHanaSR" | sed "s/.*/true2/"
+          cat /etc/sudoers | grep -E "^Cmnd_Alias +SFAIL_SITEA += +/usr/sbin/crm_attribute" | grep -E "\-n +hana_[a-z][a-z0-9]{2}_site_srHook_.+" | grep -E "\-v +SFAIL" | grep -E "\-t +crm_config" | grep -E "\-s +SAPHanaSR" | sed "s/.*/true3/"
+          cat /etc/sudoers | grep -E "^Cmnd_Alias +SOK_SITEB += +/usr/sbin/crm_attribute" | grep -E "\-n +hana_[a-z][a-z0-9]{2}_site_srHook_.+" | grep -E "\-v +SOK" | grep -E "\-t +crm_config" | grep -E "\-s +SAPHanaSR" | sed "s/.*/true4/"
+          cat /etc/sudoers | grep -E "^Cmnd_Alias +SFAIL_SITEB += +/usr/sbin/crm_attribute" | grep -E "\-n +hana_[a-z][a-z0-9]{2}_site_srHook_.+" | grep -E "\-v +SFAIL" | grep -E "\-t +crm_config" | grep -E "\-s +SAPHanaSR" | sed "s/.*/true5/"
+          cat /etc/sudoers | grep -E "[a-z][a-z0-9]{2}adm +ALL=\(ALL\) +NOPASSWD:" | sed "s/.*: //" | grep SOK_SITEA | grep SFAIL_SITEA | grep SOK_SITEB | grep SFAIL_SITEB | sed "s/.*/true1/"
```
arbulu89, line 769: It won't change anything but it looks strange to have `true1` printed last instead of first hehe

arbulu89 (approved):
Thanks @aleksei-burlakov ,
It looks good to me

dirkmueller:
> I'm not sure we're actually implementing the required logic here.
> A valid configuration has to _either_ pass what previously was in 1.6.1 _or_ pass all the other ones. It looks to me that we are just merging all the sudoers config lines with an `and` operation instead, while it should be `(1.6.1) or (1.6.2 and 1.6.3 and 1.6.4 and 1.6.5 and 1.6.6)`; I don't see this `or` operation in the bash script, or am I missing something?

the previous and new 1.6.1 check echo true1 .. true5, so it does short circuit on evaluating 1.6.2-1.6.6. for me this looks correctl. what an I missing  @stefanotorresi ?

dirkmueller (approved):

stefanotorresi:
It wasn't like that previously. Now it's okay! :)

PR #77, opened 2021-06-03
WIP: Run consul agent within trento

rtorrero:
This is an early attempt at getting consul running within trento. Right now we are cutting many corners to get this working and it has a few limitations, most importantly:

- the `consul-agent-data` directory needs to be created in the working directory where trento runs from. This is hardcoded and will be changed later, but the most important issue is that it needs to be manually cleaned on every run. Still trying to figure out why
- `tls` has been disabled. But since it's communicating through localhost it might not be a minor problem. Still need to understand better the implications of this.
- I had to introduce a *very ugly* 60 sleep in agent/agent.go:139  so that it would not give up connecting to consul too early.
- I had to fork github.com/aquasecurity/bench-common to fix a dependency problem

The code needs to be cleaned and re-thought, so any suggestion is welcome on how to implement this better.


agent/embeddedconsul.go
```diff
@@ -0,0 +1,113 @@
...
+
+func NodeID() string {
+	id, err := uuid.GenerateUUID()
+	if err != nil {
+		panic(err)
+	}
+	return id
+}
+
+func randomPortsSource() (data string) {
```
rtorrero, line 25: Don't pay too much attention to this and TestConfigHCL(); it was taken from consul's testagent.go as a quick hack to get it running. This will go.

agent/agent.go
```diff
@@ -101,8 +112,35 @@ func DefaultConfig() (Config, error) {
...
+
 // Start the Agent which includes the registration against Consul Agent
 func (a *Agent) Start() error {
+	var wg sync.WaitGroup
+
+	// This number must match the number threads attached to the WaitGroup object
+	wg.Add(1)
+	errs := make(chan error, 4)
+
+	go func(wg *sync.WaitGroup) {
```
arbulu89, line 136: Do we want to run this in a go routine? For me, it looks like this actually should be something blocking. Nothing else should be executed unless consul is up and running.
That's why actually you are force to put that `sleep`.
I would be consul api has some method to check if the agent is running

agent/embeddedconsul.go
```diff
@@ -0,0 +1,113 @@
...
+	bd, err := consulAgent.NewBaseDeps(loader, log.Writer())
+
+	if err != nil {
+		return nil, fmt.Errorf("failed to create base deps: %w", err)
+	}
+
+	return consulAgent.New(bd)
+}
+
+func (a *Agent) startConsulAgent() error {
```
arbulu89, line 97: An `error` is not reported never

agent/embeddedconsul.go
```diff
@@ -0,0 +1,113 @@
...
+
+	consulAgent "github.com/hashicorp/consul/agent"
+	configAgent "github.com/hashicorp/consul/agent/config"
+	"github.com/hashicorp/consul/agent/connect"
+	"github.com/hashicorp/consul/sdk/freeport"
+	"github.com/hashicorp/go-uuid"
+)
+
+func NodeID() string {
+	id, err := uuid.GenerateUUID()
```
arbulu89, line 18: I don't think we should do this. At least in my head, the consul NodeID should be permanent, We will end up using it to store the data related to the nodes in the KV storage.

Maybe we should some machine specific values to create this UUID

agent/embeddedconsul.go
```diff
@@ -0,0 +1,113 @@
...
+	log.Println(portsConfig)
+	d := filepath.ToSlash("./consul-agent-data")
+	hclDataDir := fmt.Sprintf(`data_dir = "%s"`, d)
+	testHCLConfig := TestConfigHCL(NodeID())
+
+	loader := func(source configAgent.Source) (*configAgent.RuntimeConfig, []string, error) {
+		opts := configAgent.BuilderOpts{
+			HCL: []string{testHCLConfig, portsConfig, hcl, hclDataDir},
+		}
+		overrides := []configAgent.Source{
```
arbulu89, line 74: Do we need the overrides here?
Shouldn't we get the information already from the source?
At the end we manage the source too

arbulu89 (commented):
@rtorrero Added some comments.
I cannot understand why you were forced to fork the benchmark package. This looks something related to your environment, because I don't find any relationship between consul agent and this library

agent/agent.go
```diff
@@ -29,6 +30,9 @@ type Agent struct {
 	ctx              context.Context
 	ctxCancel        context.CancelFunc
 	templateRunner   *manager.Runner
+	consulAgent      *consulAgent.Agent
+	consulOverrides  string
```
arbulu89, line 34: I cannot find where `consulOverrides` and `consulDataDir` are used

agent/agent.go
rtorrero: > @rtorrero Added some comments.
> I cannot understand why you were forced to fork the benchmark package. This looks something related to your environment, because I don't find any relationship between consul agent and this library

@stefanotorresi might be able to explain this better as he helped me address this, but in summary, they are requiring `k8s.io/client-go v11.0.0+incompatible` (see https://github.com/aquasecurity/bench-common/blob/main/go.mod) which is incompatible with `v0.16.9` which is what `consul` uses.

dirkmueller: can we fix that in aquasecurity's codebase?

stefanotorresi: we're going to drop this dependency soon anyway

stefanotorresi: @arbulu89 I don't think that would work: `agent.Start(ctx)` will keep running in the background and block until `cancel()` is called on `ctx`, it will not return once started so, if you don't put this in a goroutine, nothing else will ever be started until the consul agent is stopped explicitly.

agent/embeddedconsul.go
```diff
@@ -0,0 +1,103 @@
...
+
+	if err != nil {
+		return nil, fmt.Errorf("failed to create base deps: %w", err)
+	}
+
+	return consulAgent.New(bd)
+}
+
+func (a *Agent) startConsulAgent(status chan int) error {
+	a.consulAgent.Start(context.Background())
```
stefanotorresi, line 90: Shouldn't we pass our `a.ctx` to `Start()` here, instead of a new context?

agent/agent.go
```diff
@@ -101,26 +123,51 @@ func DefaultConfig() (Config, error) {
...
+
+	// Attempt to register against the consul client
+	wg.Add(1)
+	go func(wg *sync.WaitGroup) {
+		log.Println("Registering the agent service with Consul...")
+		defer wg.Done()
+		err := a.startRegisterLoop()
 		if err != nil {
-			log.Println("An error occurred while trying to deregisterConsulService the agent service with Consul:", err)
+			log.Fatal("Error while registering consul")
```
stefanotorresi, line 151: hmm, I think we should just change `Println` with `Fatal` and leave the error message as it was, otherwise we're not reporting the actual error.

In the future, we might need to avoid `Fatal` and properly handle the error in orchestration with the other goroutines, but it's okay for now.

agent/agent.go
```diff
@@ -101,26 +123,51 @@ func DefaultConfig() (Config, error) {
...
 		} else {
-			log.Println("Consul service de-registered.")
+			status <- 1
```
stefanotorresi, line 153: I'm not sure we need this channel at all.

agent/agent.go
```diff
@@ -209,6 +256,21 @@ func (a *Agent) registerConsulService() error {
...
+			log.Println("An error ocurred while attempting to register:", err)
+			time.Sleep(3 * time.Second)
+			attempts++
+			continue
+		}
+
+		return err
+	}
+}
+
```
stefanotorresi, line 273: as @arbulu89 pointed out, we should instead wait for the consul agent startup and then call `registerConsulService()` without retries.

stefanotorresi (changes requested):

agent/agent.go
rtorrero: It was supposed to be used on `embeddedconsul.go:98` (the `ServiceDeregister` call). In the end, it's working fine without checking if the service actually needs de-register before deregistering it, but I wanted to be sure before removing it. I'll re-test and ensure we can remove it.

agent/agent.go
rtorrero: @stefanotorresi the thing with that approach is that `trento` will attempt to register with the consul instance too quick (before it's actually ready). It will fail and immediately quit. The previous approach was just sleeping for a few seconds, but that is a very ugly workaround.

We could change indeed the order so that the embedded consul instance starts first (it makes more logical sense), but I don't think it will remove the need of the register loop. Initially I was hoping I could somehow signal trento from the embedded consul waitgroup, but doing so requires access to some private attributes / functions from consul that are we can't use.

internal/hosts/hosts.go
```diff
@@ -8,9 +8,9 @@ import (
 	"sort"
 	"strings"
 
-	"github.com/aquasecurity/bench-common/check"
 	consulApi "github.com/hashicorp/consul/api"
 	"github.com/pkg/errors"
+	"github.com/rtorrero/bench-common/check"
```
aleksei-burlakov, line 13: Is it changed on purpose?

stefanotorresi:
This was a great exercise, but we ended up solving the problem with the installer, and Consul is eventually going to go away in our future architecture plans.

PR #78, opened 2021-06-10
WIP: Landing scope page

rtorrero:
This PR is about two things:
 - scope.md document, where we describe trento's general scope and the different scope of each view
 - Changes to the landing page that use this scope.md document to present this content to a new trento user looking to find out more about what it can do.
 
 These two will probably be split in two PRs.
 
 I've set it as WIP because i'm not entirely sure this is what @lee-martin and @stefanotorresi  where looking for. I'm looking for suggestions on how to fill the parts of the documents that have FIXMEs or TODOs and to review the whole content. The landing page changes are just are proposal following UX suggestions.

landing

### DIFF
--- a/docs/scope.md
+++ b/docs/scope.md
@@ -1,0 +1,111 @@
+# Table of contents
+- [About this document](#about-this-document)
+- [Goal](#goal)
+- [Scope](#scope)
+  - [Hosts](#hosts)
+    - [Discovery](#discovery)
+    - [Checks](#checks)
+      - [HA Checker: Scenario SAP HANA Performance Optimized on Azure](#ha-checker-scenario-sap-hana-performance-optimized-on-azure)
+      - [HA Checker: Scenario Pacemaker on Azure](#ha-checker-scenario-pacemaker-on-azure)
+      - [More coming](#more-coming)
+  - [Clusters](#clusters)
+  - [Environments](#environments)
+  - [Landscapes](#landscapes)
+  - [SAP Systems](#sap-systems)
+  - [Agents](#agents)
+
+# About this document
+This document aims to provide information about the current features that `trento`
+tries to cover and its main scope. Further, we provide details on the functionality
+provided by each view.
+
+# Goal
+>Provide simple to use front end for all SAP relevant OS related tasks for SAP 
+>workloads
+
+This means that `trento` focuses on the OS-related tasks without stepping into 
+specific functionality that is already covered by the relevant application layer,
+centered around SAP workloads and simplicity.
+
+# Scope
+The scope of each component of `trento` is as follows:
+
+## Hosts
+In the hosts overview, the user can get a list of all the hosts that are running
+the trento agent. For each host it will be possible to check basic information
+such as the hostname, its IP address, to what cluster it belongs and a list of
+all the tags that have been set by the discovery mechanisms to classify each
+host accordingly. 
+
+When accessing the details of each host, it will be possible to access in-depth
+information about the role each host has in the cluster such as the status of the
+core HDB processes, SAP System ID (SID), etc. Here, the SAP administrator can
+get a list of potential problems and improvements on the configuration of the
+host that the discovery mechanisms have found.
+
+### Discovery
+The discovery mechanisms are the base of the checkers. These implement the code
+in the trento agent to gather information about the nodes, their roles, the
+configuration files that each node uses (depending on their role).
+
+
+### Checks
+Under the checks view it's possible to see a representation on the data collected
+by the agents against predefined values that are provided by our own recommendations
+and our partners such as SAP. The base for the checks are the above described
+[Discovery mechanisms](#discovery)
+
+We currently support 2 checkers:
+
+#### HA Checker: Scenario SAP HANA Performance Optimized on Azure
+This checker uses Azure's recommendations to provide an overview of how well
+the deployed SAP HANA cluster meets their indicators. More details about this
+recommendation can be seen [here](https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/high-availability-guide-suse-pacemaker)
+
+#### HA Checker: Scenario Pacemaker on Azure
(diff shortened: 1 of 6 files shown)


docs/scope.md
```diff
@@ -0,0 +1,113 @@
+# Table of contents
+- [Table of contents](#table-of-contents)
```
arbulu89, line 2: I would remove this `Table of contents` entry. It looks redundant.

docs/scope.md
```diff
@@ -0,0 +1,113 @@
...
+- [About this document](#about-this-document)
```
arbulu89, line 3: I wouldn't put eveything under thing `About this document`. If you want to keep the entry is fine, but I would put the rest of the bullet entries at least in the same level

docs/scope.md
```diff
@@ -0,0 +1,113 @@
...
+    - [Discovery](#discovery)
+  - [Clusters](#clusters)
+  - [Environments](#environments)
+  - [Landscapes](#landscapes)
+  - [SAP Systems](#sap-systems)
+  - [Agents](#agents)
+
+# About this document
+This document aims to provide information about the current features that `trento`
+tries to cover and it's main scope. Further, we provide details on the functionality
```
arbulu89, line 19: Typo: s/it's/its/

docs/scope.md
```diff
@@ -0,0 +1,113 @@
...
+>workloads
+
+This means that `trento` focuses on the OS-related tasks without stepping into 
+specific functionality that is already covered by the relevant application layer,
+centered around SAP workloads and simplicity.
+
+## Hosts
+In the hosts overview, the user can get a list of all the hosts that are running
+the trento agent. For each host it will be possible to check basic information
+such as the hostname, it's IP address, to what cluster it belongs and a list of
```
arbulu89, line 33: s/it's/its/

docs/scope.md
```diff
@@ -0,0 +1,113 @@
...
+When accessing the details of each host, it will be possible to access in-depth
+information about the role each host has in the cluster such as the status of the
+core HDB processes, SAP System ID (SID), etc. Here, the SAP administrator can
+get a list of potential problems and improvements on the configuration of the
+host that the discovery mechanisms have found.
+
+### Checks
+> TODO: Put some additional details here.
+
+We currently support 3 checkers:
```
arbulu89, line 46: 3? I just see 2 here

docs/scope.md
```diff
@@ -0,0 +1,113 @@
...
+
+### Discovery
+> FIXME: Based on what I wrote in [Checks](#checks) I think we should probably
+> have a separate part here: Discovery are the parts of trento which are
+> attempting to detect e.g. the role of a SAP HANA node and the processes that
+> run on each node; while the checkers are additionally checking those parameters
+> against predefined values.
+
+## Clusters
+This view currently allows to check the status of the all the discovered HANA DB
```
arbulu89, line 76: I would remove `HANA DB` here and maybe only mention clusters. We will have SAP app clusters eventually (and maybe more on the road)

docs/scope.md
```diff
@@ -0,0 +1,113 @@
...
+to discover the role of the node and run a set of checks which result in the
+recommendations described above in the [Checkers](#checkers).
+The information that is gathered by them is stored through a distributed KV store
+and is made visible on the `trento` web UI through the `trento` web UI component.
+
+The agents implement the core functionality of trento. They are responsible for
+the discovery of all the clustered components that are required in order to run
+highly available SAP Applications. These agents implement discovery for:
+  - Pacemaker
+  - Corosync
```
arbulu89, line 112: Corosync data gathering is not implemented yet, but well, we can put it here hehe

docs/scope.md
```diff
@@ -0,0 +1,113 @@
...
+  - SAP HANA components
```
arbulu89, line 113: Maybe better `SAP components`

web/home.go
```diff
@@ -13,8 +13,9 @@ type HomeData struct {
 
 func HomeHandler(c *gin.Context) {
 	data := HomeData{
-		Title:     defaultLayoutData.Title,
-		Paragraph: "This is the home page",
+		Title: defaultLayoutData.Title,
```
arbulu89, line 16: As we already put all the data in the template, I would vote to pass the `title` and `paragraph` texts there too, without having them here in the controller code

web/layout.go
```diff
@@ -37,7 +37,7 @@ type SubmenuItem struct {
 }
 
 var defaultLayoutData = LayoutData{
-	Title:     "Trento Console",
+	Title:     "Trento Console for SAP Applications",
```
arbulu89, line 40: I would leave it as `Trento console`, but I leave this decision to @lee-martin 
Otherwise, we will need to widen back again the sidebar

web/templates/home.html.tmpl
```diff
@@ -4,4 +4,111 @@
...
+            <h5 class='card-header'>Agents</h5>
+            <div class='card-body'>
+                <p class='card-text'>
+                    The agents implement the core functionality of trento. They
+                    are responsible for the discovery of all the clustered components
+                    that are required to run SAP Applications in a highly
+                    available way. These agents implement discovery for:
+                     - Pacemaker
+                     - Corosync
+                     - HANA
```
arbulu89, line 19: SAP component in general, not only HANA

arbulu89 (changes requested):
Thanks @rtorrero 
The page looks good in general (referring to styles and html format).
I have added many comments about the text itself.

3 major details:
- As rule of thumb we shouldn't use the `we` pronoum in the text, it looks strange. We should talk about the project.
- The description of the boxes shows the current state, which is just something I created myself. I guess this is fine by now, but we wouldn't need to specify that much (like in the hosts, saying that it shows the SAP processes). They should be more generic in my opinion
- The document always talks about HANA. I would try to talk about SAP components in general, as trento is not HANA specific, is SAP specific.

PR #82, opened 2021-06-11
Landing page update with scope.md references

rtorrero:
This is a re-submission of https://github.com/trento-project/trento/pull/78

(see https://github.com/trento-project/trento/pull/78#issuecomment-859355825)

### DIFF
--- a/web/home.go
+++ b/web/home.go
@@ -7,14 +7,12 @@ import (
 )
 
 type HomeData struct {
-	Title     string
-	Paragraph string
+	Title string
 }
 
 func HomeHandler(c *gin.Context) {
 	data := HomeData{
-		Title:     defaultLayoutData.Title,
-		Paragraph: "This is the home page",
+		Title: defaultLayoutData.Title,
 	}
 	c.HTML(http.StatusOK, "home.html.tmpl", data)
 }
--- a/web/templates/home.html.tmpl
+++ b/web/templates/home.html.tmpl
@@ -91,9 +91,9 @@
     </div>
     <div class='mt-4'>
         For more information and getting started, plese refer to the 
-        <a target='blank' href='https://github.com/trento-project/trento/blob/9f1c29951ac49669382c9812f0cd0fe91c1f3aa2/docs/scope.md'>scope, </a>
-        <a target='blank' href='https://github.com/trento-project/trento/blob/9f1c29951ac49669382c9812f0cd0fe91c1f3aa2/README.md'>readme</a> or
-        <a target='blank' href='https://github.com/trento-project/trento/blob/9f1c29951ac49669382c9812f0cd0fe91c1f3aa2/docs/trento-architecture.md'>architecture</a>
+        <a target='blank' href='https://github.com/trento-project/trento/blob/518135b7152a384b59714f59d3ae72722a03b458/docs/scope.md'>scope, </a>
+        <a target='blank' href='https://github.com/trento-project/trento/blob/518135b7152a384b59714f59d3ae72722a03b458/README.md'>readme</a> or
+        <a target='blank' href='https://github.com/trento-project/trento/blob/518135b7152a384b59714f59d3ae72722a03b458/docs/trento-architecture.md'>architecture</a>
         documents.
     </div>
 </section>    
--- a/web/templates/home.html.tmpl
+++ b/web/templates/home.html.tmpl
@@ -1,7 +1,100 @@
 {{- /*gotype: github.com/trento-project/trento/web.HomeData*/ -}}
 {{ define "content" }}
 <div class="col">
-    <h1>{{ .Title }}</h1>
-    <p>{{ .Paragraph }}</p>
+    <h1 class='display-4 lead'>{{ .Title }}</h1>
+    <p class='subheadline'>An open cloud-native web console improving the
+				life of SAP Applications administrators</p>
 </div>
+<hr/>
+<section>
+    <div class='mb-5'>
+        This project is a reboot of the "SUSE Console for SAP Applications", also known 
+        as the <a target='blank' href='https://github.com/SUSE/blue-horizon-for-sap'>Blue Horizon for SAP</a>
+        prototype, which is focused on automated infrastructure deployment and provisioning for SAP Applications
+    </div>
+    <div class='card-deck'>
+        <div class='card'>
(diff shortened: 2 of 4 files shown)


web/layout.go
```diff
@@ -37,7 +37,7 @@ type SubmenuItem struct {
 }
 
 var defaultLayoutData = LayoutData{
-	Title:     "Trento Console",
+	Title:     "Trento Console for SAP Applications",
```
dirkmueller, line 40: why this change? it is already long enough imho ;)

web/templates/home.html.tmpl
```diff
@@ -4,4 +4,113 @@
...
+<div class='container'>
+    <div class='card-deck'>
+        <div class='card'>
+            <h5 class='card-header'>Agents</h5>
+            <div class='card-body'>
+                <p class='card-text'>
+                    Though not specifically a view/section, as it is the core component of Trento,
+                    it is important to understand what they are.
+
+                    The agents are the processes that run in each of the nodes that conform
```
dirkmueller, line 16: this (and the others) duplication of text with the docs/ file is a bit unfortunate. how difficult would it be to pull something like https://github.com/gomarkdown/markdown to embed the documentation ?

stefanotorresi: We thought about this, and we already have `gomarkdown/markdown` integrated and available via the `markdown` template function (see #62).
That said, the homepage is likely to change very soon anyway, so we agreed that it would make a simple static HTML page more complicated than it really should.

web/layout.go
rtorrero: I thought I saw it in some of the UI mocks but I can't seem to find it anymore. I'll change it!

web/templates/home.html.tmpl
rtorrero: @dirkmueller I already explored this and had a rudimentary version working. But if we want to keep markdown for the scope document it gets quite ugly as one needs to iterate over markdown (or what I did, convert to html and navigate the <h3> tags to render each one of them in a separate card). `JSON` or `yaml` would be better for this but at the expense of losing readability in the document when accessed separately; but IMO this would be overcomplicating this for something that is just meant to be temporal anyways.

stefanotorresi (changes requested):
Hey team, tlease see https://github.com/trento-project/trento/discussions/86 before merging.

Beside that, I think the sections should be ordered according to the sidebar (also see #84 about that), which would put the "Agents" section last, which also reflects the source documentation structure.

Also, I'm having second thoughts about the Call To Action link going to the docs: I would expect the main action of the user, once they open the homepage, to be going to any of the main navigation items of the application. 

We might be misinterpreting Sorin's original indications: the Call To Action of each section should go to the relevant item, while we should have one single link to the source documentation at the bottom of the page. This also means that we should produce a meaningful abstract text instead of letting Bootstrap's automatic ellipsis truncate the text with `...`.

web/templates/home.html.tmpl
```diff
@@ -4,4 +4,113 @@
     <h1>{{ .Title }}</h1>
     <p>{{ .Paragraph }}</p>
 </div>
+<div class='container'>
```
en3sis, line 7: The `container` class will add some max-width and extra margins and it only should be used once.

We usually divide the content of a section in `<section>  </section>` that has no extra style and it won't push the content related to the other elements (titles or descriptions)

en3sis:
@stefanotorresi the way it was implemented would be the correct one based on the very early discussion that was about having a quick description of each functionality where we would talk about the current status (brief explanation) and a link to the documentation for more details.
If the scope changed we could rework it, definitely. 

> This also means that we should produce a meaningful abstract text instead of letting Bootstrap's automatic ellipsis truncate the text with   

Agree :P 

> single link to the source documentation at the bottom of the page

Would be also useful

rtorrero:
I've done the requested changes, some of them have a little interpretation on my side; let me know if something else needs change.
@en3sis about using `<section>`, the thing is that the margins don't exactly match what you did in your mock. I think it's close enough but be sure to check it. Also, the fonts are not exactly the same, let me know what bootstrap classes I need to use if you'd like to have the same. Thanks!
landing2

en3sis:
@rtorrero I think the only thing it's missing is that the  `.subheadline` class for the subheadline.

For the fonts please refer to https://productbrand.suse.com/typography but basically, all headers should use `Poppins` while `p` tags should use `Lato`. (it should be part of the DS CSS already so no extra work should be needed) 

Everything else LGTM!

rtorrero:
@en3sis Thanks! I guess the differences I saw on my local tests vs your screenshot mockup are due to OS?

@stefanotorresi should I squash or do you see anything else that needs change? I used the mockup text as I find them a good starting point.

web/home.go
```diff
@@ -13,8 +13,9 @@ type HomeData struct {
 
 func HomeHandler(c *gin.Context) {
 	data := HomeData{
-		Title:     defaultLayoutData.Title,
-		Paragraph: "This is the home page",
+		Title: defaultLayoutData.Title,
+		Paragraph: `An open cloud-native web console improving on the
```
arbulu89, line 17: I find this `on` from `...console improving on the...` out of place there

web/home.go
```diff
@@ -13,8 +13,9 @@ type HomeData struct {
 
 func HomeHandler(c *gin.Context) {
 	data := HomeData{
-		Title:     defaultLayoutData.Title,
-		Paragraph: "This is the home page",
+		Title: defaultLayoutData.Title,
```
arbulu89, line 16: As commented, I think the `Title` and `Paragraph` should go in the html template itself, together with everything else, but well, who cares, we will touch this many time on the road hehe

arbulu89 (approved):
It looks great for me. Added some really non meaningful comments hehe

PD: I guess the `SAP systems` text in the sidebar is changed in other PR, right?

web/home.go
rtorrero: I think both are right, but since your suggestion has two characters less... I agree :-p

web/home.go
rtorrero: The title is also rendered in the menu so I'll leave it, but I see no point in keeping the paragraph there. I'll put it in the template. Thanks!

stefanotorresi (approved):

stefanotorresi:
gj 🍺

PR #87, opened 2021-06-11
2-nodes cluster must either have disk-based sbd or qdevice

aleksei-burlakov:


examples/azure-rules.yaml
```diff
@@ -437,6 +437,40 @@ groups:
...
+            echo passed
+            exit 0
+          fi
+          echo failed
+        tests:
+          test_items:
+            - flag: passed
+        remediation: |
+          ## Remediation
+          HA cluster with 2 nodes must wither have a disk-based SBD of a Qdevice.
```
dirkmueller, line 469:
```suggestion
          HA cluster with 2 nodes must either have a disk-based SBD of a Qdevice.
```

aleksei-burlakov: Thank you, fixed.

dirkmueller (approved):

PR #90, opened 2021-06-11
Add project logo to header

stefanotorresi:
Screenshot from 2021-06-11 18-27-45

Also removes the rendering of the header submenu if empty (which it is, atm), because it causes some problems on the smallest viewport size.


dirkmueller:
I only see trento-agent.png used - what are all the other icons for that you added?

en3sis:
According to[ our guidelines]( https://productbrand.suse.com/product-labels) SUSE products should not have any icons/illustration

> As we do not create product brands, we don’t apply logos for commercial products. We label them with their names, written in our corporate font Poppins and without any additional picture mark. As with any rule, there is one exception: SUSE Rancher

en3sis:
Yes, the logo can be there! 

BTW maybe this could be a good moment to also add it as the websites `favicon` ?

stefanotorresi:
> I only see trento-agent.png used - what are all the other icons for that you added?

hmm, I added those for potential future usage.

dirkmueller (approved):

PR #91, opened 2021-06-14
Check the UCAST is used by corosync with at least 2 com-n rings

aleksei-burlakov:
Ref: CFSA-133, CFSA-134

### DIFF
--- a/examples/azure-rules.yaml
+++ b/examples/azure-rules.yaml
@@ -214,6 +214,68 @@ groups:
           ## References
           - https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/high-availability-guide-suse-pacemaker
         scored: true
+      - id: 1.1.9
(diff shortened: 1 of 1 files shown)


aleksei-burlakov:
As agreed in the **CFSA-134**, it doesn't fail but warns when there is only 1 network interface.

stefanotorresi (approved):

examples/azure-rules.yaml
```diff
@@ -214,6 +214,68 @@ groups:
           ## References
           - https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/high-availability-guide-suse-pacemaker
         scored: true
+      - id: 1.1.9
```
diegoakechi, line 217: @aleksei-burlakov Isn't it duplicated of 1.1.6?

PR #92, opened 2021-06-14
Embed rulesets in the application

arbulu89:
Embed the current rulesets in the application.

- Move the rulesets from `examples` to `rulesets`
- Add an integer to the ruleset file names. This is used to provide order when the rulesets are executed and showed
- Copy the rulesets to the `ruleset` code folder during the `make` process to embed them. **The embed usage doesn't allow relative folders with `../` entry
- Create a new `rulesets` package to manage the embedded and user provided rulesets. The user can provide rulesets as it was doing before, as an argument at the end of the agent operation. The logic is moved to `internal`, as this will be used in the future in the `web` part too, to load the used rulesets.
- The `embed` package is used to embed the files in the app: https://golang.org/pkg/embed/

**PD: This PR doesn't implement the option to choose the rulesets. All of the loaded rulesets are applied always. This feature will come next.**


internal/ruleset/ruleset.go
```diff
@@ -0,0 +1,104 @@
...
+	for _, path := range r.UserFiles {
+		datum, err := ioutil.ReadFile(path)
+		if err != nil {
+			return nil, errors.Wrap(err, "could not read user rulesets file")
+		}
+		data = append(data, datum)
+	}
+
+	first_yaml_file := string(data[0])
+	// Appent all next files to the first
```
dirkmueller, line 98:
```suggestion
	// Append all next files to the first
```

internal/ruleset/ruleset.go
```diff
@@ -0,0 +1,104 @@
...
+		return r, err
+	}
+
+	r.EmbeddedFiles = embeddedFiles
+	r.UserFiles = userFiles
+
+	return r, nil
+}
+
+func loadEmbeddedFiles() ([]string, error) {
```
dirkmueller, line 37: comment missing, and possibly misnamed. it isn't loading the files, its just enumerating (listing) them

internal/ruleset/ruleset.go
```diff
@@ -0,0 +1,104 @@
...
+	pattern := "\ngroups:"
+	index := strings.Index(yaml_text, pattern)
+	if index == -1 {
+		log.Fatal("the yaml file doesn't contain the 'id:' field and is possibly broken")
+	}
+	index += len(pattern)
+	return yaml_text[index:]
+}
+
+func (r *RuleSet) GetRulesetFiles() []string {
```
dirkmueller, line 74: unused, what is it good for?

dirkmueller:
It is an immediate improvement, but going forward goes further into the wrong direction - rules need to be fetched from a central location and kept in sync to avoid having to lockstep-update the agent all the time on all the nodes?

arbulu89:
> It is an immediate improvement, but going forward goes further into the wrong direction - rules need to be fetched from a central location and kept in sync to avoid having to lockstep-update the agent all the time on all the nodes?

I'm a bit surprised with this comment. The task implemented in this PR just embeds the rulesets into the binary/app, which was the requirement. All of the nodes will have the same embedded rulesets.

About the user provided rulesets. Yes, you are right, it should be centralized, and this will come in the future (and I'm not sure how we will do it, as it is a really tricky thing).

**PD:** Anyway, I have created much more changes on top of this to enable the option to choose the specific rulesets to execute. Many parts of this implementation will go away. I just created the PR to have a small piece of changes, but I don't know if its a good idea, as it will take really long to be merged most probably...
Maybe, I will close this PR and create the other one, even though it will much larger than this one

arbulu89:
@dirkmueller I'm closing this PR in order to open a new one with more advanced features about this topic

internal/ruleset/ruleset.go
```diff
@@ -0,0 +1,104 @@
...
+	"embed"
+	"io/ioutil"
+	"log"
+	"path"
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+//go:embed rulesets
```
rtorrero, line 13: Should we use rulesets/*.yaml here maybe to limit it a bit more?

dirkmueller:
> I'm a bit surprised with this comment. The task implemented in this PR just embeds the rulesets into the binary/app, which was the requirement. All of the nodes will have the same embedded rulesets.


in a larger deployment you have 1-3 trentos in "server mode" and 10+ in "agent mode". it might be challenging for some users to keep all of them in lockstep on the same version at any  point in time. so we'll have to allow for them to deviate somewhat from each other. 

one way to do so would be to fetch anything important the agent needs to be "current" from the server, which is why we used the consul kv store. now obviously the rules are an unsolved problem in that regard, they would have to be fetched from the server as well by the agent repeatedly, or the agent would have to self-update from the binary on the server (at which the rules could be embedded again).

PR #96, opened 2021-06-14
Remove repeating UCAST check 1.1.9

aleksei-burlakov:
Checking that the corosync uses UCAST was already implemented in 1.1.6,
so there is no need in 1.1.9 that does the same. The runtime version however remains.

### DIFF
--- a/examples/generic-azure.yaml
+++ b/examples/generic-azure.yaml
@@ -185,7 +185,7 @@ groups:
           It will be difficult to change the corosync MCAST transport to UCAST
           on an existing cluster. Besides adding the 'transport: udpu' parameter one
           would have to configure each node separatelly. The easiest way to
-          make the cluster UCAST it to bootstrap it. When creatin the cluster
+          make the cluster UCAST it to bootstrap it. When creating the cluster
           you would have to add the key -u, for example
           $ crm cluster init -u ...
 
--- a/examples/generic-azure.yaml
+++ b/examples/generic-azure.yaml
@@ -164,7 +164,30 @@ groups:
             - flag: "udpu"
         remediation: |
           ## Remediation
-          Adjust the corosync transport parameter as recommended on the Azure best practices.
+          It will be difficult to change the corosync MCAST transport to UCAST
+          on an existing cluster. Besides adding the 'transport: udpu' parameter one
+          would have to configure each node separatelly. The easiest way to
+          make the cluster UCAST it to bootstrap it. When creatin the cluster
+          you would have to add the key -u, for example
+          $ crm cluster init -u ...
+
+          ## References
+          - https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/high-availability-guide-suse-pacemaker
+        scored: true
+      - id: 1.1.6.runtime
+        description: "Corosync transport parameter is set to udpu (runtime)"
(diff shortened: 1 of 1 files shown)


examples/generic-azure.yaml
```diff
@@ -166,6 +166,19 @@ groups:
...
+        scored: true
+      - id: 1.1.6.runtime
+        description: "Corosync transport parameter is set to udpu (runtime)"
+        audit: 'corosync-cmapctl | grep "totem.transport (str) = " | sed "s/.*= //"'
+        tests:
+          test_items:
+            - flag: "udpu"
+        remediation: |
+          ## Remediation
+          Adjust the corosync transport parameter as recommended on the Azure best practices.
```
arbulu89, line 180: I think this kind of remediation are not really helpful.
We should explain here what the user should do, and later on reference them to the main documentation page.

At least, personally, and even knowing our cluster, I cannot really get any meaning from this message.

Shouldn't we have something like `Adjust corosync transport parameter to udpu. This is configured in /etc/corosync/corosync, in the totem.transport field`. Or something similar.

arbulu89 (commented):
Hi @aleksei-burlakov ,

The check looks good. In the other hand I don't really like how the remediation text is created. I think we would need to have something more specific.
Anyway, maybe this is something already agreed with @stefanotorresi @lee-martin

examples/generic-azure.yaml
aleksei-burlakov: > I think this kind of remediation are not really helpful.
> We should explain here what the user should do, and later on reference them to the main documentation page.

I totally agree with this. 

> Adjust corosync transport parameter to udpu. This is configured in /etc/corosync/corosync, in the totem.transport field

It's still not f a big use. I will amend it and explain in detail how to fix it.

examples/generic-azure.yaml
```diff
@@ -164,7 +164,30 @@ groups:
             - flag: "udpu"
         remediation: |
           ## Remediation
-          Adjust the corosync transport parameter as recommended on the Azure best practices.
+          It will be difficult to change the corosync MCAST transport to UCAST
+          on an existing cluster. Besides adding the 'transport: udpu' parameter one
+          would have to configure each node separatelly. The easiest way to
+          make the cluster UCAST it to bootstrap it. When creatin the cluster
```
stefanotorresi, line 170:
```suggestion
          make the cluster UCAST it to bootstrap it. When creating the cluster
```

examples/generic-azure.yaml
```diff
@@ -164,7 +164,30 @@ groups:
...
+        audit: 'corosync-cmapctl | grep "totem.transport (str) = " | sed "s/.*= //"'
+        tests:
+          test_items:
+            - flag: "udpu"
+        remediation: |
+          ## Remediation
+          It will be difficult to change the corosync MCAST transport to UCAST
+          on an existing cluster. Besides adding the 'transport: udpu' parameter one
+          would have to configure each node separatelly. The easiest way to
+          make the cluster UCAST it to bootstrap it. When creatin the cluster
```
stefanotorresi, line 188:
```suggestion
          make the cluster UCAST it to bootstrap it. When creating the cluster
```

stefanotorresi (changes requested):

stefanotorresi (approved):

stefanotorresi:
gj 🍻

PR #100, opened 2021-06-16
Check that the HANA and HANA SPS versions are compatible

aleksei-burlakov:
For SAP HANA 1.0 you need version SPS10 rev3, SPS11 or newer. SAP HANA 2.0 is supported with all known SPS versions.

I'm not 100% sure if in the lines 434 and 435, @arbulu89 need your help here.

### DIFF
--- a/examples/hana-scale-up-perf-optimized-azure.yaml
+++ b/examples/hana-scale-up-perf-optimized-azure.yaml
@@ -390,7 +390,7 @@ groups:
           Update sbd packages
         scored: true
       - id: 2.2.6
-        description: "Check the SAPHana version is at least"
+        description: "Check the SAPHanaSR version is at least"
         audit: 'echo --version=$(rpm -q --qf "%{VERSION}\n" SAPHanaSR | cut -f2 -d.)'
         tests:
           test_items:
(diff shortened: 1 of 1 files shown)


arbulu89 (commented):
@aleksei-burlakov To be honest, I wouldn't have any check based on HANA 1.0.
I don't even konw if we should support HANA 1.0 installations. @lee-martin ?

Besides that, the script looks good. Most probably @angelabriel or @fmherschel know more about this

angelabriel:
If I remember right SAP HANA 1.0SPS12, the latest and last SPS for HANA1.0, will reach end of maintenance in June 2021, which means 'NOW' :-)

lee-martin:
@aleksei-burlakov  This story is about implementing the check from the SUSE HA setup guide "    SAP HANA 1.0 is supported since SPS09 (095) for all mentioned setups. SAP HANA 2.0 is supported with all known SPS versions. " ( https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-sr-guide-PerfOpt-15/ )

We are not intending to make any HANA/SLES support statements with this check, just make sure those minimal versions are in place with the Resource Agents expect. i.e. it does not matter what is or is not currently supported by SAP for this check. Its just a minimal version. HANA 1.0 is supported on SLE15, so makes sense to keep.

Wheres does the quote "For SAP HANA 1.0 you need version SPS10 rev3, SPS11 or newer. " come from?

angelabriel:
> Wheres does the quote "For SAP HANA 1.0 you need version SPS10 rev3, SPS11 or newer. " come from?

It's from the same document you reffered to  - page 13 in the pfd. But it's related to a special case (multi-tenancy (MDC) databases  - if you want to stop tenants during production and you want the cluster to be able to take over.)

So it depends on which use case you will check against, if you need HANA1.0 SPS09 or HANA1.0 SPS10.

dirkmueller (approved):
LGTM

aleksei-burlakov:
The Merge-button was green, I couldn't resist.

PR #102, opened 2021-06-16
Fix empty tables with a user-friendly message in case of no records

### LINKED ISSUE #83: Fix empty tables with a user-friendly message in case of no records.
Instead of showing empty tables with no rows, we should show a "There are currently no records to be shown" message below the table header.
This means adding a conditional beside every single a range loop in any html template involving tables.

fabriziosestito:
Resolves #83

Adds a chore commit that formats the html templates and resolves some problems with a few unclosed tags etc.

It avoids to add extra logic to the templates where it is known that the ranged resource can't be empty.

### DIFF
--- a/web/templates/blocks/empty_table_body.html.tmpl
+++ b/web/templates/blocks/empty_table_body.html.tmpl
@@ -1,0 +1,5 @@
+{{ define "empty_table_body" }}
+    <tr>
+        <td colspan="{{ . }}">There are currently no records to be shown</td>
+    </tr>
+{{ end }}
--- a/web/templates/blocks/footer.html.tmpl
+++ b/web/templates/blocks/footer.html.tmpl
@@ -1,8 +1,8 @@
 {{ define "footer" }}
-<footer class="footer-content js-footer-content">
-  <ul class="footer-list">
-    <li class="footer-list-item">{{ template "license" }}</li>
-    <li class="footer-list-item">{{ .Copyright }}</li>
-  </ul>
-</footer>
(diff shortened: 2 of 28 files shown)


arbulu89 (changes requested):
@fabriziosestito Do you think we could use a block template to store this duplicated entry?

fabriziosestito:
> @fabriziosestito Do you think we could use a block template to store this duplicated entry?
> I guess we could pass the columns number as parameter. This way we would have everything in just one file.
> Let me know if this could improve somehow the duplications
It might help but it will also result in another nested block and at the end it will not avoid the duplication (you still need to import it around the html template files).
As long as it does not grow more than one string (probably localized in the future) I will keep it as is.
If we find out that we need something more than a message (e.g. some proper html with icons and style) then I will be up for moving the content in a separate block.

arbulu89 (approved):
@fabriziosestito It looks nice to me!
 PD: I didn't know that the `range` method has `else` option (it reminds me somehow the same option from python. Do you know if this is a feature available in raw go code, or is it available only in the templates

fabriziosestito:
> @fabriziosestito It looks nice to me!
> PD: I didn't know that the `range` method has `else` option (it reminds me somehow the same option from python. Do you know if this is a feature available in raw go code, or is it available only in the templates


 
 I think this is templates-only, I can;t find any reference on it on TGPL.
 Yes, it's similar to the `empty` in django templates!

dirkmueller (approved):

PR #105, opened 2021-06-17
Check NTP time synchronization is configured

aleksei-burlakov:


diegoakechi:
@aleksei-burlakov  Time synchronization is a requirement of any pacemaker cluster, so I would say it need to be included on the generic checks.

stefanotorresi (changes requested):
Yeah, I confirm that: this needs to go in the `generic` rule set.

aleksei-burlakov:
Done. The check is now generic.

examples/generic-azure.yaml
```diff
@@ -637,4 +637,41 @@ groups:
 
           ## References
           - Cluster installation in https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/high-availability-guide-suse-pacemaker
-        scored: false
\ No newline at end of file
+        scored: false
+      - id: 1.5.2
+        description: "npt service"
```
dirkmueller, line 642:
```suggestion
        description: "ntpd service running"
```

examples/generic-azure.yaml
```diff
@@ -637,4 +637,41 @@ groups:
...
+        audit: |
+          if ! sudo systemctl status ntpd; then echo "result=failed"; fi
+          if sudo ntpq -p | grep "No association ID's returned"; then echo "result=failed"; fi
+          if [[ $(sudo ntpq -p | wc -l) -lt 3 ]]; then echo "result=failed"; fi
```
dirkmueller, line 646: why would it matter if its 3 or more? the important part is one of those is accepted, e.g. there is `grep  '^\*` matching it

dirkmueller: ah, it just checks whether there is at least one result returned.

aleksei-burlakov: Yes, correct, the output is something like
```
cloudadmin@vmhana01:~> sudo ntpq -p
     remote           refid      st t when poll reach   delay   offset  jitter
==============================================================================
 20.101.61.75    .INIT.          16 u  597 1024    0    0.000    0.000   0.000

```

dirkmueller: yeah, this output is actually a fail (stratum is 16, sync is missing). your check makes it a pass

aleksei-burlakov: > if kvm virtualization is used, assume kvm-clock (e.g. real time clock is passed in from hypervisor host and not emulated within the virtual machine) -> no ntp check, or verify that ntp is not running, because it shouldn't
> in other cases, check that the clock has sync (there is one chosen peer, as indicated by an '*' in the first line.

It's amended. If it's a kvm-guest it checks the kvm-clocks, otherwise, it checks ntp with stratum != 16.

stefanotorresi: @dirkmueller we are already assuming that we are on Azure cloud, which kinda excludes that we're running on a kvm guest, doesn't it? So I would safely assume that we can rely on just ntp.

gereonvey: Pretty much every SAP Note related to setting up a system will tell you that it's strongly recommended to set up time sync (i.e. NTP). We also do that in our SLES15 SAP Note. 
I didn't find any indication to deviate from that recommendation for KVM VMs, any other virtualization or cloud. 

When you check the NTP configuration/status, please also make sure it's configured with "slew" option (i.e. only minimal jumps in time when adjusting). Bad things can happen if it's configured with "step".

lee-martin: @gereonvey thanks for your feedback!  this was my recollection as well. Ah, yes, the SAP Note is "2578899 - SUSE Linux Enterprise Server 15: Installation Note"

@aleksei-burlakov  can you please remove the KVM exception and also add reference to SAP Note "2578899 - SUSE Linux Enterprise Server 15: Installation Note" in addition to our SUSE HA document.
Can you also check for "slew" as mentioned by Gereon?  (If this is too much for one check, we could consider splitting) @stefanotorresi & @dirkmueller - are you two ok with this?

stefanotorresi: Yes, it's fine: after some googling it seems that running NTP is always a good idea, on any hypervisor, so I wouldn't treat KVM any differently.
@aleksei-burlakov the url of the sap note to refer is: https://launchpad.support.sap.com/#/notes/2578899.

aleksei-burlakov: Thanks for reviewing, I have

- removed the kvm check
- added the slew check
- changed the reference

dirkmueller: https://documentation.suse.com/sles/15-SP1/html/SLES-all/sec-kvm-managing-clock.html#sec-kvm-managing-clock-kvm recommends to run ntp even with kvm_clock. so that part seems fine.

examples/generic-azure.yaml
```diff
@@ -672,4 +672,39 @@ groups:
...
           ## References
           - section 9.1.2 https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-sr-guide-PerfOpt-15/
+      - id: 1.5.3
+        description: "ntp clock synchronization is working and using the slew mode"
+        audit: |
+          # The third field of ntpq -p is the stratum and if it's 16
+          # the sync is disabled. Make sure it's not 16
+          if ! sudo ntpq -pn | grep "No association ID's returned" \
+          && sudo ntpq -pn | awk '{print $3}' | grep -e "[0-9]" | grep -v 16; then
+            if ps aux | grep "/usr/sbin/ntpd" | grep "\-x"; then echo passed; fi
```
dirkmueller, line 682: this is quite a weak way of checking for this, as you're listing all processes (so including the grep process). the easiest fix is to ensure that grep doesn't find itself by making a regexp that doesn't match itself, e.g.

`grep "/usr/sbin/[n]tpd`

but you could also do a `grep -q --- -x /proc/$(</run/ntpd.pid)/cmdline` which would do pretty much the same thing..

examples/generic-azure.yaml
```diff
@@ -672,4 +672,39 @@ groups:
 
           ## References
           - section 9.1.2 https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-sr-guide-PerfOpt-15/
+      - id: 1.5.3
+        description: "ntp clock synchronization is working and using the slew mode"
+        audit: |
+          # The third field of ntpq -p is the stratum and if it's 16
+          # the sync is disabled. Make sure it's not 16
+          if ! sudo ntpq -pn | grep "No association ID's returned" \
+          && sudo ntpq -pn | awk '{print $3}' | grep -e "[0-9]" | grep -v 16; then
```
dirkmueller, line 681: stratum of 16 is one indicator of why a ntp source isn't used, there can be others (Ilike too high jitter). I really suggest to check for sync flag ('*') instead, like previously mentioned.

aleksei-burlakov: > stratum of 16 is one indicator of why a ntp source isn't used, there can be others (Ilike too high jitter). I really suggest to check for sync flag ('*') instead, like previously mentioned.

@dirkmueller thank you for the feedback. I have strengthened the grep expression and grep only the ntp servers with the '* flag'

examples/generic-azure.yaml
rtorrero: Marking this as resolved as it was already addressed

rtorrero:
I'm dismissing the pending change request as it was already addressed. Hope you don't mind me stepping in @aleksei-burlakov @stefanotorresi but this seems like everything is already looking good.

I'm giving my green light, feel free to merge if you don't see anything else that requires changes @aleksei-burlakov 

Good work & thanks!

rtorrero (approved):
LGTM

PR #106, opened 2021-06-17
Get system replication and landscape data for HANA

arbulu89:
Store more information for the SAP systems (only available for HANA).

**The changes in `sbd` are coming because I have moved and changed slightly the `FindMatches` function to the `utils.go` file**

- The `DEFAULT.PFL` profile content (this is always the active profile content of the whole SAP system)
- The output from `systemReplicationStatus.py` and `landscapeHostConfiguration.py` commands.

**These 2 commands are provided as stable APIs from SAP. They are used in our Resource agents, and there is an agreement that the output format and content type won't be changed**

As a comment, some of the SAP information is already stored as hierarchy, that is applied using `/` to describe them. This means that these parameters are stored as maps in the KV, which is good to consume.

As a contra, this means, that `SAPSystem` created with the `NewSAPSystem` and loaded with KV `Load` method might have different structure for these new entries. Example:

1. `NewSAPSystem` -> `"dbs/hdb/dbname": "PRD"` (one key/value)
2. `Load` -> `dbs[hdb][dbname]` : `PRD` as a map

I didn't find any issue with this so far.

Example:
Look that the `site` is a map
image


arbulu89:
@stefanotorresi @fabriziosestito @rtorrero @aleksei-burlakov C'mon team, I need a review here

fabriziosestito (dismissed):

stefanotorresi (dismissed):
LGTM, feel free to merge right away after rebase.

Issue #57, opened 2021-05-13
Implement levelled logging


stefanotorresi:
with https://github.com/sirupsen/logrus

PR #107, opened 2021-06-17
Change: standard logging --> github.com/sirupsen/logrus

aleksei-burlakov:
This PR is addressed to #57


agent/consultemplate.go
```diff
@@ -1,13 +1,12 @@
...
 
 import (
-	"log"
 	"path"
 
-	"github.com/pkg/errors"
-
 	consultemplateconfig "github.com/hashicorp/consul-template/config"
 	"github.com/hashicorp/consul-template/manager"
+	"github.com/pkg/errors"
```
aleksei-burlakov, line 8: The VS-Code has moved the errors package down the list, I think alphabetically.

dirkmueller:
@aleksei-burlakov update to go.sum is missing (see CI failure)

arbulu89 (commented):
@aleksei-burlakov I would take advantage to work more on this task. 
We would need to setup a proper logging system, which send the logs to a configured log file with used log levels etc.
I would expect to have some logfile in `/var/log/trento/agent.log` (be default) or so, and be able to configure the logging level.

I think it would make sense to work it in this PR

stefanotorresi:
> @aleksei-burlakov I would take advantage to work more on this task.
> We would need to setup a proper logging system, which send the logs to a configured log file with used log levels etc.
> I would expect to have some logfile in `/var/log/trento/agent.log` (be default) or so, and be able to configure the logging level.
> 
> I think it would make sense to work it in this PR

Hmm, hey Xabi, I disagree here: as we previously discussed, I don't think we need to implement logging on files. Cloud-native applications can just log to stdout, and rely on other external mechanism to gather the logs into something else.

stefanotorresi (changes requested):
Hey Aleksei, can we have a new cli flag to change the log level?

aleksei-burlakov:
> Hey Aleksei, can we have a new cli flag to change the log level?

Sure. It will come in a new comit.

arbulu89:
> > @aleksei-burlakov I would take advantage to work more on this task.
> > We would need to setup a proper logging system, which send the logs to a configured log file with used log levels etc.
> > I would expect to have some logfile in `/var/log/trento/agent.log` (be default) or so, and be able to configure the logging level.
> > I think it would make sense to work it in this PR
> 
> Hmm, hey Xabi, I disagree here: as we previously discussed, I don't think we need to implement logging on files. Cloud-native applications can just log to stdout, and rely on other external mechanism to gather the logs into something else.

I commented this because most of the well known logging libraries in all languages don't implement this in the code, it comes just in a configuration file. I just commented to have the option to have this configuration, that again, comes from a configuration file, so it would be agnostic.
If this is not the case in `golang/logrus`, omit my comment.

stefanotorresi:
> PD: I'm sure we will finish implementing the log to file mechanism. This is how many sys admins like to work, and even though you say `cloud-native`, the trento agent will be just yet another package/daemon running in a machine ;)

Yes, of course we'll do that, but my point is that we don't need this _within_ the Trento application itself, because there are other OS facilities that we can use for it, e.g. the systemd journal. :)

aleksei-burlakov:
> > PD: I'm sure we will finish implementing the log to file mechanism. This is how many sys admins like to work, and even though you say `cloud-native`, the trento agent will be just yet another package/daemon running in a machine ;)
> 
> Yes, of course we'll do that, but my point is that we don't need this _within_ the Trento application itself, because there are other OS facilities that we can use for it, e.g. the systemd journal. :)

aleksei-burlakov:
I have accidentally closed the PR. Reopening it.

aleksei-burlakov:
> Hey Aleksei, can we have a new cli flag to change the log level?

Done. In the new commit.

stefanotorresi (approved):

cmd/root.go
```diff
@@ -32,6 +34,7 @@ func Execute() {
 func init() {
 	cobra.OnInitialize(initConfig)
 	rootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.trento.yaml)")
+	rootCmd.PersistentFlags().StringVar(&logLevel, "log-level", "", "then minimum severity (error, warn, info, debug) of logs to output")
```
arbulu89, line 37: I think you could put the default value as the 3rd parameter here, instead of a an empty value, which you use later to get the default value

cmd/root.go
```diff
@@ -51,6 +54,11 @@ func initConfig() {
 		viper.SetConfigName(".trento")
 	}
 
+	if logLevel == "" {
```
arbulu89, line 57: As commented before, this looks strange as we should use the default value and not this comparison

cmd/root.go
arbulu89: By the way, the description of the flag has a typo (`then`).
Being picky, I would rewrite it to something like `The minimum logging level; levels are, in ascending order: debug, info, warn, error"` based on the `sap_host_exporter` code

arbulu89 (commented):
Hey @aleksei-burlakov ,
I know that the PR is merged, but the usage of the default value for the `log-level` flag looks wrong (actually, the not usage of the default value)

PR #108, opened 2021-06-22
Add self-hosted actions runner continuous deployment

fabriziosestito:
This PR adds continuous deployment jobs to the GitHub actions.

GitHub self-hosted runners are used for the deployment process, this is triggered by pushing or merging  PRs to the main branch or manually by using a workflow dispatch event.
Every node needs a running instance of a self-hosted runner and consul. 
A provisioning script which bootstraps Consul, Trento and the runner environment was added to the hack folder.

By configuring the runners on forks, it is possible to deploy on private development nodes.
Please refer to the [hack/README.md ](https://github.com/fabriziosestito/trento/blob/actions/hack/README.md) for setup instructions and detailed documentation.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -54,12 +54,8 @@ jobs:
         with:
           name: trento-amd64-binary
           path: build/trento-amd64.gz
-      - uses: actions/upload-artifact@v2
-        with:
-          name: trento-examples-rules
-          path: examples
 
-  consul-cleanup-monitoring:
+  consul-cleanup:
     runs-on: [ self-hosted, vmmonitoring ]
     needs: build
     if: (github.event_name == 'push' && github.ref == 'refs/heads/main') || github.event_name == 'workflow_dispatch'
@@ -70,7 +66,7 @@ jobs:
 
   deploy-hana01:
     runs-on: [ self-hosted, vmhana01 ]
-    needs: consul-cleanup-monitoring
+    needs: consul-cleanup
     if: (github.event_name == 'push' && github.ref == 'refs/heads/main') || github.event_name == 'workflow_dispatch'
     steps:
       - uses: actions/checkout@v2
@@ -79,9 +75,6 @@ jobs:
       - uses: actions/download-artifact@v2
         with:
           name: trento-amd64-binary
-      - uses: actions/download-artifact@v2
-        with:
-          name: trento-examples-rules
       - id: deploy
         uses: ./.github/actions/deploy
         with:
@@ -89,7 +82,7 @@ jobs:
 
   deploy-hana02:
     runs-on: [ self-hosted, vmhana02 ]
-    needs: consul-cleanup-monitoring
+    needs: consul-cleanup
     if: (github.event_name == 'push' && github.ref == 'refs/heads/main') || github.event_name == 'workflow_dispatch'
     steps:
       - uses: actions/checkout@v2
@@ -98,9 +91,6 @@ jobs:
       - uses: actions/download-artifact@v2
         with:
           name: trento-amd64-binary
-      - uses: actions/download-artifact@v2
-        with:
-          name: trento-examples-rules
       - id: deploy
         uses: ./.github/actions/deploy
         with:
@@ -108,7 +98,7 @@ jobs:
 
   deploy-monitoring:
     runs-on: [ self-hosted, vmmonitoring ]
-    needs: consul-cleanup-monitoring
+    needs: consul-cleanup
     if: (github.event_name == 'push' && github.ref == 'refs/heads/main') || github.event_name == 'workflow_dispatch'
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,6 +1,7 @@
 on:
   push:
   pull_request:
+  workflow_dispatch:
   release:
     types: [ published, created, edited ]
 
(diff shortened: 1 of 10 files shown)


hack/provision.sh
```diff
@@ -0,0 +1,186 @@
...
+if [ "$ROLE" = "agent" ]; then
+  CONSUL_HCL="consul-client.hcl"
+elif [ "$ROLE" = "web" ]; then
+  CONSUL_HCL="consul-server.hcl"
+else
+  echo "Please specify a valid role"
+  exit
+fi
+
+CONSUL_HCL_TEMPLATE_URL="https://raw.githubusercontent.com/trento_project/trento/main/hack/$CONSUL_HCL.template"
```
rtorrero, line 46: Sorry that I didn't ask this before, but wouldn't it be possible to also use here `$ACTIONS_RUNNER_REPO_OWNER` and `$ACTIONS_RUNNER_REPO_NAME`? I was just testing the script on my fork and had to manually edit this.

rtorrero:
This is going to be *really* helpful for us, glad to work with you on this @fabriziosestito !

hack/README.md
```diff
@@ -1,28 +1,104 @@
...
+
+#### Provision
+
+[`provision.sh`](./provision.sh) provides an easy way to boostrap a GitHub self-hosted runner environment on a private
+node.
+
+The script sets up users, configuration files, systemd unit and a self-hosted runner instance required in order to run
+Trento, Consul and the deployment process.
+
+`# ./setup.sh <github-user> <github-repo> <actions-runner-token> <actions-runner-name> <consul-bind-ip> <consul-server-ip> [agent|web]`
```
rtorrero, line 65: You probably missed to update the name of the script here. This would be `provision.sh` right?

hack/provision.sh
```diff
@@ -0,0 +1,186 @@
...
+if [ "$EUID" -ne 0 ]; then
+  echo "Please run as root"
+  exit
+fi
+
+echo ""
+echo "Provisioning..."
+
+if [ $# -lt 7 ]; then
+  echo "Usage: ./setup.sh <github-user> <github-repo> <actions-runner-token> <actions-runner-name> <consul-bind-ip> <consul-server-ip> [agent|web]"
```
rtorrero, line 14: Here too.

hack/provision.sh
fabriziosestito: Not sure about this since it's not gonna work till it's merged because of the hardcoded branch name. Adding an optional branch parameter just for this seems a bit overkill.

arbulu89 (commented):
@fabriziosestito @rtorrero
It looks great to have an updated working instance up and running all the time.

Besides this, my 2 cents:
- I would split the CI and provisioning jobs in 2 files (or 2 logical components). I think both have different purposes, and I wouldn't like to have the CI failing due provisioning failure
-  The daemons look something that might be interesting out of the CI. Maybe moving them to the a new folder (`daemons`) would be interesting, to have the more accessible.

Offtopic:
- I know that this goes out of scope, but I think we should put some effort in the current `deploy.sh` script to make an easy installation out of github actions. Something easy to use from people that don't use github actions (most of the user base). Maybe this script, could download the consul/trento binaries, and use the daemons to install the whole app

rtorrero:
> @fabriziosestito @rtorrero
> It looks great to have an updated working instance up and running all the time.
> 
> Besides this, my 2 cents:
> 
>     * I would split the CI and provisioning jobs in 2 files (or 2 logical components). I think both have different purposes, and I wouldn't like to have the CI failing due provisioning failure

Not sure I follow here; the provisioning.sh file is already separated from the CI stuff and supposed to be required only once. Do you propose any further separation?

>     * The daemons look something that might be interesting out of the CI. Maybe moving them to the a new folder (`daemons`) would be interesting, to have the more accessible.
This is an idea we already had. Probably, the trento systemd units could be reused for packaging but they still require some changes about the running users and some other changes to make them more production ready; I'd try to address this in the future (as it's going to unnecessarily complicate the CI for now) when we address packaging, but you are right
> 
> Offtopic:
> 
>     * I know that this goes out of scope, but I think we should put some effort in the current `deploy.sh` script to make an easy installation out of github actions. Something easy to use from people that don't use github actions (most of the user base). Maybe this script, could download the consul/trento binaries, and use the daemons to install the whole app

The installation of the github actions is currently done in the provision.sh. But about the deploy.sh improvements; yes, we should do this. I'm going to take a look and see what from this PR can be used to improve it.

arbulu89:
> > @fabriziosestito @rtorrero
> > It looks great to have an updated working instance up and running all the time.
> > Besides this, my 2 cents:
> > ```
> > * I would split the CI and provisioning jobs in 2 files (or 2 logical components). I think both have different purposes, and I wouldn't like to have the CI failing due provisioning failure
> > ```
> 
> Not sure I follow here; the provisioning.sh file is already separated from the CI stuff and supposed to be required only once. Do you propose any further separation?
> 
Yes, I would like to have 2 different jobs. 1 for the CI and a 2nd for the provisioning (which depends on the 1st job passing). This way, if for some reason the provisioning fails, the CI job would be still passing, which helps to find out what's going on super easily.
> > ```
> > * The daemons look something that might be interesting out of the CI. Maybe moving them to the a new folder (`daemons`) would be interesting, to have the more accessible.
> > ```
> 
> This is an idea we already had. Probably, the trento systemd units could be reused for packaging but they still require some changes about the running users and some other changes to make them more production ready; I'd try to address this in the future (as it's going to unnecessarily complicate the CI for now) when we address packaging, but you are right
> 
> > Offtopic:
> > ```
> > * I know that this goes out of scope, but I think we should put some effort in the current `deploy.sh` script to make an easy installation out of github actions. Something easy to use from people that don't use github actions (most of the user base). Maybe this script, could download the consul/trento binaries, and use the daemons to install the whole app
> > ```
> 
> The installation of the github actions is currently done in the provision.sh. But about the deploy.sh improvements; yes, we should do this. I'm going to take a look and see what from this PR can be used to improve it.

fabriziosestito:
> Yes, I would like to have 2 different jobs. 1 for the CI and a 2nd for the provisioning (which depends on the 1st job passing). This way, if for some reason the provisioning fails, the CI job would be still passing, which helps to find out what's going on super easily.

@arbulu89 The jobs are already configured this way. 
The deployment jobs tree depends on the CI passing (build job) and it starts only when merging PRs on master or manually. 
The ci job (build) is independent and it gets triggered at every push on every branch. See the checks below for instance, the CI passed but the deployment is skipped.

They are jobs of the same workflow since it's easier to manage dependencies  and share artefacts.
We can change this but the trade off is that we need to upload the artefacts somewhere to download them in the other workflow.

arbulu89:
> > Yes, I would like to have 2 different jobs. 1 for the CI and a 2nd for the provisioning (which depends on the 1st job passing). This way, if for some reason the provisioning fails, the CI job would be still passing, which helps to find out what's going on super easily.
> 
> @arbulu89 The jobs are already configured this way.
> The deployment jobs tree depends on the CI passing (build job) and it starts only when merging PRs on master or manually.
> The ci job (build) is independent and it gets triggered at every push on every branch. See the checks below for instance, the CI passed but the deployment is skipped.
> 
> They are jobs of the same workflow since it's easier to manage dependencies and share artefacts.
> We can change this but the trade off is that we need to upload the artefacts somewhere to download them in the other workflow.
> 
> Also, we can select which jobs are required individually, so the branch protection rule only applies to the build job.

Ok them, I buy your explanation

arbulu89 (dismissed):

fabriziosestito:
@arbulu89 as @rtorrero proposed maybed this could make it for a mob-review since everyone is gonna use it.
It could be an occasion to receive more feedbacks and explain a little bit what are the reasons of certain choices.

stefanotorresi (dismissed):
Overall, really good work, guys! I have some reserves, though.
We'll need to have a live chat about some details, but for the time being, here is a first round of feedback.

.github/workflows/ci.yaml
```diff
@@ -1,54 +1,126 @@
...
+      - uses: actions/download-artifact@v2
+        with:
+          name: trento-amd64-binary
+      - uses: actions/download-artifact@v2
+        with:
+          name: trento-examples-rules
+      - id: deploy
+        uses: ./.github/actions/deploy
+        with:
+          role: agent
```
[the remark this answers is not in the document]
fabriziosestito: We weren't able to target two nodes with the same job since the `runs-on` directive expects all the tags to be present.
I will investigate more on the topic and get back if I find something!

.github/workflows/ci.yaml
```diff
@@ -1,54 +1,126 @@
...
+          name: trento-binaries
+          path: build
+      - uses: actions/upload-artifact@v2
+        with:
+          name: trento-amd64-binary
+          path: build/trento-amd64.gz
+      - uses: actions/upload-artifact@v2
+        with:
+          name: trento-examples-rules
+          path: examples
```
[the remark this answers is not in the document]
fabriziosestito: Oh yeas, this is a leftover since we weren't using the checkout in the first experiments. Good call!

.github/workflows/ci.yaml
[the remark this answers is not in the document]
fabriziosestito: Also if the `runs-on` tags filter results in multiple nodes, just one action is executed in a round robin fashion, so just one runner will be actually running the deploy job.

rtorrero:
The requested changes where applied with the exception of merging the two deploy actions to the different hana nodes. As discussed this is currently not possible.

rtorrero:
Just realized we need to rebase, doing now

rtorrero (approved):
LGTM

PR #111, opened 2021-06-24
Add version command to CLI

### LINKED ISSUE #104: Add version command
It should be possible to display the version number via `trento version`.

We can take this as inspiration to inject the current tag to a version package constant:
https://github.com/SUSE/sap_host_exporter/blob/master/.ci/get_version_from_git.sh

And in the Makefile:
https://github.com/SUSE/sap_host_exporter/blob/68bbf2f1b490ab0efaa2dd7b878b778f07fba2ab/Makefile#L2

https://github.com/SUSE/sap_host_exporter/blob/68bbf2f1b490ab0efaa2dd7b878b778f07fba2ab/Makefile#L32

fabriziosestito:
This PR adds the `version` command to trento CLI by injecting the version number when building.

Resolves #104

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,5 @@
-GO_BUILD = CGO_ENABLED=0 go build -trimpath -ldflags "-s -w"
+VERSION ?= $(shell hack/get_version_from_git.sh)
+GO_BUILD = CGO_ENABLED=0 go build -trimpath -ldflags "-s -w -X github.com/trento-project/trento/cmd.version=$(VERSION)"
 ARCHS ?= amd64 arm64 ppc64le s390x
 
 default: clean mod-tidy fmt vet-check test build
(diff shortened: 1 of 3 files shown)


.ci/get_version_from_git.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
```
stefanotorresi, line 1: I would probably move this script into the `hack` directory, instead, since we have that to hold various scripts.

stefanotorresi (changes requested):

arbulu89 (approved):
@fabriziosestito It looks good.
Could you paste in one of the GH comments an output example?
To see how it looks like

fabriziosestito:
> @fabriziosestito It looks good.
> Could you paste in one of the GH comments an output example?
> To see how it looks like

Peek 2021-06-28 09-28

stefanotorresi (approved):

PR #112, opened 2021-06-25
Fix: polish the verion checks

aleksei-burlakov:
According to the last comment in the ticket related to the vrsion checks CFSA-83:

Check 2.2.6: Agree to skip SAPHanaSR-doc

Check 2.2.1
Description:  Check SLES4SAP is installed
Remediation: SAPHanaSR is only supported on SLES4SAP , please install
SLES4SAP.

Check 2.2.2: Description: "Check OS version"

### DIFF
--- a/examples/hana-scale-up-perf-optimized-azure.yaml
+++ b/examples/hana-scale-up-perf-optimized-azure.yaml
@@ -304,7 +304,7 @@ groups:
     description: "Check OS and package versions are supported and up-to-date"
     checks:
       - id: 2.2.1
-        description: "Check the base product"
+        description: "Check a supported OS distribution is being used"
         # Assume the perf-optimized scenario and
         # check only that the base probuct is always sles_sap.prod (case-insensitively)
         audit: 'readlink /etc/products.d/baseproduct | sed -e "s/\(.*\)/\L\1/"'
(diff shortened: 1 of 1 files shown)


examples/hana-scale-up-perf-optimized-azure.yaml
```diff
@@ -304,18 +304,23 @@ groups:
...
         audit: 'readlink /etc/products.d/baseproduct | sed -e "s/\(.*\)/\L\1/"'
         tests:
           test_items:
             - flag: 'sles_sap.prod'
         remediation: |
-          Install a SUSE supported OS release
+          ## Abstract
+          You need to use SUSE Linux Enterprise Server for SAP Applications
+          ## Remediation
+          SAPHanaSR is only supported on SLES4SAP , please install SLES4SAP.
```
stefanotorresi, line 318: I would suggest to switch these two: the abstract should explain why a check is in place, and the remediation tells the user what they need to do.

examples/hana-scale-up-perf-optimized-azure.yaml
```diff
@@ -304,18 +304,23 @@ groups:
     description: "Check OS and package versions are supported and up-to-date"
     checks:
       - id: 2.2.1
-        description: "Check the base product"
+        description: "Check SLES4SAP is installed"
```
stefanotorresi, line 307: Maybe we should make this more generic, so that this could be in future parameterised to different values.
Something like:
```suggestion
        description: "Check a supported OS distribution is being used"
```

examples/hana-scale-up-perf-optimized-azure.yaml
stefanotorresi, line 318: Also, we should never use the `SLES4SAP` moniker in user-facing stuff, we should always use the full product name, `SUSE Linux Enterprise Server for SAP Applications`.

stefanotorresi (changes requested):

aleksei-burlakov:
> I would suggest to switch these two: the abstract should explain why a check is in place, and the remediation tells the user what they need to do.
> Maybe we should make this more generic, so that this could be in future parameterised to different values.
Something like:
> Also, we should never use the SLES4SAP moniker in user-facing stuff, we should always use the full product name, SUSE Linux Enterprise Server for SAP Applications.

Done!

stefanotorresi (approved):

stefanotorresi:
Thank you, Aleksei!

PR #113, opened 2021-06-28
Add: Set log timestamp format

aleksei-burlakov:
Without setting the formatter, the logrus shows zeroes instead of timestamps.


arbulu89 (approved):
Nice @aleksei-burlakov

Issue #88, opened 2021-06-11
Identify and store cluster entries with unique ID in the KV store


arbulu89:
By now the cluster entries are identified by the cluster name. The cluster name is not unique at all, and actually, it is not even mandatory to have, it is yet another attribute of the cluster.

By now, we store the data as:

`trento/v0/cluster/$clustername/...`

We need to change this $clustername by something unique. The cluster doesn't have any widely unique attribute unfortunately, but we could use the corosync authentication key for that (get the md5sum of it). This file is located at `/etc/corosync/auth`, or maybe create some other combination of data to create the unique permanent id of the cluster.

After that, we would need to store the cluster name as well, to use it the frontend, and if it doesn't exist, use some autogenerated name (`london`, `paris`, you know, these kind of names...).

The frontend needs to use this name in all of the visualization, and take care to do something if 2 cluster with the same name are available (maybe add something like `hana-cluster*1`, `hana-cluster*2`).

Finally, we need to manage the `ha-cluster` metadata tags as well, to use the proper tag in consul filterings.

arbulu89:
As a comment, the cluster name is just an attribute: `cib-bootstrap-options-cluster-name`.

rtorrero:
Thanks @arbulu89 , I already started working on this and should have something soon

arbulu89:
Closing as it is implemented (at least the 1st iteration) at https://github.com/trento-project/trento/pull/114

PR #114, opened 2021-06-28
Generate new consistent ID and human readable name for clusters

rtorrero:
This PR addresses [issue#88](https://github.com/trento-project/trento/issues/88)

- The new ID is based on the `/etc/corosync/authkey` file that each cluster node should have:
  - It hashes the key using md5sum. The resulting 32 digit hex string is used as unique ID.
  - Using MD5 as source for the ID means that even when we reset the data in consul, the ID of each cluster will remain the same.
  
- This PR also generates names for the cluster that can be easily remembered as often clusters will have identical generic names; preventing users for correctly distinguishing them.
  - We take advantage of the md5-based ID that we generate for the cluster and use it as random seed.
  - As the number is too large for an integer, we use crc32 of the ID and use this to feed the pseudorandom number generator.
  - Then we use `golang-petname` to generate a name based on [RFC1178](https://datatracker.ietf.org/doc/html/rfc1178)
  - Since we always use the same seed for the RNG, even after a consul data wipe, the generated name will remain stable.

Question: Right now, we are just using one word for the random name. This can be increased to anything, but probably 2 is more advisable?


internal/cluster/cluster.go
```diff
@@ -24,6 +28,8 @@ type Cluster struct {
 	Cib    cib.Root    `mapstructure:"cib,omitempty"`
 	Crmmon crmmon.Root `mapstructure:"crmmon,omitempty"`
 	SBD    SBD         `mapstructure:"sbd,omitempty"`
+	Id     string      `mapstructure:"id"`
+	Name   string      `mapstructure:"name,omitempty"`
```
arbulu89, line 32: Minor details
Maybe we should remove the `omitempty` from `Name` as without the name other functionalities could fail.
What do you think?

internal/cluster/cluster.go
```diff
@@ -59,15 +73,20 @@ func NewCluster() (Cluster, error) {
...
-		}
+func getCorosyncAuthkey(corosyncKeyPath string) (string, error) {
+	kp, err := internal.Md5sum(corosyncKeyPath)
+	return kp, err
+}
+
+func getName() string {
+	keyStr, err := getCorosyncAuthkey(corosyncKeyPath)
+	if err != nil {
+		return ""
```
arbulu89, line 84: I don't know if we should return an empty string here. The unique reason to have an error here is that the corosync key file doesn't exist, so maybe we should handle in a better way.
But maybe there is a reason to have it this way

PD: In any case, as this is executed after executing `getCorosyncAuthkey` already in the `NewCluster` it shouldn't never fail I guess

internal/cluster/cluster_kvstore_test.go
```diff
@@ -29,7 +29,7 @@ func TestStore(t *testing.T) {
 
 	consulInst.On("KV").Return(kv)
 
-	kvPath := fmt.Sprintf("%s%s", consul.KvClustersPath, "cluster_name")
+	kvPath := path.Join(consul.KvClustersPath, "")
```
arbulu89, line 32: I guess we can directly remove this line and use `consul.KvClustersPath`

web/clusters.go
```diff
@@ -13,6 +13,7 @@ import (
 
 type Cluster struct {
 	Name string
+	Id   string
```
arbulu89, line 16: I don't know if we use this `Cluster` and `ClusterList` types for something. Maybe it is just an old not removed data.
If this is the case we could remove them

internal/hosts/metadata.go
```diff
@@ -2,6 +2,7 @@ package hosts
 
 type Metadata struct {
 	Cluster     string `mapstructure:"ha-cluster,omitempty"`
+	ClusterId   string `mapstructure:"ha-cluster-id,omitempty"`
```
arbulu89, line 5: Now that we store the `ClusterId`, I don't know if we need the `Cluster` entry (the name). This is just used to filter the consul catalog. Maybe for the future if we add some filters by name?

arbulu89 (changes requested):
Hey @rtorrero ,
The code looks good. I think the petname usage is valid (at least at this point of the project), maybe we could change it in the future, but it is as simple as changing some few functions. If somebody has some easy/straightforward improvement for this, we could work it in this PR, otherwise I would let it as it is

My unique change request would be to try to add some new UT to test this new functionality. To test the `/etc/corosync/authkey` reading and operations applied to this file. I think that it is a good learning practice for unit testing. You could make the `corosyncKeyPath` a variable, and change it on the tests.

internal/cluster/cluster.go
```diff
@@ -59,15 +73,20 @@ func NewCluster() (Cluster, error) {
 	return cluster, nil
 }
 
-func (c *Cluster) Name() string {
-	// Handle not named clusters
-	for _, prop := range c.Cib.Configuration.CrmConfig.ClusterProperties {
-		if prop.Id == clusterNameProperty {
-			return prop.Value
-		}
+func getCorosyncAuthkey(corosyncKeyPath string) (string, error) {
```
arbulu89, line 76: Maybe we should rename the method to `getCorosyncAuthkeyMd5`

internal/cluster/cluster.go
rtorrero: Agreed, if Id exists then same should too.

internal/cluster/cluster.go
rtorrero: noted!

internal/hosts/metadata.go
rtorrero: For now I'm using it to display the name (while we use the ID on the link for the href). This is a bit lazy and we could just pull that information, but since it was there already I thought it might be simpler for now to keep it. WDYT?

internal/cluster/cluster.go
rtorrero: Agreed

web/clusters.go
rtorrero: I 've check and yes, I auto-edited this but it's not really used anyway

rtorrero:
I've added the suggested changes, thanks for the input!

internal/hosts/metadata.go
arbulu89: Ok, it is fine

arbulu89 (approved):
Hi @rtorrero .
From my side the code changes look good.
Let's wait to see what other people think about this

fabriziosestito (approved):

PR #115, opened 2021-06-28
Collect hdbnsutil -sr_state output

arbulu89:
`hdbnsutil -sr_state` command collection on the KV storage (only applies to HANA).

Basically, it calls the command and stores the information.
Unfortunately the output of this command is not that good (spaces on the keys, duplicated keys...), so I have needed to do some changes in the `utils.go` and kv functions.

Find an example of the output at: https://github.com/trento-project/trento/compare/main...arbulu89:feature/hdbnsutil_srstate?expand=1#diff-3cb2d1a90eb474b2de290c02412e79c94d2d5ea4066ee823dd3fc400d86def00

I took advantage to remodelate a bit the KV function to make them a bit cleaner and add some new UT scenario.

The data is not used yet in the frontend, but it will at some point.

This is the kv data:
image
image
image

### DIFF
--- a/internal/sapsystem/sapsystem.go
+++ b/internal/sapsystem/sapsystem.go
@@ -47,6 +47,7 @@ type SAPSystem struct {
 type SAPProfile map[string]interface{}
 type SystemReplication map[string]interface{}
 type HostConfiguration map[string]interface{}
+type HdbnsutilSRstate map[string]interface{}
 
 type SAPInstance struct {
 	Name       string      `mapstructure:"name,omitempty"`
@@ -56,6 +57,7 @@ type SAPInstance struct {
 	// Only for Database type
 	SystemReplication SystemReplication `mapstructure:"systemreplication,omitempty"`
 	HostConfiguration HostConfiguration `mapstructure:"hostconfiguration,omitempty"`
+	HdbnsutilSRstate  HdbnsutilSRstate  `mapstructure:"hdbnsutilsrstate,omitempty"`
 }
 
 type SAPControl struct {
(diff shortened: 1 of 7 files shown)


internal/sapsystem/sapsystem.go
```diff
@@ -255,6 +258,15 @@ func landscapeHostConfiguration(sid, instance string) map[string]interface{} {
 	return runPythonSupport(sid, instance, "landscapeHostConfiguration.py")
 }
 
+func hdbnsutilSrstate(sid, instance string) map[string]interface{} {
+	user := fmt.Sprintf("%sadm", strings.ToLower(sid))
+	cmdPath := path.Join("/usr/sap", sid, instance, "exe", "hdbnsutil")
```
rtorrero, line 263: We should probably use `sapInstallationPath` there instead of `"/usr/sap"` right?

internal/utils.go
```diff
@@ -56,7 +57,16 @@ func FindMatches(pattern string, text []byte) map[string]interface{} {
 	r := regexp.MustCompile(pattern)
 	values := r.FindAllStringSubmatch(string(text), -1)
 	for _, match := range values {
-		configMap[match[1]] = match[2]
+		key := strings.Replace(match[1], " ", "_", -1)
```
rtorrero, line 60: Not really related to this PR but it would probably make sense from now on to start adding comments to the functions that we add to `utils.go` to promote that they get reused when it's possible

rtorrero:
@arbulu89  Other than what I commented, LGTM!

arbulu89:
@rtorrero Changes done

aleksei-burlakov:
I have not found anything to complain of :)

aleksei-burlakov:
Looks good to me.

arbulu89:
@rtorrero @fabriziosestito @stefanotorresi @aleksei-burlakov I need a review here, please

aleksei-burlakov (approved):
LGTM.

PR #117, opened 2021-06-29
Add: Check if the SAP Host Agent is installed

aleksei-burlakov:


aleksei-burlakov:
@fmherschel could you please have a look at this check. In Section 6.1 in https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-sr-guide-PerfOpt-15/ we ask users to "Check if the SAP Host Agent is installed". As I understand we can do it from any sid-user?

fmherschel:
> @fmherschel could you please have a look at this check. In Section 6.1 in https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-sr-guide-PerfOpt-15/ we ask users to "Check if the SAP Host Agent is installed". As I understand we can do it from any sid-user?

As SAPHostAgent is "neutral" to SIDs you can even check it as root.

examples/hana-scale-up-perf-optimized-azure.yaml
```diff
@@ -453,3 +453,20 @@ groups:
           ## References
           - section 2 in https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-sr-guide-PerfOpt-15/
         scored: true
+      - id: 2.2.9
+        description: "SAP Host Agent is installed"
+        audit: |
+          sid=$(sudo crm configure show | grep -m1 SID= | sed -e "s/.*SID=\(...\).*/\1/" | tr '[:upper:]' '[:lower:]')
+          inst=$(sudo crm configure show | grep -m1 InstanceNumber= | sed -e "s/.*InstanceNumber=\(..\).*/\1/")
+          sudo runuser -l ${sid}adm -c "sapcontrol -nr $inst -function CheckHostAgent"
```
fmherschel, line 461: Now I understand why you are asking, if you can check from any <sid>adm user.

fmherschel: An alternative would be (as root):
cd /usr/sap/hostctrl/exe
./saphostctrl -function Ping   (and/or)
./sapcontrol -nr 99 -function CheckHostAgent

In your case you might get false positives, if the sapstartsrv for the found instance is not available, because it is running on a different cluster node,

fmherschel (changes requested):
I would like to request a change to use the either saphostctr or the sapcontrol of the "central" sapstartsrc (instance nimber 99) as user root.
This avoids false-positives as sapstarsrv of a given instance is not always running. Only a one of the nodes per time.

examples/hana-scale-up-perf-optimized-azure.yaml
aleksei-burlakov: @fmherschel Thank you for the advice, I have amended the PR!

rtorrero:
@aleksei-burlakov this needs to be rebased, could you take care of it?

aleksei-burlakov:
> @aleksei-burlakov this needs to be rebased, could you take care of it?

Rebased!

stefanotorresi (approved):

examples/hana-scale-up-perf-optimized-azure.yaml
lee-martin: @fmherschel Thanks also for the feedback.
Regarding the potential false positives - isn't the Host Agent always local to each cluster node? If so I assume we would be safe from false positives for the Host Agent. 

I guess instance agents/sapstartsrv are the ones to be careful with?

fmherschel: @lee-martin I am not sure, if I understand your comment. False-Positives are always bad. As well as False-Negatives. If we feel save with a planend false test, why do we implement it? And the code I requested a check was running against the instance sapstartsrv and not against the real saphostagent.

fmherschel:
@aleksei-burlakov : Could we close that issue a sit seams to be merged already?

PR #118, opened 2021-06-29
Add: Check technical OS users and groups for SAP are defined locally …

aleksei-burlakov:
…on Linux Host

### DIFF
--- a/examples/generic-azure.yaml
+++ b/examples/generic-azure.yaml
@@ -673,3 +673,73 @@ groups:
           ## References
           - section 9.1.2 https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-sr-guide-PerfOpt-15/
         scored: true
+      - id: 1.5.3
+        description: "<sid>adm user and its <sid> group exist and enabled"
+        audit: |
+          # Get SIDs from cibadmin
+          IFS=' ' read -r -a sid_list_raw <<< $(sudo cibadmin -Q | xmllint -xpath '//primitive//nvpair[@name="SID"]/@value' -)
+          # Parse sids and makes them unique. There is possibly a more beautiful way to do it with regexp-s rather than with loops.
+          sid_list_not_unique=$(for i in "${sid_list_raw[@]}"; do echo $i | sed -e "s/.*value=\"\(...\)\".*/\1/" | tr '[:upper:]' '[:lower:]' ; done)
+          sid_list_sorted_and_unique=$(echo ${sid_list_not_unique[@]} | tr ' ' '\n' | sort -u | tr '\n' ' ')
+
+          # Check the /etc/nsswitch.conf
+          if ! cat /etc/nsswitch.conf | grep "passwd:" | grep -q "compat\|files"; then echo failed=true; fi
+          if ! cat /etc/nsswitch.conf | grep "group:" | grep -q "compat\|files"; then echo failed=true; fi
+
+          # Check the /etc/passwd and /etc/group
+          for sid in ${sid_list_sorted_and_unique}
+          do
+            if ! cat /etc/passwd | cut -d: -f1 | grep -q ${sid}adm; then echo failed=true; fi
+            if cat /etc/passwd | grep -e "^${sid}adm" | cut -d: -f7 | grep -q "nologin"; then echo failed=true; fi
+            if ! cat /etc/group | cut -d: -f1,4 | grep "^sapsys:" | grep -q ${sid}adm; then echo failed=true; fi
+          done
+        tests:
+          test_items:
+            - flag: failed
+              compare:
+                op: noteq
+                value: true
+        remediation: |
+          ## Abstract
+          The technical users <sid>adm is in the sapsys group and he is defined locally.
+
+          ## Remediation
+          Create and enable the technical user and group.
+
+          ## References
+          - Section 2 in https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-sr-guide-PerfOpt-15/
+        scored: true
+      - id: 1.5.4
+        description: "<sid>adm user should not be in an Active Directory"
+        audit: |
+          if cat /etc/krb5.conf | grep -q "default_realm ="; then
+
+            # Get SIDs from cibadmin
+            IFS=' ' read -r -a sid_list_raw <<< $(sudo cibadmin -Q | xmllint -xpath '//primitive//nvpair[@name="SID"]/@value' -)
+            # Parse sids and makes them unique. There is possibly a more beautiful way to do it with regexp-s rather than with loops.
+            sid_list_not_unique=$(for i in "${sid_list_raw[@]}"; do echo $i | sed -e "s/.*value=\"\(...\)\".*/\1/" | tr '[:upper:]' '[:lower:]' ; done)
+            sid_list_sorted_and_unique=$(echo ${sid_list_not_unique[@]} | tr ' ' '\n' | sort -u | tr '\n' ' ')
+
+            # Check if the user exists and reachable from active directory
+            for sid in ${sid_list_sorted_and_unique}
+            do
+              realm=$(cat /etc/krb5.conf | grep "default_realm =" | sed "s/^.*=//" | xargs)
+              if kinit -R ${sid}adm@$realm; then echo ad_admuser_exists=true; fi
+            done
+          fi
+        tests:
+          test_items:
+            - flag: ad_admuser_exists
+              compare:
+                op: noteq
+                value: true
+        remediation: |
+          ## Abstract
+          The technical users <sid>adm should not be reachable via Active Directory.
+
+          ## Remediation
+          Create and enable the technical user locally.
+
+          ## References
+          - Section 2 in https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-sr-guide-PerfOpt-15/
+        scored: false


aleksei-burlakov:
@angelabriel @lee-martin could you have a look, please?

examples/generic-azure.yaml
```diff
@@ -673,3 +673,26 @@ groups:
           ## References
           - section 9.1.2 https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-sr-guide-PerfOpt-15/
         scored: true
+      - id: 1.5.3
+        description: "<sid>adm user and its <sid> group exist and enabled"
+        audit: |
+          sid=$(sudo crm configure show | grep -m1 SID= | sed -e "s/.*SID=\(...\).*/\1/" | tr '[:upper:]' '[:lower:]')
+          if cat /etc/passwd | cut -d: -f1 | grep ${sid}adm; then echo userexists; fi
+          if ! cat /etc/passwd | grep -e "^${sid}adm" | cut -d: -f7 | grep "nologin"; then echo userenabled; fi
+          if cat /etc/group | cut -d: -f1,4 | grep "${sid}...:{$sid}adm"; then echo groupexists; fi
```
lee-martin, line 682: @aleksei-burlakov  it looks like you are only checking if the sidadm user exists in any group, but the important part is to make sure sidadm is in the sapsys group. Please take a look at your test system, I would expect the sidadm user to be a member of sapsys, and this is what we want to check for.  Also good idea to verify with @angelabriel of course .

angelabriel:
@aleksei-burlakov in addition to @lee-martin 2 points from my side:
1.  Even that we do not official support the use of Active Directory, many customers - especially in cloud environments - simply use an AD integration instead of local defined users. And this often cause a lot of trouble.
It would be really helpful, if we can check this situation and give a big warning, that this is unsupported (at the moment).
So not only report 'no local users defined', but additional something like 'use of AD not supported'

2. the user sidadm needs to be member of the group 'haclient' too, if the customer is using 'sap_suse_cluster_connector' for HA (ENSA1 or ENSA2).
So this needs to be checked too. If not in this check, then we need an additional check.

fmherschel:
May I also add my thoughts?
1. Please do NOT assume there is only one single SID in the cluster. We allow multiple in different resources.
2. Testing /etc/passwd is not sufficient because the entry could be replaced by a remote definition. 
     I do nt know if pam gives us a change to calculate, if a user resolution is finally made locally or using remotes.

examples/generic-azure.yaml
```diff
@@ -673,3 +673,26 @@ groups:
           ## References
           - section 9.1.2 https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-sr-guide-PerfOpt-15/
         scored: true
+      - id: 1.5.3
+        description: "<sid>adm user and its <sid> group exist and enabled"
+        audit: |
+          sid=$(sudo crm configure show | grep -m1 SID= | sed -e "s/.*SID=\(...\).*/\1/" | tr '[:upper:]' '[:lower:]')
```
fmherschel, line 679: We need to handle multi-sid also. Do like me to provide a multi-sid code here (SIDs=...)?

examples/generic-azure.yaml
```diff
@@ -673,3 +673,26 @@ groups:
...
+          if cat /etc/passwd | cut -d: -f1 | grep ${sid}adm; then echo userexists; fi
+          if ! cat /etc/passwd | grep -e "^${sid}adm" | cut -d: -f7 | grep "nologin"; then echo userenabled; fi
```
fmherschel, line 681: We also need to check other ways to "disable" users. (eg first character disables the user).

examples/generic-azure.yaml
```diff
@@ -673,3 +673,26 @@ groups:
           ## References
           - section 9.1.2 https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-sr-guide-PerfOpt-15/
         scored: true
+      - id: 1.5.3
+        description: "<sid>adm user and its <sid> group exist and enabled"
+        audit: |
+          sid=$(sudo crm configure show | grep -m1 SID= | sed -e "s/.*SID=\(...\).*/\1/" | tr '[:upper:]' '[:lower:]')
+          if cat /etc/passwd | cut -d: -f1 | grep ${sid}adm; then echo userexists; fi
```
angelabriel, line 680: @aleksei-burlakov is it intended, that the 'sid'adm user name is displayed in addition to the 'userexists'?
Because of the 'grep' command you will get something like
prdadm
userexists
as the result.

fmherschel, line 680: Also need to take other files into account. eg. nsswitch conf could tell that first a remote source is been used.
I do not know if any pam-functionality could tell us, if a user-name is finally resolved locally or remote.

examples/generic-azure.yaml
```diff
@@ -673,3 +673,26 @@ groups:
...
+          if ! cat /etc/passwd | grep -e "^${sid}adm" | cut -d: -f7 | grep "nologin"; then echo userenabled; fi
+          if cat /etc/group | cut -d: -f1,4 | grep "sapsys:{$sid}adm"; then echo groupexists; fi
```
fmherschel, line 682: Also need to take other files into account. eg. nsswitch conf could tell that first a remote source is been used.
I do not know if any pam-functionality could tell us, if a group-name is finally resolved locally or remote.

fmherschel (changes requested):
I would like to request the changes named in my comments. For the multi-sid I have an idea how to get this done using xmllint/xpath.

examples/generic-azure.yaml
angelabriel, line 682: @aleksei-burlakov 
1. type {$sid}adm - needs to be ${sid}adm
2. if more than one user is added as member to the sapsys group, the grep will not work.

fmherschel:
> @aleksei-burlakov in addition to @lee-martin 2 points from my side:

>     2. the user sidadm needs to be member of the group 'haclient' too, if the customer is using 'sap_suse_cluster_connector' for HA (ENSA1 or ENSA2).
>        So this needs to be checked too. If not in this check, then we need an additional check.

Yes, this is for ENSA1 and ENSA2. For HANA we need sudoers entries to be checked, so I would move these checks to
separate tests.

angelabriel (changes requested):

examples/generic-azure.yaml
fmherschel: > @aleksei-burlakov it looks like you are only checking if the sidadm user exists in any group, but the important part is to make sure sidadm is in the sapsys group. Please take a look at your test system, I would expect the sidadm user to be a member of sapsys, and this is what we want to check for.

For me that is a different check.
One check would be if the permissions (group membership) is as expected (see also haclient group). An other
check (this here) is to check only, if the users are defined locally. Maybe we need a layer to combine such check results to a combined result like "users are dined as expected".

examples/generic-azure.yaml
aleksei-burlakov: It's ok, that the user name is displayed, it will be ignored.

examples/generic-azure.yaml
aleksei-burlakov: Good catch, thanks!

aleksei-burlakov:
> May I also add my thoughts?
> 
> 1. Please do NOT assume there is only one single SID in the cluster. We allow multiple in different resources.
> 2. Testing /etc/passwd is not sufficient because the entry could be replaced by a remote definition.
>    I do nt know if pam gives us a change to calculate, if a user resolution is finally made locally or using remotes.

The first point makes sense, @stefanotorresi, @lee-martin  what do you think about this?
About the second point, this check checks explicitelly if the user is set up locally as we recommend in https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-sr-guide-PerfOpt-15/ in section 2, "Technical users and groups, such as <sid>adm are defined locally in the Linux system."

fmherschel:
> About the second point, this check checks explicitelly if the user is set up locally as we recommend in https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-sr-guide-PerfOpt-15/ in section 2, "Technical users and groups, such as adm are defined locally in the Linux system."

No? We do not say in the SBP that cecking /etc/passwd is sufficient. We only say that you need to make sure that the user, group and host names must be resolved locally - or do I have overseen something?

examples/generic-azure.yaml
fmherschel: At least with "grep -q" you could easy remove the <sid>adm from the output which would be more clean from my point of view.

fmherschel:
> > 1. Please do NOT assume there is only one single SID in the cluster. We allow multiple in different resources.
> The first point makes sense, @stefanotorresi, @lee-martin what do you think about this?

If you like I could help you with the coding to get all SID=xxx out of the cluster.

aleksei-burlakov:
> > About the second point, this check checks explicitelly if the user is set up locally as we recommend in https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-sr-guide-PerfOpt-15/ in section 2, "Technical users and groups, such as adm are defined locally in the Linux system."
> 
> No? We do not say in the SBP that cecking /etc/passwd is sufficient. We only say that you need to make sure that the user, group and host names must be resolved locally - or do I have overseen something?

This is correct, so I assumed that local=/etc/passwd

aleksei-burlakov:
> > > 1. Please do NOT assume there is only one single SID in the cluster. We allow multiple in different resources.
> > >    The first point makes sense, @stefanotorresi, @lee-martin what do you think about this?
> 
> If you like I could help you with the coding to get all SID=xxx out of the cluster.

This will be helpfull. And it would also be helpfull if you say how to check that the sap_suse_cluster_connector was used?

examples/generic-azure.yaml
aleksei-burlakov: > We also need to check other ways to "disable" users. (eg first character disables the user).

Could you please give more details, what do you mean by disabling a user by a first character?

fmherschel:
> > > About the second point, this check checks explicitelly if the user is set up locally as we recommend in https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-sr-guide-PerfOpt-15/ in section 2, "Technical users and groups, such as adm are defined locally in the Linux system."
> > 
> > 
> > No? We do not say in the SBP that cecking /etc/passwd is sufficient. We only say that you need to make sure that the user, group and host names must be resolved locally - or do I have overseen something?
> 
> This is correct, so I assumed that local=/etc/passwd and AD=remote

 I did some research and discussions with colleagues.
 I would say (at the moment) taht we at least also need to check in nsswitch.conf that the entry for passwd and group are set to either "compat" for "files". If it is set to an other source the entries in /etc/passwd does not play any role.

fmherschel:
> > > > 1. Please do NOT assume there is only one single SID in the cluster. We allow multiple in different resources.
> > > >    The first point makes sense, @stefanotorresi, @lee-martin what do you think about this?
> > 
> > 
> > If you like I could help you with the coding to get all SID=xxx out of the cluster.
> 
> This will be helpfull. 

Please give me some time ...

> And it would also be helpfull if you say how to check that the sap_suse_cluster_connector was used?

For SAPInstance resources please assume you have sap_suse_cluster_connector integration, so for those SIDs you can check the haclient group membership. For SAPHana, SAPHanaController and SAPHanaTopology we do not need to check this.

For SAPHana* we could (in the future) add checks for the needed sudoers entries (sudo -U xxx -l).

fmherschel:
To get a list of SIDs defined in SAPHanaController primitives: (in format value="SID1" value="SID2" ...)
cibadmin -Q | xmllint -xpath '//primitive[@type="SAPHanaController"]//nvpair[@name="SID"]/@value' - ;
Same for SAPHana, SAPHanaTopology and SAPInstance by just replaicing SAPHanaController by the other strings.

With cibadmin -Q | xmllint -xpath '//primitive//nvpair[@name="SID"]/@value' - ; echo
you get any value="SID" (nvpair) of any primitive resource in the cluster which might also hit dummy resources or other
resources which uses a SID=xxx parameter.

examples/generic-azure.yaml
```diff
@@ -673,3 +673,39 @@ groups:
           ## References
           - section 9.1.2 https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-sr-guide-PerfOpt-15/
         scored: true
+      - id: 1.5.3
+        description: "<sid>adm user and its <sid> group exist and enabled"
+        audit: |
+          # Get SIDs from cibadmin
+          IFS=' ' read -r -a sid_list_raw <<< $(sudo cibadmin -Q | xmllint -xpath '//primitive//nvpair[@name="SID"]/@value' -)
```
fmherschel, line 680: Looks good for me! I like that change. It addresses both the multi-sid and the nsswitch problematic.
Maybe we need to get more allowed values in nsswitch.conf but for now I would say we are fine.

examples/generic-azure.yaml
```diff
@@ -673,3 +673,39 @@ groups:
...
+          # Check the /etc/nsswitch.conf
+          if ! cat /etc/nsswitch.conf | grep "passwd:" | grep -q "compat\|files"; then echo failed=true; fi
+          if ! cat /etc/nsswitch.conf | grep "files:" | grep -q "compat\|files"; then echo failed=true; fi
+
+          # Check the /etc/passwd and /etc/group
+          for sid in ${sid_list_sorted_and_unique}
+          do
+            if ! cat /etc/passwd | cut -d: -f1 | grep -q ${sid}adm; then echo failed=true; fi
+            if cat /etc/passwd | grep -e "^${sid}adm" | cut -d: -f7 | grep -q "nologin"; then echo failed=true; fi
+            if ! cat /etc/group | cut -d: -f1,4 | grep -q "sapsys:${sid}adm"; then echo failed=true; fi
```
angelabriel, line 694: @aleksei-burlakov 
sid is 'prd'
entry in /etc/group looks like 'sapsys:x:79:briel,prdadm,ha1adm'
if ! cat /etc/group | cut -d: -f1,4 | grep -q "sapsys:${sid}adm"; then echo failed=true; fi
returns 'failed=true'
It's not so unusual that more than one user is added to a group. And if we have more than one SID defined, we may have more sidadm user added to the group sapsys.

aleksei-burlakov: Thank you, good catch!

examples/generic-azure.yaml
aleksei-burlakov: Thank you!)

examples/generic-azure.yaml
fmherschel: Just as a remark. AWK can combine things like cutting into fields and comparing multiple fields very well:

awk -F: 'BEGIN { found=1 } (($1 == "sapsys") && ($2 ~ user)) { found=0 } END { exit found }' user=hugadm

Just 2 ct "hugadm" is just the <sid>adm example and awk can also open the file, so we can use 1 command instead
of 3 to 4 using pipes.

aleksei-burlakov:
> @aleksei-burlakov in addition to @lee-martin 2 points from my side:
> 
> 1. Even that we do not official support the use of Active Directory, many customers - especially in cloud environments - simply use an AD integration instead of local defined users. And this often cause a lot of trouble.
>    It would be really helpful, if we can check this situation and give a big warning, that this is unsupported (at the moment).
>    So not only report 'no local users defined', but additional something like 'use of AD not supported'

@angelabriel @fmherschel I have added a check 1.5.4 that explicitly checks that the adm user is not available from an active directory.

fmherschel:
> > @aleksei-burlakov in addition to @lee-martin 2 points from my side:
> > 
> > 1. Even that we do not official support the use of Active Directory, many customers - especially in cloud environments - simply use an AD integration instead of local defined users. And this often cause a lot of trouble.
> >    It would be really helpful, if we can check this situation and give a big warning, that this is unsupported (at the moment).
> >    So not only report 'no local users defined', but additional something like 'use of AD not supported'

+1 

> 
> @angelabriel @fmherschel I have added a check 1.5.4 that explicitly checks that the adm user is not available from an active directory.

+1

aleksei-burlakov:
Could somebody please approve the PR?

stefanotorresi:
I'm afraid this work is lost, and this check needs to be completely re-implemented in Ansible now.

PR #120, opened 2021-06-30
Single cluster template revamp

fabriziosestito:
This PR implements the single cluster page template according to the mock-up.

It also adds structs in the web cluster module that are needed to represent the KV data in a form that can be consumed by the frontend.

NOTE: Pacemaker status is still missing.


web/clusters.go
```diff
@@ -11,6 +12,182 @@ import (
...
+
+func (node *Node) Role() string {
+	if r, ok := node.Attributes["hana_prd_roles"]; ok {
+		role := r[strings.LastIndex(r, ":")+1:]
+		return strings.Title(role)
+	}
+	return "-"
+}
+
+func (node *Node) HealthState() string {
```
arbulu89, line 57: What do you parse here?
Assuming something like `1:P:master1::worker:`, the first value? What does this number mean?

web/clusters.go
```diff
@@ -11,6 +12,182 @@ import (
...
+	var nodes Nodes
+
+	for _, n := range c.Crmmon.NodeAttributes.Nodes {
+		node := &Node{Name: n.Name, Attributes: make(map[string]string)}
+
+		for _, a := range n.Attributes {
+			node.Attributes[a.Name] = a.Value
+		}
+
+		for _, r := range c.Crmmon.Resources {
```
arbulu89, line 89: I think we need to loop through the `Groups` and `Clones` too. Individual resources are just one part of the resources.

web/clusters.go
```diff
@@ -11,6 +12,182 @@ import (
...
+			if r.Node.Name == n.Name {
+				resource := &Resource{
+					Id:   r.Id,
+					Type: r.Agent,
+					Role: r.Role,
+				}
+
+				for _, p := range c.Cib.Configuration.Resources.Primitives {
+					if r.Agent == "ocf::heartbeat:IPaddr2" && r.Id == p.Id {
```
arbulu89, line 98: I think AWS doesn't have this resource. It might be fine by now, but it will need to have some improvement in the future for sure

web/clusters.go
```diff
@@ -11,6 +12,182 @@ import (
...
+	}
+
+	return nodesBySite
+}
+
+func (nodes Nodes) CriticalCount() int {
+	var critical int
+
+	for _, n := range nodes {
+		if n.Health == "failed" || n.Health == "maintenance" || n.Health == "" {
```
arbulu89, line 171: I wouldn't count the nodes on maintenance as critical. And I don't even know if we have the `failed` state. Is this coming from `consul` ?

arbulu89 (commented):
Hi @fabriziosestito ,
I have done an initial review. I think there are many things to discuss about yet and some work pending, specially in `NewNodes`. We will need to have conversations with the cluster experts.

Besides this, unfortunately, this code would only work with a HANA cluster, and we will have other types too, so I don't know now how we should proceed, but it is not a simple thing

web/clusters.go
fabriziosestito: This should be `critical` not `failed`, I think it's a typo from my side.

About `maintenance`, I assumed it was a critical state by interpreting the if-then-else logic in the other templates, I'll remove the condition.

web/clusters.go
fabriziosestito: The first number is the `SAPHanaSR health state`



4 | SAP HANA database is up and OK. The cluster does interpret this as a correctly running database.
-- | --
3 | SAP HANA database is up and in status info. The cluster does interpret this as a correctly running database.
2 | SAP HANA database is up and in status warning. The cluster does interpret this as a correctly running database.
1 | SAP HANA database is down. If the database should be up and is not down by intention, this could trigger a takeover.
0 | Internal Script Error – to be ignored.

fabriziosestito:
I've re-added the old template and renamed it to "generic".
Now  we render the new page only if we detect that the cluster type is Hana scale-up/scale-out by (by looking at the resources), and we fallback to the old page in case we don't.

web/clusters.go
```diff
@@ -11,6 +12,222 @@ import (
...
+	Type      string
+	Role      string
+	Status    string
+	FailCount int
+}
+
+type Nodes []*Node
+
+const (
+	ClusterTypeScaleUp  = "scale-up"
```
arbulu89, line 35: Maybe better `hana-scale-up` to avoid confussions

web/clusters.go
```diff
@@ -11,6 +12,222 @@ import (
...
+				hasSapHanaTopology = true
+			case "ocf::suse:SAPHana":
+				hasSAPHana = true
+			case "ocf::suse:SAPHanaController":
+				hasSAPHanaController = true
+			}
+		}
+	}
+
+	if hasSapHanaTopology && hasSAPHana {
```
arbulu89, line 56: This looks other `switch/case` implementation to me

web/clusters.go
```diff
@@ -11,6 +12,222 @@ import (
...
+	Status    string
+	FailCount int
+}
+
+type Nodes []*Node
+
+const (
+	ClusterTypeScaleUp  = "scale-up"
+	ClusterTypeScaleOut = "scale-out"
+	ClusterTypeUnkown   = "unknown"
```
arbulu89, line 37: Typo in `ClusterTypeUnkown`

web/clusters.go
```diff
@@ -41,16 +258,27 @@ func NewClusterHandler(client consul.Client) gin.HandlerFunc {
...
+				"Cluster": cluster,
+				"Hosts":   hosts,
+			})
+			return
+		}
+
 		c.HTML(http.StatusOK, "cluster.html.tmpl", gin.H{
-			"Cluster": cluster,
-			"Hosts":   hosts,
+			"Cluster":          cluster,
```
arbulu89, line 278: I would rename the template to `cluster_hana.html.tmpl`

arbulu89 (changes requested):
Hi @fabriziosestito ,

Good improvements. I'm still concerned about the naming of all of these structs and function. Most of them are hana specific, so I would prefer to have something related to it. Eventually, depending on the cluster type, we will need to use different function, so this looks like something to be implemented by interfaces.

I don't know how you want to proceed, maybe we could open a follow up ticket and work on that later, to have this change in the next release, but this still needs some work

web/clusters.go
```diff
@@ -11,6 +12,239 @@ import (
...
+			}
+			stoppedResources = append(stoppedResources, resource)
+		}
+	}
+
+	return stoppedResources
+}
+
+func (node *Node) Role() string {
+	if r, ok := node.Attributes["hana_prd_roles"]; ok {
```
arbulu89, line 82: Maybe as a last request. Could we put here, as docstring, the meaining of this field and contents? We will forget otherwise hehe

web/clusters.go
```diff
@@ -11,6 +12,239 @@ import (
...
+	}
+
+	return nodesBySite
+}
+
+func (nodes Nodes) CriticalCount() int {
+	var critical int
+
+	for _, n := range nodes {
+		if n.Health == "critical" {
```
arbulu89, line 216: I don't know if we could use the `consulApi.HealthPassing` and kind of constants

web/clusters.go
```diff
@@ -41,16 +275,35 @@ func NewClusterHandler(client consul.Client) gin.HandlerFunc {
...
+		c.HTML(http.StatusOK, "cluster_hana.html.tmpl", gin.H{
+			"Cluster":          cluster,
+			"Nodes":            nodes,
+			"StoppedResources": stoppedResources(cluster),
+			"ClusterType":      clusterType,
+			"HealthContainer": &HealthContainer{
+				CriticalCount: nodes.CriticalCount(),
+				WarningCount:  nodes.WarningCount(),
+				PassingCount:  nodes.PassingCount(),
+				Layout:        "horizontal",
```
arbulu89, line 305: Do we need the `horizontal` layout here? I thought that we wanted the vertical. If this works there is something wrong with the styles XD

arbulu89 (commented):
@fabriziosestito We are almost there!!! I just added some comments that create me doubts

web/clusters.go
fabriziosestito: yes my bad, my code-reloader didn't get the CSS style change so it was actually vertical

web/clusters.go
fabriziosestito: yes! also this should be HANARole I think

arbulu89 (dismissed):
@fabriziosestito Great! Green light from me

web/templates/cluster_hana.html.tmpl
```diff
@@ -0,0 +1,115 @@
...
+        <div class="col-xl-12">
+            {{ template "sites" .Nodes.GroupBySite }}
+        </div>
+    </div>
+    <hr>
+
+    <h3>SBD/Stonith</h3>
+    {{- if .Cluster.IsFencingSBD }}
+        {{ template "sbd" .Cluster.SBD }}
+        <hr/>
```
rtorrero, line 108: Not sure if the <hr> closure is needed, but in any case, is this the right place?

fabriziosestito: It took me a while to understand but then I realized github was rendering the html ahah

No, you are right it is not needed I will remove this

rtorrero:
LGTM @fabriziosestito great work :1st_place_medal:

web/templates/cluster_hana.html.tmpl
rtorrero: haha, fixed my comment by adding the `<hr>` as code

rtorrero (approved):

PR #121, opened 2021-06-30
Lock mockery dependency version

fabriziosestito:
This PR introduces a way to track build tools dependency versions, as explained by the official wiki: https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module

It also locks the mockery dependency, this mean that the command `go install github.com/vektra/mockery/v2` will always install the pinned version (in this case `v2.9.0`).

The Makefile message was changed accordingly.

Note that the v2.9.0 version of mockery shows 0.0.0dev, so the same will appear in the header of the generated mocks, this due some troubles they had with the shipped binary https://github.com/vektra/mockery/pull/390
It should be fixed in the next releases so I think we can just wait and update later.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -30,7 +30,7 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-node-
     - name: install-mockery
-      run: go get github.com/vektra/mockery/v2
+      run: go install github.com/vektra/mockery/v2
     - name: test
       run: make test
     - name: static analysis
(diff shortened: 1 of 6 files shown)


arbulu89 (approved):

rtorrero:
@fabriziosestito you can rebase now with main; sorry for the hassle caused by the broken test!

rtorrero:
LGTM

rtorrero (approved):

PR #122, opened 2021-06-30
Add: Check that HANA's autostart is disabled

aleksei-burlakov:


examples/hana-scale-up-perf-optimized-azure.yaml
```diff
@@ -450,6 +450,32 @@ groups:
...
           
+          ## References
+          - section 2 in https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-sr-guide-PerfOpt-15/
+        scored: true
+      - id: 2.2.10
+        description: "HANA's autostart is disabled"
+        audit: |
+          sid=$(sudo crm configure show | grep -m1 SID= | sed -e "s/.*SID=\(...\).*/\1/" | tr '[:lower:]' '[:upper:]')
+          inst=$(sudo crm configure show | grep -m1 InstanceNumber= | sed -e "s/.*InstanceNumber=\(..\).*/\1/")
+          cat /hana/shared/${sid}/profile/${sid}_HDB${inst}_$(hostname)
```
arbulu89, line 461: I'm not sure if using `/hana/shared` is a good idea. I think this folder is configurable. We could use `/usr/sap/${sid}/SYS/profile` instead, which is always the same

arbulu89 (commented):
@aleksei-burlakov Just a comment about the usage of `/hana/shared`. I'm not 100% sure if that folder is always named the same

examples/hana-scale-up-perf-optimized-azure.yaml
aleksei-burlakov: Thank you! This makes sense. Changed.

dirkmueller (approved):

PR #123, opened 2021-06-30
Separate tags in the host list view

rtorrero:
This PR changes the hosts list view "tags" column to be split in separate columns. The goal is to provide the user a better representation of what environment, landscape, and SID a host might belong to.
EDIT: Add screenshot
hosts-table

### DIFF
--- a/web/templates/blocks/hosts_table.html.tmpl
+++ b/web/templates/blocks/hosts_table.html.tmpl
@@ -6,28 +6,41 @@
                 <th scope='col'>Name</th>
                 <th scope='col'>Address</th>
                 <th scope='col'>Cluster</th>
-                <th scope='col'>Tags</th>
+                <th scope='col'>System</th>
+                <th scope='col'>Landscape</th>
+                <th scope='col'>Environment</th>
                 <th scope='col'>Status</th>
             </tr>
             </thead>
             <tbody>
             {{- range .Hosts }}
-                <tr class='clickable' onclick="window.location='/hosts/{{ .Name }}'">
-                    <td>{{ .Name }}</td>
+                <tr>
+                    <td>
+                        <a href='/hosts/{{ .Name }}'>
+                            {{ .Name }}
+                        </a>
+                    </td>
                     <td>{{ .Address }}</td>
                     <td>
-                        <a href="/clusters/{{ index .TrentoMeta "trento-ha-cluster-id" }}">{{ index .TrentoMeta "trento-ha-cluster" }}</a>
+                        <a href='/clusters/{{ index .TrentoMeta "trento-ha-cluster-id" }}'>
+                            {{ index .TrentoMeta "trento-ha-cluster" }}
+                        </a>
                     </td>
                     <td>
-                        {{- range $Key, $Value := .TrentoMeta }}
-                            {{- if eq $Key "trento-ha-cluster" }}
-                            {{- else if eq $Key "trento-ha-cluster-id" }}
-                            {{- else }}
-                                <span class='badge badge-pill badge-info'>{{ $Value }}</span>
-                            {{- end }}
-                        {{- end }}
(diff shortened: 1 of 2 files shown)


arbulu89:
@rtorrero Could you upload a picture of the end result?

web/templates/blocks/hosts_table.html.tmpl
```diff
@@ -18,14 +20,10 @@
...
-                    <td>
-                        {{- range $Key, $Value := .TrentoMeta }}
-                            {{- if eq $Key "trento-ha-cluster" }}
-                            {{- else }}
-                                <span class='badge badge-pill badge-info'>{{ $Value }}</span>
-                            {{- end }}
-                        {{- end }}
-                    </td>
+
+                    <td>{{ .TrentoMetaKey "trento-sap-environment" }}</td>
```
arbulu89, line 24: I don't know if we need this new `.TrentoMetaKey` method. In other parts of the code we just use `index .TrentoMeta "trento-ha-cluster"` to get the value (check the cluster entry above). `.TrentoMeta` already returns a map with all of the metadata entries, and you only need to use the `index` method.
In any case, use the same option in all of them

web/templates/blocks/hosts_table.html.tmpl
```diff
@@ -6,7 +6,9 @@
                 <th scope='col'>Name</th>
                 <th scope='col'>Address</th>
                 <th scope='col'>Cluster</th>
-                <th scope='col'>Tags</th>
+                <th scope='col'>Environment</th>
```
arbulu89, line 9: I'm not sure about the order. In the sidebar, we have `Systems`, `Landscapes` and `Environments` top to down, maybe we should follow the same order here, from left to right

web/templates/blocks/hosts_table.html.tmpl
```diff
@@ -18,14 +20,10 @@
...
-                        {{- range $Key, $Value := .TrentoMeta }}
-                            {{- if eq $Key "trento-ha-cluster" }}
-                            {{- else }}
-                                <span class='badge badge-pill badge-info'>{{ $Value }}</span>
-                            {{- end }}
-                        {{- end }}
-                    </td>
+
+                    <td>{{ .TrentoMetaKey "trento-sap-environment" }}</td>
+                    <td>{{ .TrentoMetaKey "trento-sap-landscape" }}</td>
```
arbulu89, line 25: I think it would be good to put urls to the specific entries as we do in the `cluster` column, so you can move around easily

arbulu89 (changes requested):
Hi @rtorrero ,
Added some comments from my side. Add a picture of the view if you can too

rtorrero:
@arbulu89 thanks for the input! It should be ready now!

Attaching new screenshot:
updated-host-list

arbulu89 (approved):
Thanks @rtorrero 
Being picky, maybe we should re-order the filters on the top to, from `Systems` to `Environments`.
Besides this, I'm a bit worried of the `row` click change, that is still used in many of other rows (the rest of the tables).
I guess we will change them all, but honestly, I would wait until having real mockups feedback from UX team

rtorrero:
@arbulu89 thanks for the input! It should be ready now!

Attaching new screenshot:


> Thanks @rtorrero
> Being picky, maybe we should re-order the filters on the top to, from `Systems` to `Environments`.
> Besides this, I'm a bit worried of the `row` click change, that is still used in many of other rows (the rest of the tables).
> I guess we will change them all, but honestly, I would wait until having real mockups feedback from UX team
I'll take care of the filters reordering the next PR I send for this page so that we don't stall this PR too long if that's ok

Thanks!

fabriziosestito (approved):

stefanotorresi:
Hey @arbulu89 

> Besides this, I'm a bit worried of the `row` click change, that is still used in many of other rows (the rest of the tables).
> I guess we will change them all, but honestly, I would wait until having real mockups feedback from UX team

During our last UX session we established that we should have clickable rows, as most of them contain clickable links inside.

PR #126, opened 2021-07-02
Add cloud metadata details - Azure

arbulu89:
Add Azure instances metadata details. This implies a new Cloud discovery mechanism

Cloud instance metadata is stored in the KV storage under `trento/v0/hosts/$host/cloud`.
Within this entry, we have the `provider` string to define which provider belongs the instance, and the `metadata` entry will all the data.

The azure metadata is retreived using the internal RESP API mechanism: https://docs.microsoft.com/en-us/azure/virtual-machines/windows/instance-metadata-service?tabs=linux#instance-metadata

**If the VM is not recognized as an already implemented cloud or bare-metal, the cloud entry will be just empty)

I have only displayed some few options (all the metadata is stored just in case).

Some pics: 
image
image
image

Trento:
image
(I will adapt the table together with: https://github.com/trento-project/trento/pull/123)
image
As commented, only some few params are displayed


arbulu89:
Updated host list after rebasing `main`:
image

internal/cloud/azure.go
```diff
@@ -0,0 +1,209 @@
...
+	azureApiAddress = "169.254.169.254"
+	azurePortalUrl  = "https://portal.azure.com/#@SUSERDBillingsuse.onmicrosoft.com/resource"
+)
+
+type AzureMetadata struct {
+	Compute Compute `json:"compute,omitempty" mapstructure:"compute,omitempty"`
+	Network Network `json:"network,omitempty" mapstructure:"network,omitempty"`
+}
+
+type Compute struct {
```
rtorrero, line 29: This `Compute` struct in the `cloud` package feels to be specific to azure, right? How will we handle this when multiple cloud providers are implemented? should we take these out of `azure.go` maybe into a `cloud.go` file and make them more generic or will each cloud provider implementation include their own structs?

rtorrero:
Great work @arbulu89! appart from the comment above, would it make sense to adjust the templates to show a generic name for unsupported cloud providers?

fabriziosestito (approved):
LGTM!

internal/cloud/azure.go
arbulu89: Yes, this `compute` is Azure specific, and it is on purpose.
Later on the `cloud.go` integrates this in its `metadata`, which is an `interface`, so it will take anything you give. This time is azure data, but in other providers will be other information.
So it should be fine

arbulu89: And yes, each cloud provider will have their own structs depending on how the structure their metadata information

PR #127, opened 2021-07-05
Add: Output for ha_cluster_discovery and sap_system_discovery

aleksei-burlakov:
This is an init PR for the ticket CFSA-166.

Quote from CFSA-166: 

"""

We currently do have the Discovery health check loops report any output when they tick.

We should always report some output from all the health checks, both in case of success and failure.

It might also be worth re-thinking their error handling: for example, if the Agent is executed in a host that is not part of any SAP System, that discovery loop will still appear as green, while this is probably either a configuration mistake or an Agent malfunction.

Bear in mind that the output size is limited to 4KB, so we can't just dump extensive data there. Thus, this rework should go hand in hand with better logging.

Acceptance Criteria
The discovery loops have to report some human readable output to the Consul API, both in case of success and failure.
In case such health checks fail, a full trace of the occurred errors must be reported in the Agent logs.
A discovery loop health check failure must be reproduced and demonstrated live.

"""

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -241,18 +241,19 @@ func (a *Agent) startCheckTicker() {
 func (a *Agent) startDiscoverTicker() {
 	tick := func() {
 		for _, d := range a.discoveries {
-			var status string
+			var status, result string
 			var err error
 
-			err = d.Discover()
+			result, err = d.Discover()
 			if err != nil {
 				log.Printf("Error while running discovery '%s': %s", d.GetId(), err)
+				result = err.Error()
 				status = consulApi.HealthWarning
 			} else {
 				status = consulApi.HealthPassing
 			}
 
-			err = a.consul.Agent().UpdateTTL(d.GetId(), "", status)
+			err = a.consul.Agent().UpdateTTL(d.GetId(), result, status)
 			if err != nil {
 				log.Println("An error occurred while trying to update TTL with Consul:", err)
 			}
--- a/agent/discovery/cloud.go
+++ b/agent/discovery/cloud.go
@@ -1,6 +1,8 @@
 package discovery
 
 import (
+	"fmt"
+
 	"github.com/trento-project/trento/internal/cloud"
 	"github.com/trento-project/trento/internal/consul"
 	"github.com/trento-project/trento/internal/hosts"
@@ -24,23 +26,23 @@ func (d CloudDiscovery) GetId() string {
 	return d.id
 }
 
-func (d CloudDiscovery) Discover() error {
+func (d CloudDiscovery) Discover() (string, error) {
 	cloudData, err := cloud.NewCloudInstance()
 	if err != nil {
-		return err
+		return "", err
 	}
 
 	err = cloudData.Store(d.discovery.client)
 	if err != nil {
-		return err
+		return "", err
 	}
 
 	err = storeCloudMetadata(d.discovery.client, cloudData.Provider)
 	if err != nil {
-		return err
+		return "", err
 	}
 
-	return nil
+	return fmt.Sprintf("Cloud provider %s discovered", cloudData.Provider), nil
 }
 
 func storeCloudMetadata(client consul.Client, cloudProvider string) error {
--- a/agent/discovery/cluster.go
+++ b/agent/discovery/cluster.go
@@ -1,6 +1,8 @@
 package discovery
 
 import (
+	"fmt"
+
 	"github.com/trento-project/trento/internal/cluster"
 	"github.com/trento-project/trento/internal/consul"
 	"github.com/trento-project/trento/internal/hosts"
@@ -27,26 +29,26 @@ func (c ClusterDiscovery) GetId() string {
 }
 
 // Execute one iteration of a discovery and store the result in the Consul KVStore.
-func (d ClusterDiscovery) Discover() error {
+func (d ClusterDiscovery) Discover() (string, error) {
 	cluster, err := cluster.NewCluster()
 
 	if err != nil {
-		return err
+		return "", err
 	}
 
 	d.Cluster = cluster
 
 	err = d.Cluster.Store(d.discovery.client)
 	if err != nil {
-		return err
+		return "", err
 	}
 
 	err = storeClusterMetadata(d.discovery.client, cluster.Name, cluster.Id)
 	if err != nil {
-		return err
+		return "", err
 	}
 
-	return nil
+	return fmt.Sprintf("Cluster with name: %s successfully discovered", cluster.Name), nil
 }
 
 func storeClusterMetadata(client consul.Client, clusterName string, clusterId string) error {
--- a/agent/discovery/discovery.go
+++ b/agent/discovery/discovery.go
@@ -10,7 +10,7 @@ type Discovery interface {
 	// Returns an arbitrary unique string identifier of the discovery, so that we can associate it to a Consul check ID
 	GetId() string
 	// Execute the discovery mechanism
-	Discover() error
+	Discover() (string, error)
 }
 
 type BaseDiscovery struct {
@@ -24,9 +24,9 @@ func (d BaseDiscovery) GetId() string {
 }
 
 // Execute one iteration of a discovery and store the result in the Consul KVStore.
-func (d BaseDiscovery) Discover() error {
+func (d BaseDiscovery) Discover() (string, error) {
 	d.host, _ = os.Hostname()
-	return nil
+	return "Basic discovery example", nil
 }
 
 // Return a Host Discover instance
--- a/agent/discovery/sapsystem.go
+++ b/agent/discovery/sapsystem.go
@@ -1,6 +1,8 @@
 package discovery
 
 import (
+	"fmt"
+
 	"github.com/trento-project/trento/internal/consul"
 	"github.com/trento-project/trento/internal/environments"
 	"github.com/trento-project/trento/internal/hosts"
@@ -26,28 +28,33 @@ func (d SAPSystemsDiscovery) GetId() string {
 	return d.id
 }
 
-func (d SAPSystemsDiscovery) Discover() error {
+func (d SAPSystemsDiscovery) Discover() (string, error) {
 	systems, err := sapsystem.NewSAPSystemsList()
 
 	if err != nil {
-		return err
+		return "", err
 	}
 
 	d.SAPSystems = systems
 	for _, s := range d.SAPSystems {
 		err := s.Store(d.discovery.client)
 		if err != nil {
-			return err
+			return "", err
 		}
 	}
 
 	// Store SAP System, Landscape and Environment names on hosts metadata
 	err = storeSAPSystemTags(d.discovery.client, d.SAPSystems)
 	if err != nil {
-		return err
+		return "", err
 	}
 
-	return nil
+	sysNames := systems.GetSIDsString()
+	if sysNames != "" {
+		return fmt.Sprintf("SAP system(s) with ID: %s discovered", sysNames), nil
+	}
+	output := "No SAP systems were found (you possibly need to run the trento agent with root privileges)"
+	return output, nil
 }
 
 func storeSAPSystemTags(client consul.Client, systems sapsystem.SAPSystemsList) error {


aleksei-burlakov:
@stefanotorresi is it going in the right direction? What should be added here?

stefanotorresi (changes requested):
Hey Aleksei, what I had in mind is a bit different.

Instead of using the log message as output, what I meant is that the actual `Discovery` interface should return both a `string` and and `error` (instead of just an `error`, as it is currently), and report a meaningful one-line message with some contextual information (for example, the number of systems newly discovered, or anything else).

To get an example of this, you can look at the check ticker:
https://github.com/trento-project/trento/blob/d49ebb555f7eb30154350b8c448388d597810c4c/agent/agent.go#L212-L227
As you can see there, we use the actual result of the checker as the Consul health check output.

What I would like to introduce is an actual result for the discovery loops as well, which means you need to slightly alter the Discovery interface and all its implementations.

I wouldn't just turn the existing logging message into the output, because that is meant to go in the console log, rather than the Consul Health Check metadata.

aleksei-burlakov:
> Hey Aleksei, what I had in mind is a bit different.
> 
> Instead of using the log message as output, what I meant is that the actual `Discovery` interface should return both a `string` and and `error` (instead of just an `error`, as it is currently), and report a meaningful one-line message with some contextual information (for example, the number of systems newly discovered, or anything else).
> 
> To get an example of this, you can look at the check ticker:
> https://github.com/trento-project/trento/blob/d49ebb555f7eb30154350b8c448388d597810c4c/agent/agent.go#L212-L227
> 
> 
> As you can see there, we use the actual result of the checker as the Consul health check output.
> What I would like to introduce is an actual result for the discovery loops as well, which means you need to slightly alter the Discovery interface and all its implementations.
> 
> I wouldn't just turn the existing logging message into the output, because that is meant to go in the console log, rather than the Consul Health Check metadata.

@stefanotorresi I have amended the PR, the `Discover()` function besides an `error` returns a `string` now. I really lack imagination, what should be in the string, except the same that in the error, but maybe in other words. I assume this is not what is expected, so I left the strings empty. The ClusterDiscovery class calls the cibadmin to check if the cliuster is ok, and if it's not it says it in the error (either "error while executing cibadmin" or "could not parse cibadmin status from XML").

agent/discovery/cluster.go
```diff
@@ -27,26 +27,26 @@ func (c ClusterDiscovery) GetId() string {
 }
 
 // Execute one iteration of a discovery and store the result in the Consul KVStore.
-func (d ClusterDiscovery) Discover() error {
+func (d ClusterDiscovery) Discover() (string, error) {
 	cluster, err := cluster.NewCluster()
 
 	if err != nil {
-		return err
+		return "", err
```
arbulu89, line 34: I guess we could return here the error converted to a string. Have a look how you can do this as the errors might be wrapped one on top of each other.
And basically expand this to all the error cases

agent/discovery/cluster.go
```diff
@@ -27,26 +27,26 @@ func (c ClusterDiscovery) GetId() string {
...
 	}
 
 	err = storeClusterMetadata(d.discovery.client, cluster.Name, cluster.Id)
 	if err != nil {
-		return err
+		return "", err
 	}
 
-	return nil
+	return "", nil
```
arbulu89, line 49: Maybe we could return something like `Cluster with name: %s successfully discovered`

agent/discovery/discovery.go
```diff
@@ -24,9 +24,9 @@ func (d BaseDiscovery) GetId() string {
 }
 
 // Execute one iteration of a discovery and store the result in the Consul KVStore.
-func (d BaseDiscovery) Discover() error {
+func (d BaseDiscovery) Discover() (string, error) {
 	d.host, _ = os.Hostname()
-	return nil
+	return "", nil
```
arbulu89, line 29: We could put something like `Basic discovery example`

agent/discovery/sapsystem.go
```diff
@@ -26,28 +26,28 @@ func (d SAPSystemsDiscovery) GetId() string {
...
 		// Store SAP System, Landscape and Environment names on hosts metadata
 		err = storeSAPSystemTags(d.discovery.client, s)
 		if err != nil {
-			return err
+			return "", err
 		}
 	}
 
-	return nil
+	return "", nil
```
arbulu89, line 50: We could put maybe something like `SAP system(s) with ID %s discovered`, and put there all the SAP ids (you will need to pack them in some array in the loop maybe)

arbulu89 (changes requested):
@aleksei-burlakov Even though you think you don't have imagination, give a try hehe
Empty strings are useless, we need to put something. I have given some examples (take in mind that I took some seconds to make them, so you can for sure get better texts)

aleksei-burlakov:
> @aleksei-burlakov Even though you think you don't have imagination, give a try hehe
> Empty strings are useless, we need to put something. I have given some examples (take in mind that I took some seconds to make them, so you can for sure get better texts)

@arbulu89 thank you for reviewing. The important point is that I'm lacking imagination not to write something, but to write the text that should be there (#Occam's razor, #dry code). The first (string) variable looks excessive to me, because it repeats the second variable the err. In my understanding, we need the first variable to return more than 4 thousand characters of text.

Well, anyway I will write something not to return empty strings.

aleksei-burlakov:
I found in https://golang.org/doc/tutorial/handle-errors that it's a common practice in go-lang to return a message alongside the error.

arbulu89:
> I found in https://golang.org/doc/tutorial/handle-errors that it's a common practice in go-lang to return a message alongside the error.

Yes, it is. We do it all the time in the project

agent/discovery/cluster.go
aleksei-burlakov: Done! Also to the other cases.

agent/discovery/cluster.go
aleksei-burlakov: Changed!

agent/discovery/discovery.go
aleksei-burlakov: Changed!

agent/discovery/sapsystem.go
aleksei-burlakov: Changed!

agent/discovery/sapsystem.go
```diff
@@ -26,28 +29,39 @@ func (d SAPSystemsDiscovery) GetId() string {
 	return d.id
 }
 
-func (d SAPSystemsDiscovery) Discover() error {
+func (d SAPSystemsDiscovery) Discover() (string, error) {
```
arbulu89, line 32: The changes in this `Discover` will conflict with changes coming from: https://github.com/trento-project/trento/pull/160
I think we should take the change from that other PR, as the sid concatenation looks better

agent/discovery/sapsystem.go
```diff
@@ -26,28 +29,39 @@ func (d SAPSystemsDiscovery) GetId() string {
...
 		err = storeSAPSystemTags(d.discovery.client, s)
 		if err != nil {
-			return err
+			return fmt.Errorf("%v", err).Error(), err
 		}
 	}
 
-	return nil
+	if sids != "" {
+		return fmt.Sprintf("Cluster%s with name%s: %s successfully discovered", plural, plural, sids), nil
```
arbulu89, line 61: This message doesn't make much sense as we are not referring to clusters here, we are talking about `SAP systems`

agent/discovery/sapsystem.go
```diff
@@ -26,28 +29,39 @@ func (d SAPSystemsDiscovery) GetId() string {
...
+	}
+	output := "No clusters were found, pelase run the trento agent with root privileges"
```
arbulu89, line 63: The same here, we should refer to `SAP systems`. In fact, not finding any SAP system is not an error, so we should still return a `nil` value in the error parameter at the end

arbulu89 (changes requested):
@aleksei-burlakov Some comments.
Besides them, I think that you forget to apply these changes to the cloud discovery. I don't know if you need to even rebase this PR to get those changes, but as it implements the same interface, if we don't change it, it will fail

agent/discovery/sapsystem.go
aleksei-burlakov: Rebased!

agent/discovery/sapsystem.go
aleksei-burlakov: Thanks! I've mistakenly copied the text from the cluster module. Fixed.

aleksei-burlakov (commented):
Changed!

aleksei-burlakov:
> @aleksei-burlakov Some comments.
> Besides them, I think that you forget to apply these changes to the cloud discovery. I don't know if you need to even rebase this PR to get those changes, but as it implements the same interface, if we don't change it, it will fail

I think it was previously compiling w/o cloud discovery somehow. Strange. I have added the message string return value to the CloudDiscovery::Discover() method

agent/discovery/cloud.go
```diff
@@ -24,23 +26,23 @@ func (d CloudDiscovery) GetId() string {
...
 	}
 
 	err = storeCloudMetadata(d.discovery.client, cloudData.Provider)
 	if err != nil {
-		return err
+		return fmt.Errorf("%v", err).Error(), err
 	}
 
-	return nil
+	return "Found " + cloudData.Provider + " cloud provider", nil
```
arbulu89, line 45: We should use the `fmt.Sprintf` options here to format the string

agent/discovery/sapsystem.go
```diff
@@ -26,28 +28,39 @@ func (d SAPSystemsDiscovery) GetId() string {
...
 	systems, err := sapsystem.NewSAPSystemsList()
 
 	if err != nil {
-		return err
+		return fmt.Errorf("%v", err).Error(), err
 	}
 
 	d.SAPSystems = systems
 	for _, s := range d.SAPSystems {
+		if sids != "" {
```
arbulu89, line 41: We already have the method to get this SID list.
https://github.com/trento-project/trento/blob/e70765b81566f41a1b0f2df86916e8f96a851bcd/internal/sapsystem/sapsystem.go#L103

You can simply is this at then end to create the output string.
I wouldn't be worried about the plural. You can simply put `SAP system(s)` maybe

agent/discovery/cloud.go
arbulu89: And to follow the same convention as the other discovery systems, I would rename to `fmt.Sprintf("Cloud provider %s discovered", cloudData.Provider)`

arbulu89 (changes requested):
Some changes requested @aleksei-burlakov

agent/discovery/cloud.go
aleksei-burlakov: Changed!

agent/discovery/sapsystem.go
aleksei-burlakov: Changed!

agent/discovery/cluster.go
arbulu89: Sorry, @aleksei-burlakov ,
You were right here. I don't know why I recommended to send something in the 1st variable in a error case...
It doesn't make sense. You can put back an empty string here

arbulu89: **Edit**. Ok, now I understand why I requested this...
We needed to return back a correct message because this is the value used to send the message later.
I will run a complete review now, but something smells strange now

agent/agent.go
```diff
@@ -241,18 +241,18 @@ func (a *Agent) startCheckTicker() {
...
+			result, err = d.Discover()
 			if err != nil {
 				log.Printf("Error while running discovery '%s': %s", d.GetId(), err)
 				status = consulApi.HealthWarning
 			} else {
 				status = consulApi.HealthPassing
 			}
 
-			err = a.consul.Agent().UpdateTTL(d.GetId(), "", status)
+			err = a.consul.Agent().UpdateTTL(d.GetId(), result, status)
```
arbulu89, line 255: This is the strange thing. If we always return `result` this might not be a correct message in case that the `Discover()` function returns an error.
I suggest the next:
- We store the sent result in `a.consul.Agent()...` depending on which if statement we enter. If we go in the `if` case (error), we will need to unpack the error to get a string. If we are in the `else` side, we directly send the result. You can use the `result` variable to store the error message in the `if` case, that's all. Something like

```
result, err = d.Discover()
if err != nil {
    log.Printf("Error while running discovery '%s': %s", d.GetId(), err)
    result = err.Error()
    status = consulApi.HealthWarning
} else {
...
```

agent/discovery/cloud.go
```diff
@@ -24,23 +26,23 @@ func (d CloudDiscovery) GetId() string {
 	return d.id
 }
 
-func (d CloudDiscovery) Discover() error {
+func (d CloudDiscovery) Discover() (string, error) {
 	cloudData, err := cloud.NewCloudInstance()
 	if err != nil {
-		return err
+		return fmt.Errorf("%v", err).Error(), err
```
arbulu89, line 32: You are right, we can remove the 1st parameter in the error cases after changing how we use the result in the main discovery loop

arbulu89 (changes requested):
Hey @aleksei-burlakov ,
We are almost there.
Sorry for the confusion I created with the errors...
I didn't check well the 1st time. Let's try to use properly the `return string, err`.
I have proposed a solution in one of the comments

arbulu89 (approved):
Thanks @aleksei-burlakov 
Green light!

agent/discovery/sapsystem.go
```diff
@@ -26,28 +28,35 @@ func (d SAPSystemsDiscovery) GetId() string {
...
 	systems, err := sapsystem.NewSAPSystemsList()
 
 	if err != nil {
-		return err
+		return "", err
 	}
 
 	d.SAPSystems = systems
 	for _, s := range d.SAPSystems {
+		sysNames = systems.GetSIDsString()
```
arbulu89, line 41: By the way, you should move the `systems.GetSIDsString` out of the loop. It will always do the same

aleksei-burlakov:
> By the way, you should move the systems.GetSIDsString out of the loop. It will always do the same

Good catch, changed!

arbulu89 (approved):

PR #129, opened 2021-07-05
Add templates sanity check

fabriziosestito:
Gin does not handle properly errors when they occur during the template rendering.
This because the `ResponseWriter` status is immutable see: https://github.com/gin-gonic/gin/blob/ee4de846a894e9049321e809d69f4343f62d2862/response_writer.go#L63 
so when `context.HTML` is called the Status is already set and it's not changed by the panic recovering to 500.

To fix this behavior, this PR introduces a `LayoutHTML` renderer which is returned instead of gin's `render.HTML` by our `LayoutRender.Instance` method.
The `Render` method simply tries to render the template on a buffer, and renders an error template/error status code in case something goes wrong.

Now the `TestSAPSystemHandler` fails as expected, I might need some help fixing this.

### DIFF
--- a/web/errors.go
+++ b/web/errors.go
@@ -1,6 +1,10 @@
 package web
 
-import "github.com/gin-gonic/gin"
+import (
+	"net/http"
+
+	"github.com/gin-gonic/gin"
+)
 
 func ErrorHandler(c *gin.Context) {
 	c.Next()
@@ -53,7 +57,15 @@ func (e *HttpError) Template() string {
 func NotFoundError(msg string) *HttpError {
 	return &HttpError{
 		msg,
-		404,
+		http.StatusNotFound,
 		"error404.html.tmpl",
+	}
+}
+
+func InternalServerError(msg string) *HttpError {
+	return &HttpError{
+		msg,
+		http.StatusInternalServerError,
+		"error.html.tmpl",
 	}
 }
--- a/web/layout.go
+++ b/web/layout.go
@@ -5,7 +5,10 @@ import (
 	"fmt"
 	"html/template"
 	"io/fs"
+	"net/http"
 	"path/filepath"
+
+	log "github.com/sirupsen/logrus"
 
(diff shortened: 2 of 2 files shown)


arbulu89 (approved):
Nice and clean! Fantastic.

Now we can do 2 things.
- We can fix the now reported error in this PR
- We merge with a broken test and send the fix in a new PR

Up to you Fabrizio

fabriziosestito:
> Nice and clean! Fantastic.
> 
> Now we can do 2 things.
> 
>     * We can fix the now reported error in this PR
> 
>     * We merge with a broken test and send the fix in a new PR
> 
> 
> Up to you Fabrizio

@arbulu89 I would tend towards merging this + opening an issue since I need some time on the single cluster page and it will be staling for a bit, as long as we are ok with a test failing

arbulu89:
> > Nice and clean! Fantastic.
> > Now we can do 2 things.
> > ```
> > * We can fix the now reported error in this PR
> > 
> > * We merge with a broken test and send the fix in a new PR
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Up to you Fabrizio
> 
> @arbulu89 I would tend towards merging this + opening an issue since I tried fixing the test yesterday but it was a bit trickier than i thought and I need some time on the single cluster page and it will be staling for a bit, as long as we are ok with a test failing.

Ok.
Unfortunately merging this PR will make other open PRs to fail, so we will need to do more rebasing work.
But, well, I guess it will be the best choice to avoid broken tests again

stefanotorresi (approved):

stefanotorresi:
good catch, Fabrizio! I would also suggest to merge this and fix the pre-existing failure in another PR.

fabriziosestito:
@stefanotorresi @arbulu89 we could comment the test on this one and fix it in another issue?

PR #130, opened 2021-07-07
Fix TestSAPSystemHandler test

arbulu89:
Fix TestSAPSystemHandler test.

In the future, we will need to think a better system for mocking all of these things...


fabriziosestito (approved):
LGTM

rtorrero (approved):
LGTM

PR #139, opened 2021-07-12
Implement new styles for the host list view

arbulu89:
New styles for the Host list view, following the cluster list mockup.

- `Updated at` doesn't come from the backend. It shows just the time when the page was refreshed
- The `Filter` and `Reset` buttons might go away if we implement the ajax mechanism to reload the table page
- The `health_container` partial should be reusable in other pages, just changing the `horizontal/vertical` display option (this comes now from the backend, I didn't find a better option)

image

### DIFF
--- a/web/hosts.go
+++ b/web/hosts.go
@@ -15,6 +15,28 @@ import (
 	"github.com/trento-project/trento/internal/sapsystem"
 )
 
+type HealthContainer struct {
+	Passing  int
+	Warning  int
+	Critical int
+	Layout   string
+}
+
+func NewHealthContainer(hostList hosts.HostList) *HealthContainer {
+	h := &HealthContainer{}
+	for _, host := range hostList {
+		switch host.Health() {
+		case "passing":
+			h.Passing += 1
+		case "warning":
+			h.Warning += 1
+		case "critical":
+			h.Critical += 1
+		}
+	}
+	return h
+}
+
 func NewHostListHandler(client consul.Client) gin.HandlerFunc {
 	return func(c *gin.Context) {
 		query := c.Request.URL.Query()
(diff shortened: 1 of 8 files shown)


web/hosts.go
```diff
@@ -15,6 +15,28 @@ import (
...
+	h := &HealthContainer{}
+	for _, host := range hostList {
+		switch host.Health() {
+		case "passing":
+			h.Passing += 1
+		case "warning":
+			h.Warning += 1
+		case "critical":
+			h.Critical += 1
+		}
```
fabriziosestito, line 35: I think we have `maintenance` status too, in the previous template logic it was squashed to the critical state

fabriziosestito (approved):

web/hosts.go
arbulu89: Yes, true. But we don't use this state at all. If we start using it sometime, we will need to include it

web/hosts.go
```diff
@@ -15,6 +15,28 @@ import (
...
 
+type HealthContainer struct {
+	Passing  int
+	Warning  int
+	Critical int
+	Layout   string
+}
+
+func NewHealthContainer(hostList hosts.HostList) *HealthContainer {
+	h := &HealthContainer{}
```
fabriziosestito, line 26: Maybe we could move the `HealthContainer` to a new file and rename this to `NewHostsHealthContainer`  or use an `[]interface` with `.Health()` as constructor parameter so that we can use the same logic in other pages as well.

arbulu89: Ok, if this helps for reusability I will move it to a new file and apply these changes

arbulu89: @fabriziosestito Changes done in: https://github.com/trento-project/trento/pull/139/commits/f5612982eb924c4c2efacaa903579291a527f7aa

Let's go with this easy approach by now.

If you agree, we can merge afterwards

fabriziosestito (approved):

PR #142, opened 2021-07-14
List clusters template revamp

fabriziosestito:
This PR updates the cluster list view with the filters, pagination and health component.
Added autocomplete behaviour to some of the the filter inputs, such as `name` and `sid`

Structs were added to abstract the clusters list.

### DIFF
--- a/web/clusters.go
+++ b/web/clusters.go
@@ -5,6 +5,8 @@ import (
 	"net/http"
 	"strings"
 
+	"github.com/trento-project/trento/internal"
+
 	consulApi "github.com/hashicorp/consul/api"
 
 	"github.com/gin-gonic/gin"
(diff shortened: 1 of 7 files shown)


web/clusters.go
```diff
@@ -261,16 +263,183 @@ func (nodes Nodes) PassingCount() int {
...
+		var sids []string
+
+		hl, ok := hosts[id]
+
+		if ok && len(hl) > 0 {
+			health = hl.Health()
+			set := make(map[string]struct{})
+
+			for _, h := range hl {
+				ss, err := h.GetSAPSystems()
```
arbulu89, line 292: I don't really think we should do this `GetSAPSystems`. We should get the SID from the cluster resources (it should be in some cluster node attribute).
In fact, I'm not sure if the cluster should have multiple sids. It's true that the cost-optimized manages 2 sids though

web/templates/blocks/health_icon.html.tmpl
```diff
@@ -0,0 +1,11 @@
+{{ define "health_icon" }}
```
arbulu89, line 1: I guess we should reuse this template in the hosts list too

web/templates/clusters.html.tmpl
```diff
@@ -1,33 +1,61 @@
...
+            $(document).ready(function () {
+                {{- range $Key, $Value := .AppliedFilters }}
+                $("#{{ $Key }}").selectpicker("val", {{ $Value }});
+                {{- end }}
+                $('#clean').click(function () {
+                    $('.selectpicker').selectpicker("deselectAll")
+                });
+            });
+        </script>
+        <select name="health" id="health" class="selectpicker" multiple
```
arbulu89, line 30: This piece is duplicated too

arbulu89 (commented):
@fabriziosestito Nice work!
I just found that there will be some duplicated code between the `hosts` view and the `clusters` view.
Besides that, I did a comment about the usage of this `GetSAPSystems`, which I'm not sure we should use.

Anyway, let's merge and maybe open follow up tickets

web/clusters.go
stefanotorresi: to follow up in #143

PR #149, opened 2021-07-15
Fix SID retrieval

### LINKED ISSUE #143: Avoid using `GetSAPSystems` in the Cluster List view model.
> I don't really think we should do this `GetSAPSystems`. We should get the SID from the cluster resources (it should be in some cluster node attribute).
> In fact, I'm not sure if the cluster should have multiple sids. It's true that the cost-optimized manages 2 sids though

_Originally posted by @arbulu89 in https://github.com/trento-project/trento/pull/142#discussion_r669723594_

fabriziosestito:
Resolves #143 

`SID` is retrieved from from `SAPHanaController` instead of `GetSAPSystems`.
Removes unused code, updates relevant tests.

### DIFF
--- a/web/clusters.go
+++ b/web/clusters.go
@@ -82,6 +82,20 @@ func stoppedResources(c *cluster.Cluster) []*Resource {
 	return stoppedResources
 }
 
+func getHanaSID(c *cluster.Cluster) string {
+	for _, r := range c.Cib.Configuration.Resources.Clones {
+		if r.Primitive.Type == "SAPHanaTopology" {
+			for _, a := range r.Primitive.InstanceAttributes {
+				if a.Name == "SID" {
+					return a.Value
+				}
+			}
+		}
+	}
+
+	return ""
+}
+
 // HANARole parses the hana_prd_roles string and returns the HANA Role
 // Possible values: master, slave
 // e.g. 4:P:master1:master:worker:master returns master (last element)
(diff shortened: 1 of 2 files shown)


web/clusters.go
```diff
@@ -82,6 +82,20 @@ func stoppedResources(c *cluster.Cluster) []*Resource {
 	return stoppedResources
 }
 
+func getHanaSID(c *cluster.Cluster) string {
```
arbulu89, line 85: This new function is not used anywhere

web/clusters.go
```diff
@@ -275,31 +289,19 @@ type ClustersRow struct {
...
-					continue
-				}
-
-				for _, s := range ss {
-					_, ok := set[s.SID]
-					if !ok {
-						set[s.SID] = struct{}{}
-						sids = append(sids, s.SID)
-
+		for _, r := range c.Cib.Configuration.Resources.Clones {
```
arbulu89, line 300: I don't understand why you re-implemented this code again. Does not `getHanaSID` fill this purpose?

arbulu89 (changes requested):
Hi @fabriziosestito ,
I see that the same code has been created twice here, something strange. Can you have a look?

web/clusters.go
fabriziosestito: sorry, something went wrong when squashing commits, fixed

arbulu89 (approved):
Thanks @fabriziosestito ,
I would find valuable to add some test for `getHanaSID`, but up to you.
When can merge when you want

arbulu89 (approved):
@fabriziosestito Thanks again! You write the code faster than I can review! :)
Good work

PR #150, opened 2021-07-16
Implement the 1st part of runner component

arbulu89:
First implementation of the `runner` component.

The runner is responsible of running the ansible check by now. It depends on ansible and [ARA](https://ara.recordsansible.org/).

**I will add documentation once we have some advanced discussion on the PR**.

In order to test the PR:
- `ARA` must be running in some host (I just had it running in the monitoring host). To use it:
  - Install ansible `pip install ansible`
  - Install ara: `pip install --user ansible "ara[server]"`
  - Run the ara migration `ara-manage migrate`
  - Start the ara server: `ara-manage runserver attachedIP:port`
  - Enable ssh connection to the other hosts (update `/etc/hosts` and run `ssh-copy-id` to all the checked hosts)
- Start the trento runner (in the monitoring host in my case too)
  - `./trento runner start --ara-server http://attachedIP:port --consul-addr attachedIP:consulPort -i 1`

Here the major features.

- New command action added. `runner`. Here the options:
```
# ./trento runner start -h
Starts the runner process. This process takes care of running the checks

Usage:
  trento runner start [flags]

Flags:
      --ansible-folder string   Folder where the ansible file structure will be created (default "/usr/etc/trento")
      --ara-server string       ARA server url (ex: http://localhost:8000) (default "http://127.0.0.1:8000")
      --consul-addr string      Consul host address (ex: localhost:8500) (default "127.0.0.1:8500")
  -h, --help                    help for start
  -i, --interval int            Interval in minutes to run the checks (default 5)

Global Flags:
      --config string      config file (default is $HOME/.trento.yaml)
      --log-level string   then minimum severity (error, warn, info, debug) of logs to output (default "info")

```

This `start` action does the next:
- Starts a loop to run ansible playbooks.
- Creates the ansible files in a temporary file structure coming from the embedded files. Ansible needs to have actual files to be executed, so we need to copy the embedded files somewhere
- Run consul-template to create the ansible inventory

**Disclaimer:**
- I know that the `consul-template` implementation is almost a copy from the one used in the `agent`. This is true, I just wanted to have code changes related to this component first. Once we merge it, we will work to have a unique implementation.
- The `ansible` files are just a dummy tasks to make some tests. They will be of course upgraded and change in the future


arbulu89:
PR changed to `runner` branch

fabriziosestito (approved):

runner/runner.go
```diff
@@ -0,0 +1,198 @@
...
+
+	err = ansibleRunner.LoadAraPlugins()
+	if err != nil {
+		return
+	}
+
+	ansibleRunner.SetAraServer(c.cfg.AraServer)
+
+	tick := func() {
+		if !ansibleRunner.isAraServerUp() {
```
arbulu89, line 170: Try to add the consul-template execution here, as `run-once`

runner/runner.go
```diff
@@ -0,0 +1,198 @@
...
+
+	return runner, nil
+}
+
+func DefaultConfig() (Config, error) {
+	return Config{
+		AraServer:              "http://127.0.0.1:8000",
+		ConsulAddr:             "127.0.0.1:8500",
+		Interval:               5 * time.Minute,
+		AnsibleFolder:          "/usr/etc/trento",
```
arbulu89, line 63: Move this to `/var/lib/trento` (check for temporary folders)

runner/runner.go
arbulu89: Changed in: https://github.com/trento-project/trento/pull/150/commits/6972c8eee957f196d80f0127d95fb4f3526bd79e

runner/runner.go
arbulu89: Changed in: https://github.com/trento-project/trento/pull/150/commits/6a4281fc34ca8eee7d4a811a1bf9f042d5fa1423

PR #151, opened 2021-07-16
Fix resource status switch/case

fabriziosestito:
Resource status switch/case was mistakenly nested inside the Virtual IP retrieval for loop in the clusters detail page cluster nodes constructor.|

This PR fixes it.


arbulu89 (approved):
@fabriziosestito As always, some test would help to avoid regressions later or

rtorrero (approved):
Thanks for adding the tests! LGTM

rtorrero (approved):

PR #152, opened 2021-07-19
Add: one-line installer

aleksei-burlakov:
That's the first iteration for **CFSA-168**. It works under assumption that https://api.opensuse.org/package/show/devel:sap:trento/trento repo is on the zypper list. When the trento is in the official GA of Update repository, it will work out of box, otherwise you would manually need to add the repo e.g. https://download.opensuse.org/repositories/devel:/sap:/trento/15.3/devel:sap:trento.repo

And here arises a question. Should adding the repository be included in the script, or we may assume the repository is always there?

The description of CFSA-168:
Create a script (or binary) so that the user can install trento in one line.

E.g, a curl with a bash script

We want to reduce to the bare minimum the parameters that the user will need to pass to the script, maybe by using an interactive approach but keeping the option to provide arguments via command line for an unattended installation, which can be also leveraged by the CI.

Acceptance Criteria

- Make Agent installation a simply one liner
- All content needs to be in the installation archive for offline use (air-gap support)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,191 @@
+#!/bin/bash
+# The script requires root permissions
+
+set -e
+
+if [ "$EUID" -ne 0 ]; then
+    echo "Please run as root"
+    exit
+fi
+
+function print_help() {
+    cat <<END
+This is a trento-agent installer. Trento is a web-based graphical user interface
+that can help you deploy, provision and operate infrastructure for SAP Applications
+
+Usage:
+
+  sudo ./install.sh --bind-ip <127.0.0.1> --server-ip <192.168.122.1>
+
+Arguments:
+  --bind-ip    The bind ip.
+  --server-ip  The server ip.
+  --name       The node name.
+  --help       Print this help.
+END
+}
+
+# Treat the "--help" specially
+# It neither requires a value
+# nor is compatible with other arguments
+case "$1" in
+   --help)
+        print_help
+        exit 0;;
+esac
+
+ARGUMENT_LIST=(
+    "bind-ip"
+    "server-ip"
+    "name"
+)
+
+# read arguments
+opts=$(getopt \
+    --longoptions "$(printf "%s:," "${ARGUMENT_LIST[@]}")" \
+    --name "$(basename "$0")" \
+    --options "" \
+    -- "$@"
+)
+
+eval set --$opts
+
+while [[ $# -gt 0 ]]; do
+    case "$1" in
+        --bind-ip)
+            BIND_IP=$2
+            shift 2
+            ;;
+
+        --server-ip)
+            SERVER_IP=$2
+            shift 2
+            ;;
+
+        --name)
+            NODE_NAME=$2
+            shift 2
+            ;;
+
+        *)
+            break
+            ;;
+    esac
+done
+
+if [ -z "$BIND_IP" ]; then
+    read -p "Please provide the bind IP: "   BIND_IP
+fi
+if [ -z "$SERVER_IP" ]; then
+    read -p "Please provide the server IP: " SERVER_IP
+fi
+if [ -z "$NODE_NAME" ]; then
+    NODE_NAME="$(hostname)"
+fi
+
+TRENTO_REPO_KEY=${REPO_KEY:-"https://download.opensuse.org/repositories/devel:/sap:/trento/15.3/repodata/repomd.xml.key"}
+TRENTO_REPO=${REPO:-"https://download.opensuse.org/repositories/devel:/sap:/trento/15.3/devel:sap:trento.repo"}
+
+KVSTORE_VERSION=1.9.6
+SRV_HOME=/srv/consul
+CONFIG_PATH="$SRV_HOME/consul.d"
+HCL_TEMPLATE='data_dir = "/srv/consul/data/"
+node_name = "@NODE_NAME@"
+log_level = "DEBUG"
+datacenter = "dc1"
+ui = true
+bind_addr = "@BIND_ADDR@"
+client_addr = "0.0.0.0"
+retry_join = ["@JOIN_ADDR@"]'
+
+SERVICE_FILE_NAME="consul.service"
+SERVICE_FILE='[Unit]
+Description="HashiCorp Consul - A service mesh solution"
+Documentation=https://www.consul.io/
+Requires=network-online.target
+After=network-online.target
+ConditionFileNotEmpty=/srv/consul/consul.d/consul.hcl
+
+[Service]
+ExecStart=/srv/consul/consul agent -config-dir=/srv/consul/consul.d
+ExecReload=/bin/kill --signal HUP $MAINPID
+KillMode=process
+Restart=on-failure
+RestartSec=5
+Type=notify
+
+[Install]
+WantedBy=multi-user.target'
+
+# Check if it's SUSE
+. /etc/os-release
+if [[ ! $PRETTY_NAME =~ "SUSE" ]]; then
+    echo "Operating system is not supported. Exiting."
+    exit -1
+fi
+
+echo "Installing trento-agent..."
+
+function install_kvstore() {
+    if ! which unzip  >/dev/null 2>/dev/null; then
+        echo "* Installing unzip"
+        zypper in -y unzip
+    fi
+    mkdir -p $CONFIG_PATH
+    pushd -- "$SRV_HOME" >/dev/null
+    curl -f -sS -O -L "https://releases.hashicorp.com/consul/$KVSTORE_VERSION/consul_${KVSTORE_VERSION}_linux_amd64.zip" >/dev/null
+    unzip -o "consul_${KVSTORE_VERSION}_linux_amd64".zip >/dev/null
+    rm "consul_${KVSTORE_VERSION}_linux_amd64".zip
+    popd >/dev/null
+}
+
+function setup_kvstore() {
+    echo "$HCL_TEMPLATE" \
+        | sed "s|@JOIN_ADDR@|${SERVER_IP}|g" \
+        | sed "s|@BIND_ADDR@|${BIND_IP}|g" \
+        | sed "s|@NODE_NAME@|${NODE_NAME}|g" \
+        > ${CONFIG_PATH}/consul.hcl
+
+    if [ -f "/etc/systemd/system/$SERVICE_FILE_NAME" ]; then
+        echo "  Warning: Systemd unit already installed. Removing..."
+        systemctl stop "$SERVICE_FILE_NAME"
+        rm "/etc/systemd/system/$SERVICE_FILE_NAME"
+    fi
+
+    echo "$SERVICE_FILE" > /etc/systemd/system/$SERVICE_FILE_NAME
+    systemctl daemon-reload
+    # These are good manners to disable all installed services explicitelly
+    systemctl disable $SERVICE_FILE_NAME
+}
+
+function install_trento() {
+    rpm --import ${TRENTO_REPO_KEY} >/dev/null
+    path=${TRENTO_REPO%/*}/
+    if zypper lr --details | cut -d'|' -f9 | grep $path  >/dev/null 2>/dev/null; then
+        echo "* $path repository already exists. Skipping."
+    else
+        zypper ar $TRENTO_REPO >/dev/null
+    fi
+    zypper ref >/dev/null
+    if which trento  >/dev/null 2>/dev/null; then
+        echo "* Trento is already installed. Updating trento"
+        zypper up -y trento >/dev/null
+    else
+	echo "* Installing trento"
+        zypper in -y trento >/dev/null
+    fi
+}
+
+function setup_trento() {
+    # All setting are done by the rpm package
+    # These are good manners to disable all installed services explicitelly
+    # Pay attention, the service is called -----> trento-agent <-------
+    echo "* Setting up trento"
+    systemctl disable trento-agent
+}
+
+install_kvstore
+setup_kvstore
+install_trento
+setup_trento
+echo -e "\e[92mDone.\e[97m"


arbulu89 (changes requested):
Hi @aleksei-burlakov ,
Many things hehe

- You just focused on `trento`. We as well have the requirement to install and configure `consul`, so this part should come in the script as well. The final result should as closest to a single line utility, so these aspects must be abstracted here
-  I don't understand at all this `install_local` thing. I see that you are replicating the web pages installation, but this already makes the assumption that you are running this scripts after cloning or downloading the project. I don't think that this is the best, because it already requires pre-steps (have git running, clone the repo, move to the folder, etc)
- I would split the implementation in different parts: agent and web server. So the idea is that with a single option, the user selects between these 2 options and the selected option is intalled and started.
- The script shouldn't only install the option, it should most probably provide the options to start the tool (and this will need many different steps, like cleaning up current folders, creating others, etc)..
- I don't think we need to compile the project, we already have the executables even in github, maybe we could just download them with this script

Summarizing, this script should download and start the project, and in my opinion without even requiring to have the project downloaded

stefanotorresi:
@aleksei-burlakov you can safely assume that https://download.opensuse.org/repositories/devel:/sap:/trento/ is the canonical download location.

For the time being, what the script should do is to add that repo and leverage zypper in to install the latest tagged version of the Trento Agent.

We definitely _should not_ compile anything here, because that's done already either in the Makefile or in the rpm package itself.

arbulu89:
> @aleksei-burlakov you can safely assume that https://download.opensuse.org/repositories/devel:/sap:/trento/ is the canonical download location. Maybe put that in a variable so that we could potentially override it via the environment.
> 
> For the time being, what the script should do is to add that repo and leverage zypper in to install the latest tagged version of the Trento Agent, or in alternative the latest static binary from GitHub releases.
> 
> We definitely _should not_ compile anything here, because that's done already either in the Makefile or in the rpm package itself.

Are we sure we want to use zypper and rpm option here? This only allows the installation of the latest version first of all, which might not be ideal. Just to point out some limitations

aleksei-burlakov:
@stefanotorresi @arbulu89 thank you for reviewing. I have amended the commit. 

- It adds the repository (now only for 15.3) 
- It installs the consul.
- It doesn't compile anything.

> The script shouldn't only install the option, it should most probably provide the options to start the tool (and this will need many different steps, like cleaning up current folders, creating others, etc)..

@stefanotorresi should the script also start the agent and/or web?

stefanotorresi:
> @stefanotorresi should the script also start the agent and/or web?

Definitely not. The agent will be handled by systemd. I'm not sure yet about the server part.

install.sh
```diff
@@ -0,0 +1,98 @@
...
+}
+
+function install_consul() {
+    if which consul; then
+        echo "consul is already installed. Skipping."
+        return 0
+    fi
+    # Unzip cannot be extracted of the fly, so we store the file, extract and remove.
+    zipfile=$(mktemp)
+    # FIXME! The version is hardcoded. It should take the last existing.
```
stefanotorresi, line 39: maybe for now you could put the version into a variable, so that potentially one could override it via the environment.

install.sh
```diff
@@ -0,0 +1,98 @@
...
+
+function install_trento() {
+    . /etc/os-release
+    if [[ ! $NAME =~ "SUSE" ]]; then
+        echo "Not SUSE operating system."
+        return -1
+    fi
+    case "$VERSION_ID" in
+    "15.3")
+        repo="https://download.opensuse.org/repositories/devel:/sap:/trento/15.3/devel:sap:trento.repo"
```
stefanotorresi, line 12: I would declare this at the top of the file and rename it to all caps, to make it clearer for users that they could override it via `export REPO=foobar`

stefanotorresi (changes requested):
Good progress, Aleksei! We need to tweak it a little bit now.

stefanotorresi:
One more thing I forgot is that we clearly need to indicate somehow that this is covering just the Agent, and the Server installer will be implemented separately.

install.sh
```diff
@@ -0,0 +1,98 @@
...
+        ;;
+    *)
+        echo "Not supported OS release."
```
rtorrero, line 15: This sounds a bit weird to me, I'd write "Unsupported OS release" or "OS release not supported"

install.sh
```diff
@@ -0,0 +1,98 @@
...
+        return -1;;
+    esac
+    path=${repo%/*}/
+    if zypper lr --details | cut -d'|' -f9 | grep $path; then
+        echo "$path repository alreadt exists. Skipping."
```
rtorrero, line 20: Small typo here, I think you meant 'already'

install.sh
```diff
@@ -0,0 +1,98 @@
...
+
+function install_consul() {
+    if which consul; then
+        echo "consul is already installed. Skipping."
+        return 0
+    fi
+    # Unzip cannot be extracted of the fly, so we store the file, extract and remove.
+    zipfile=$(mktemp)
+    # FIXME! The version is hardcoded. It should take the last existing.
+    wget https://releases.hashicorp.com/consul/1.9.8/consul_1.9.8_linux_amd64.zip -O $zipfile
```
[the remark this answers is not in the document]
rtorrero: We could use also environment variables with a default value that if not overriden, gets the architecture using `arch` or `uname -m`

rtorrero, line 40: Also, it would be nice if we can check for requirements such as `wget` or `unzip` before running anything else in the script (the same way you check for `consul` being already installed or not).

install.sh
```diff
@@ -0,0 +1,98 @@
...
+    if (( $rc != 0 )); then
+        return $rc
+    fi
+    echo "Installation successfully completed."
+    return 0
+}
+
+function print_help() {
+    cat <<END
+This is an installer of trento. Trento is a web-based graphical user interface
```
rtorrero, line 62: "an installer of trento" sounds a bit weird, I'd write something in the lines of:

`this is an installer for trento`
or
`this is an one-line installation script for trento`

install.sh
```diff
@@ -0,0 +1,98 @@
+#/bin/bash
+# The script requires root permissions
+
+function install_trento() {
+    . /etc/os-release
+    if [[ ! $NAME =~ "SUSE" ]]; then
+        echo "Not SUSE operating system."
```
rtorrero, line 7: I'd write here "Unsupported OS" here, no need to mention SUSE. Also, the check in the `if` would leave out `SLES`; is that intended?

rtorrero (changes requested):
Thanks Aleksei! it's almost there, just a few minor changes

install.sh
```diff
@@ -0,0 +1,98 @@
...
+        return -1
+    fi
+    case "$VERSION_ID" in
+    "15.3")
```
arbulu89, line 11: Are we sure we only want to support this version?
I would say that the majority of our users will have older versions

install.sh
```diff
@@ -0,0 +1,98 @@
...
+  all         Installs trento and consul.
+  trento      Installs trento.
+  consul      Installs consul.
+  help        Prints this help information
+  version     Prints the version of Trento
+END
+}
+
+function print_version() {
+    echo "Trento version 0"
```
arbulu89, line 79: I guess this is a leftover. We would need to print the correct trento version I guess

install.sh
aleksei-burlakov: The older versions will come later. There is go1.16 and hence the rpm only in 15.3.

install.sh
aleksei-burlakov: I copied it from my trento output. There should be a correct version of course.

install.sh
aleksei-burlakov: Yes, sles was left intentionally.  The installer only takes care that the trento is installed. There was a discussion when implementing an OS check in the trento itself that we cannot fully rely on the NAME field in the os-release, and we also don't need it here. If the version is not supported, the user will see it in the Web Cockpit.

install.sh
aleksei-burlakov: I expect there will be other repos for older releases, so I don't think we should declare it at the top. I will however make it overridable.

install.sh
aleksei-burlakov: Good catch, thanks!

install.sh
```diff
@@ -0,0 +1,98 @@
...
+function install_consul() {
+    if which consul; then
+        echo "consul is already installed. Skipping."
+        return 0
+    fi
+    # Unzip cannot be extracted of the fly, so we store the file, extract and remove.
+    zipfile=$(mktemp)
+    # FIXME! The version is hardcoded. It should take the last existing.
+    wget https://releases.hashicorp.com/consul/1.9.8/consul_1.9.8_linux_amd64.zip -O $zipfile
+    unzip $zipfile -d /usr/bin/
```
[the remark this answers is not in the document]
aleksei-burlakov: Chenged!

install.sh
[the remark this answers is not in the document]
aleksei-burlakov: This should of course be implemented, but I would prefer to do it already in another pull request because I'm afraid it will bring as much logic as the rest of the script. It would need to get somehow the latest version from the web page and interact with user, for example if the consul for his architecture was not found if he would like to install defaults.

install.sh
aleksei-burlakov: I will implement it properly in the later PRs. I doubd we should let the user do something like
`$ CONSUL_VERSION=1.9.9 ./install.sh`
In my understanding the installer is like "lean back, I'll do everything for you". I would prefer to do the best in the script and ask the user if only we completely failed.

install.sh
aleksei-burlakov: Changed to "Operating system is not supported."

install.sh
aleksei-burlakov: Implemented.

install.sh
aleksei-burlakov: Changed to "This is an trento agent installer"

install.sh
aleksei-burlakov: Changed.

install.sh
aleksei-burlakov: Fixed!

install.sh
aleksei-burlakov: Changed to "trento agent installer"

install.sh
aleksei-burlakov: It installs unzip if it's not there.

rtorrero: Perfect, that's what I meant. Should we also do it for `wget`?

install.sh
stefanotorresi: > In my understanding the installer is like "lean back, I'll do everything for you". (with only little interaction - yes/no questions) I would prefer to do the best in the script and ask the user if only we completely failed.

Yes, but we still want to allow for overrides, because we ourselves might need them.
I'm not saying we want to force the user to always set the variable, we should of course provide a default value!

install.sh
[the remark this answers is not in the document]
stefanotorresi: Yes, this is definitely for a follow up.

install.sh
stefanotorresi: We need to make this installer immediately useful for as many people as possible.
Maybe we could download the pre-compiled binary from GitHub for all cases where the rpm is not available?

install.sh
```diff
@@ -0,0 +1,101 @@
...
+    fi
+}
+
+function install_consul() {
+    if which consul; then
+        echo "consul is already installed. Skipping."
+        return 0
+    fi
+    if ! which unzip; then
+        zypper in -y unzip
```
stefanotorresi, line 38: what happens here if we're not running in SUSE?

install.sh
aleksei-burlakov: Actually yes, I just forgot it.

install.sh
aleksei-burlakov: It will fail. Shouldn't it?

install.sh
aleksei-burlakov: I think the existing rpm should work as well for older sle 15.x releases, I will check it.

aleksei-burlakov: > We need to make this installer immediately useful for as many people as possible.
> Maybe we could download the pre-compiled binary from GitHub for all cases where the rpm is not available?

It actually works on sle 15.0 just as file. So I will relax the OS release requirements.

install.sh
aleksei-burlakov: I got what you mean. The SLES has NAME="SLES". Now the script checks the PRETTY_NAME. Either it's leap or sle, its always SUSE ... .

stefanotorresi: That's not what I meant: the check on the distro name is only on the `install_trento` function, so the `install_consul` one is not handling this fail condition gracefully!

rtorrero (dismissed):
LGTM now! Thanks @aleksei-burlakov !

aleksei-burlakov (commented):
I have put the SUSE-OS check at the very beginning of the script so that if it's not SUSE, the script won't install anything.

install.sh
```diff
@@ -0,0 +1,201 @@
...
+CONSUL_VERSION=1.9.6
+CONSUL_HOME=/srv/consul
+CONSUL_CONFIG_PATH="$CONSUL_HOME/consul.d"
+CONSUL_HCL_TEMPLATE='data_dir = "/srv/consul/data/"
+node_name = "@NODE_NAME@"
+log_level = "DEBUG"
+datacenter = "dc1"
+ui = true
+bind_addr = "@BIND_ADDR@"
+client_addr = "@BIND_ADDR@"
```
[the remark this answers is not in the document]
aleksei-burlakov: I've set it back to 0.0.0.0 in https://github.com/trento-project/trento/compare/c15f6869a2334ab103b1cb9022397216611a995d..23e5aea217296981453b30706f865d50d163f26f

aleksei-burlakov:
### How-to try this script.
**Scenario 1**: there are two nodes **S**erver (192.168.122.254) and **A**gent (192.168.122.200) with disabled firewalls.

On node S, install the consul and run it
```
$ wget https://releases.hashicorp.com/consul/1.9.5/consul_1.9.5_linux_amd64.zip
$ unzip consul_1.9.5_linux_amd64.zip
$ ./consul agent -server -bootstrap-expect=1 -data-dir=consul-data -ui -client 192.168.122.254 -bind 192.168.122.254
```

On node C, run the script
`$ curl -s https://raw.githubusercontent.com/trento-project/trento/23e5aea217296981453b30706f865d50d163f26f/install.sh | sudo bash -s -- --bind-ip 192.168.122.200 --server-ip 192.168.122.254`
(The url will change in future. If so, you can call the script as 
`$ sudo ./install.sh --bind-ip 192.168.122.200 --server-ip 192.168.122.254`
)

When the script has finished, one can also start the trento webserver
`$ CONSUL_HTTP_ADDR=http://192.168.122.200:8500 trento web serve`
then open `192.168.122.200:8080` in the browser.

aleksei-burlakov:
It's important to mention, that after the latest changes the script can't be used in a docker container because it uses the **systemctl** (docker containers don't support **systemctl**). This however should not be a problem, because the agent is not supposed to be used in containers.

fabriziosestito (changes requested):

install.sh
```diff
@@ -80,7 +80,7 @@ if [ -z "$CONSUL_SERVER_IP" ]; then
     read -p "Please provide the consul server IP: " CONSUL_SERVER_IP
 fi
 if [ -z "$NODE_NAME" ]; then
-    NODE_NAME="$(hostname)"
+    NODE_NAME="$(hostname)-agent"
 fi
```
fabriziosestito, line 84: Let's remove this

install.sh
```diff
@@ -80,7 +80,7 @@ if [ -z "$CONSUL_SERVER_IP" ]; then
     read -p "Please provide the consul server IP: " CONSUL_SERVER_IP
 fi
 if [ -z "$NODE_NAME" ]; then
-    NODE_NAME="$(hostname)"
+    NODE_NAME="$(hostname)-agent"
```
fabriziosestito, line 83: Let's remove `-agent`

install.sh
```diff
@@ -0,0 +1,212 @@
...
+    fi
+}
+
+function setup_trento() {
+    # All setting are done by the rpm package
+    # If it doesn't automatically start, let's start it here
+    # Pay attention, the service is called -----> trento-agent <-------
+    echo "* Starting trento"
+    systemctl enable --now trento-agent
+}
```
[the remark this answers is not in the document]
fabriziosestito: Follow up: we decided to not enable and to not start the server either

aleksei-burlakov:
@fabriziosestito  As agreed, in https://github.com/trento-project/trento/compare/4cce8ad4b9403b05a8d2981098195b9898a23591..a3882671b6b409ceec4d94a72557b609060163b2
are the changes:

- Decrease the **consul** mentioning
- Remove repeating installing of **curl** and **which**
- Don't appent the **-agent** to the node name
- Disable the services

aleksei-burlakov:
The PR is continued in #226

PR #156, opened 2021-07-20
Change GH actions release trigger

stefanotorresi:
Removes `edit` and `create` triggers, which might cause unwanted workflow runs. We only need to upload the assets when publishing a new release, and that's it.


rtorrero (approved):
LGTM, thanks!

PR #157, opened 2021-07-20
add more rules to the .editorconfig file

stefanotorresi:
This ensures that we all use the same conventions, when using editors that support the [editorconfig.org](https://editorconfig.org/) standard


arbulu89 (approved):

rtorrero (approved):
LGTM

PR #159, opened 2021-07-21
Implement the checks catalog view and services, based on ARA

arbulu89:
Implement a new checks service and view based on ARA. The checks catalog is a Read-only view, where the users will be able to find all the available checks. It doesn't provide the option to select the checks for each of the needed cluster executions. It is just information.

The ARA service implements some REST api calls to the ARA servers to retrieve ARA records which have the checks metadata.

The checks service uses this ARA service to provide a more high level functionalities, which are used later on in the Check catalog handler.

This way, the dependencies are decoupled and we can UT the check catalog frontend mocking all the variables.

**This PR is based on a parallel work where the Ansible checks metadata is posted in the ARA server. It might need some changes in the future, when the Ansible execution to store the metadata is finished.**

check_catalog


stefanotorresi:
I'm not sure about adding a new page for this: it might confuse users into thinking that by navigating into the "checks" section they would see the checks that are being executed. How do you envision this "Checks" section going forward?

arbulu89:
> I'm not sure about adding a new page for this: it might confuse users into thinking that by navigating into the "checks" section they would see the checks that are being executed. How do you envision this "Checks" section going forward?

I don't see any other good place to put the catalog honestly, which is easy to access.
Together with the catalog, I would put a 2nd tab with the `Check results`, where you have the historical of the results, and later on you can filter by hosts, etc, and add some link to the components where these checks were selected.

Anyway, I just wanted to put the view somewhere, to see how it is visualized. We won't have anything definitive until UX gives us the mockups.
Keep in mind that this goes to the `runner` branch, so we have plenty of time to make changes before reaching `main`.

arbulu89:
@stefanotorresi What if I remove the view itself and merge everything else by now?
I would like to use the checks service in the cluster setting modal, to get the checks information

rtorrero:
@arbulu89 @stefanotorresi what about just linking it the menu as "Catalog" and remove the "Checks" parent? I do agree that if it's called "checks" one might possible expect checks results rather than the catalog; but I also think the "catalog" view does provide some value (e.g. as a quick reference when trying to solve problems or maybe even between versions to highlight checks that have been changed/improved somehow)

If in the future we want to link the check run results for each check in the catalog, we could try to do it from that view.

In any case, it looks good to me.

arbulu89:
> @arbulu89 @stefanotorresi what about just linking it the menu as "Catalog" and remove the "Checks" parent? I do agree that if it's called "checks" one might possible expect checks results rather than the catalog; but I also think the "catalog" view does provide some value (e.g. as a quick reference when trying to solve problems or maybe even between versions to highlight checks that have been changed/improved somehow)
> 
> If in the future we want to link the check run results for each check in the catalog, we could try to do it from that view.
> 
> In any case, it looks good to me.

Thanks @rtorrero ,
Yes, there are different options available for this catalog. I would like to get the opinion from @lee-martin

lee-martin:
@arbulu89 Thanks for reaching out.

First of all, I do think the screen is in itself very useful to have, because some users will want to understand how the check is performed without checking sources. I guess the Ansible version may make that part more complicated?

Regarding where to put the page:
Here I agree with @stefanotorresi  that adding a top level menu item "Checks --> Catalogue". I would actually vote for a parent menu item called something like "Trento Configuration" and then have an entry "Checks Catalogue". I say this because I see several configuration elements on the way in Trento (e.g. checking Trento version, update UI, somewhere to customize check templates/tags), so having a "Trento Configuration" menu to host all those things would seem to make sense. Opinions? We do need to collect many such things somewhere soon.

I did not fully understand the comment "put a 2nd tab with the Check results". I think all of the actual check results need to appear within their context, e.g. in the host view.  So I would expect the user to start in the component (e.g. host detail) and then from there have the ability to see historic data, once we have a history in the backend. I think offering a history starting with the check name would be pretty challenging for a user to follow and understand the context.

We also need somewhere to list which checks are part of which check group (e.g. HANA scale-up), perhaps that could also be part of Trento configuration somewhere?


I would support submitting this with menu naming like above (to avoid the potential confusion @stefanotorresi  mentioned) and then we can iterate from there.

stefanotorresi:
Let's go with the "Settings" page and a "Checks catalog" subitem.

arbulu89:
Hi @stefanotorresi @lee-martin ,
New implementation with the renaming from `Checks` to `Settings` and the grouping/accordion done.
Here a small demo:
accordion

fabriziosestito (approved):

PR #160, opened 2021-07-21
Allow multiple SAP system tags per host

rtorrero:
This PR adds support for having multiple SIDs on the same host (useful in cost-optimized scenario). The host metadata field is now `trento-sap-systems` and the filters have been updated to allow matching e.g. `PRD` in a host where `PRD,QAS` is stored in the metadata.

Templates have been updated too to now show multiple SIDs where relevant.

Updated: Screenshots
hostdetails
hosts

### DIFF
--- a/agent/discovery/sapsystem.go
+++ b/agent/discovery/sapsystem.go
@@ -39,19 +39,19 @@ func (d SAPSystemsDiscovery) Discover() error {
 		if err != nil {
 			return err
 		}
+	}
 
-		// Store SAP System, Landscape and Environment names on hosts metadata
-		err = storeSAPSystemTags(d.discovery.client, s)
-		if err != nil {
-			return err
-		}
+	// Store SAP System, Landscape and Environment names on hosts metadata
+	err = storeSAPSystemTags(d.discovery.client, d.SAPSystems)
+	if err != nil {
+		return err
 	}
 
 	return nil
 }
 
-func storeSAPSystemTags(client consul.Client, system *sapsystem.SAPSystem) error {
-	envName, landName, sysName, err := loadSAPSystemTags(client, system.SID)
+func storeSAPSystemTags(client consul.Client, systems sapsystem.SAPSystemsList) error {
+	envName, landName, err := loadSAPSystemTags(client, systems[0].SID)
 	if err != nil {
 		return err
 	}
@@ -59,7 +59,9 @@ func storeSAPSystemTags(client consul.Client, system *sapsystem.SAPSystem) error
 	// If we didn't find any environment, we create a new default one
 	if envName == "" {
 		land := environments.NewDefaultLandscape()
-		land.AddSystem(environments.NewSystem(sysName, system.Type))
+		for _, system := range systems {
+			land.AddSystem(environments.NewSystem(system.SID, system.Type))
+		}
 		env := environments.NewDefaultEnvironment()
 		env.AddLandscape(land)
 
@@ -71,11 +73,13 @@ func storeSAPSystemTags(client consul.Client, system *sapsystem.SAPSystem) error
 		landName = land.Name
 	}
 
+	sysNames := systems.GetSIDsString()
+
 	// Store host metadata
 	metadata := hosts.Metadata{
 		Environment: envName,
 		Landscape:   landName,
-		SAPSystem:   sysName,
+		SAPSystems:  sysNames,
 	}
 
 	err = metadata.Store(client)
@@ -88,13 +92,13 @@ func storeSAPSystemTags(client consul.Client, system *sapsystem.SAPSystem) error
 
 // These methods must go here. We cannot put them in the internal/sapsystem.go package
 // as this creates potential cyclical imports
-func loadSAPSystemTags(client consul.Client, sid string) (string, string, string, error) {
+func loadSAPSystemTags(client consul.Client, sid string) (string, string, error) {
 	var env, land string
 	sys := sid
 
 	envs, err := environments.Load(client)
 	if err != nil {
-		return env, land, sys, err
+		return env, land, err
 	}
 	for envKey, envValue := range envs {
 		for landKey, landValue := range envValue.Landscapes {
@@ -107,5 +111,5 @@ func loadSAPSystemTags(client consul.Client, sid string) (string, string, string
 			}
 		}
 	}
-	return env, land, sys, err
+	return env, land, err
 }
--- a/internal/hosts/hosts.go
+++ b/internal/hosts/hosts.go
@@ -89,7 +89,7 @@ func (n *Host) HAChecks() *check.Controls {
 	return checks
 }
 
-func (n *Host) GetSAPSystems() (map[string]*sapsystem.SAPSystem, error) {
+func (n *Host) GetSAPSystems() (sapsystem.SAPSystemsMap, error) {
 	systems, err := sapsystem.Load(n.client, n.Name())
 	if err != nil {
 		return nil, err
@@ -101,17 +101,25 @@ func (n *Host) GetSAPSystems() (map[string]*sapsystem.SAPSystem, error) {
 // https://github.com/hashicorp/consul/blob/master/agent/consul/catalog_endpoint.go#L298
 func CreateFilterMetaQuery(query map[string][]string) string {
 	var filters []string
+	var operator string
 	// Need to sort the keys to have stable output. Mostly for unit testing
 	sortedQuery := sortKeys(query)
 
 	if len(query) != 0 {
 		var filter string
 		for _, key := range sortedQuery {
+			switch key {
+			case "trento-sap-systems":
+				operator = "contains"
+			default:
+				operator = "=="
+			}
 			if strings.HasPrefix(key, TrentoPrefix) {
 				filter = ""
 				values := query[key]
 				for _, value := range values {
-					filter = fmt.Sprintf("%sMeta[\"%s\"] == \"%s\"", filter, key, value)
+					filter = fmt.Sprintf("%sMeta[\"%s\"] %s \"%s\"", filter, key, operator, value)
+
 					if values[len(values)-1] != value {
 						filter = fmt.Sprintf("%s or ", filter)
 					}
--- a/internal/sapsystem/sapsystem.go
+++ b/internal/sapsystem/sapsystem.go
@@ -30,6 +30,7 @@ const (
 )
 
 type SAPSystemsList []*SAPSystem
+type SAPSystemsMap map[string]*SAPSystem
 
 // A SAPSystem in this context is a SAP installation under one SID.
 // It will have application or database type, mutually exclusive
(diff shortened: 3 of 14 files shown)


arbulu89:
@rtorrero I would appreciate some screenshots of the changed templates of the frontend

agent/discovery/sapsystem.go
```diff
@@ -33,33 +33,49 @@ func (d SAPSystemsDiscovery) Discover() error {
...
 	}
 
 	return nil
 }
 
-func storeSAPSystemTags(client consul.Client, system *sapsystem.SAPSystem) error {
-	envName, landName, sysName, err := loadSAPSystemTags(client, system.SID)
-	if err != nil {
-		return err
+func storeSAPSystemTags(client consul.Client, systems []*sapsystem.SAPSystem) error {
```
arbulu89, line 56: As commented above, I think we could use `sapsystem.SAPSystemsList` as `systems` type

rtorrero:
> @rtorrero I would appreciate some screenshots of the changed templates of the frontend

yes! I wanted to post some but my test environment  is broken right now. Deploying a new one and should be able to add them soon

agent/discovery/sapsystem.go
```diff
@@ -33,33 +33,49 @@ func (d SAPSystemsDiscovery) Discover() error {
...
+	var envName, landName string
+	sysNames := ""
+	var err error
+	for i, system := range systems {
```
arbulu89, line 60: This looks a bit over complicated.
I suggest the next:
As `envName` and `landName` will be the same independtly which `system.SID` we use, we could just call it once with the 1st SID.
After that, i would prefer the usage of `strings.Join` to create the sids string. I don't know if we could use some kind of lamdba method to get the SID, otherwise we will need to loop though all the systems, create a SID string array, and apply the `Join` function

agent/discovery/sapsystem.go
```diff
@@ -33,33 +33,49 @@ func (d SAPSystemsDiscovery) Discover() error {
...
+		if len(systems) > (i + 1) {
+			sysNames = sysNames + ","
+		}
 	}
 
 	// If we didn't find any environment, we create a new default one
 	if envName == "" {
 		land := environments.NewDefaultLandscape()
-		land.AddSystem(environments.NewSystem(sysName, system.Type))
+		for _, system := range systems {
```
arbulu89, line 76: Maybe we can run this in the same loop used above. Just think about it. Maybe it is not possible anyway

internal/hosts/hosts.go
```diff
@@ -111,7 +127,12 @@ func CreateFilterMetaQuery(query map[string][]string) string {
 				filter = ""
 				values := query[key]
 				for _, value := range values {
-					filter = fmt.Sprintf("%sMeta[\"%s\"] == \"%s\"", filter, key, value)
+					if key == "trento-sap-systems" {
```
arbulu89, line 130: Maybe a `switch/case` syntax would look nicer

internal/hosts/hosts.go
```diff
@@ -97,6 +97,22 @@ func (n *Host) GetSAPSystems() (map[string]*sapsystem.SAPSystem, error) {
 	return systems, nil
 }
 
+func (n *Host) GetSAPSystemsList() (sapsystem.SAPSystemsList, error) {
+	systems, err := sapsystem.Load(n.client, n.Name())
+	if err != nil {
+		return nil, err
+	}
+
+	systemsList := make(sapsystem.SAPSystemsList, len(systems))
```
arbulu89, line 106: Why do we have this loop? Cannot we use directly the `systems` map? At the end they have the same content

internal/hosts/hosts.go
```diff
@@ -97,6 +97,22 @@ func (n *Host) GetSAPSystems() (map[string]*sapsystem.SAPSystem, error) {
 	return systems, nil
 }
 
+func (n *Host) GetSAPSystemsList() (sapsystem.SAPSystemsList, error) {
+	systems, err := sapsystem.Load(n.client, n.Name())
```
arbulu89, line 101: I don't know if it's a good idea to do this...
Cannot we simply use the tag value and split the string by the comma entry?

arbulu89 (changes requested):
Hi @rtorrero ,
Many comments hehe
In general the implementation looks good, but I think we can remove and re-order many of the new changes.
Finally, as added in independent comment, it would be nice to have a screenshot/gif of the final result

internal/hosts/hosts.go
rtorrero: I assume that by "using directly the `systems` map you mean using the `GetSAPSystems` method?
The thing is that if I use `GetSAPSystems` in the template, I can't easily add a comma after each SID but the last one (since in a string map, each key is not an int but a string, so I can't easily figure out the `len-1` case :/... It might be that I'm not familiar enough with templates. A bit more details if you can think of a way of using the string map in the template would help

internal/sapsystem/sapsystem.go
```diff
@@ -77,6 +77,19 @@ type CustomCommand func(name string, arg ...string) *exec.Cmd
 
 var customExecCommand CustomCommand = exec.Command
 
+func (systems SAPSystemsList) GetSIDs() string {
+	var sysNames string
```
arbulu89, line 81: I would rename the func name to `GetSIDsString` maybe
Why not simply (`Join` takes care to put not putting the separator at the end):
```
        var sids []string
	for _, sid := range systems {
	  sids = append(sids, system.SID)
	}
	return strings.Join(sids, ",")
```

internal/hosts/hosts.go
arbulu89: As we just need to split the string by the `,`, why don't we just add the `Split` method to our templating system, so we can use it?
Here an example: https://stackoverflow.com/questions/19771787/golang-how-to-split-string-in-template
We would need to add the func here: https://github.com/trento-project/trento/blob/main/web/layout.go#L142

I see that you need this already 2 times, so it would make sense

arbulu89 (commented):
@rtorrero Some new comments

agent/discovery/sapsystem.go
```diff
@@ -39,27 +39,29 @@ func (d SAPSystemsDiscovery) Discover() error {
...
+		return err
 	}
 
 	return nil
 }
 
-func storeSAPSystemTags(client consul.Client, system *sapsystem.SAPSystem) error {
-	envName, landName, sysName, err := loadSAPSystemTags(client, system.SID)
+func storeSAPSystemTags(client consul.Client, systems sapsystem.SAPSystemsList) error {
+	envName, landName, _, err := loadSAPSystemTags(client, systems[0].SID)
```
arbulu89, line 54: The `loadSAPSystemTags` is only used here, so we could even remove this 3rd return value (which is not used)

internal/environments/environments.go
```diff
@@ -107,6 +107,12 @@ func (l *Landscape) AddSystem(system *SAPSystem) {
 	l.SAPSystems[system.Name] = system
 }
 
+func (l *Landscape) AddSystems(systems []*SAPSystem) {
```
arbulu89, line 110: This new method is not used anywhere. I think you created this to replace the loop used above maybe, at line 62 of `agent/discovery/sapsystem.go`.
You could whether remove this function or use it above replacing the loop (you might need to change the incoming variable type)

web/layout.go
```diff
@@ -149,6 +150,9 @@ func (r *LayoutRender) addFileFromFS(templatesFS fs.FS, file string) {
 			return a + b
 		},
 		"markdown": markdownToHTML,
+		"split": func(s string, sep string) []string {
```
arbulu89, line 153: Did you try just with `"split": strings.Split` here?

web/templates/host.html.tmpl
```diff
@@ -7,15 +7,16 @@
...
-            {{ $SAPSys := index .Host.TrentoMeta "trento-sap-system" }}
             <dt class="inline">Environment</dt>
             <dd class="inline"><a href="/environments/{{ $Env }}">{{ $Env }}</a></dd>
             <dt class="inline">Landscape</dt>
             <dd class="inline"><a href="/landscapes/{{ $Land }}?environment={{ $Env }}">{{ $Land }}</a></dd>
             <dt class="inline">SAP System</dt>
-            <dd class="inline"><a
-                        href="/sapsystems/{{ $SAPSys }}?environment={{ $Env }}&landscape={{ $Land }}">{{ $SAPSys }}</a>
+            <dd class="inline">
+                {{- range $i, $v := split (.Host).GetSIDsString ","}}{{- if $i }},{{- end }} <a
```
arbulu89, line 16: Do we need this `.Host.GetSIDsSting` function?
It will return the same value we have in the metadata, right?
I mean, why simply not: `split (index .Host.TrentoMeta "trento-sap-systems) ","`
And we get rid off `GetSIDsString` function from the `Host` struct

internal/sapsystem/sapsystem.go
```diff
@@ -77,6 +78,26 @@ type CustomCommand func(name string, arg ...string) *exec.Cmd
 
 var customExecCommand CustomCommand = exec.Command
 
+func (sm SAPSystemsMap) GetSIDsString() string {
```
arbulu89, line 81: This duplication looks horrible hehe
I'm sure you will be able to remove this 1st function once you have remove the `.Host.GetSIDstring`

arbulu89 (changes requested):
Hi @rtorrero ,
More comments

rtorrero:
@arbulu89 should be fine now! thanks for the comments

internal/sapsystem/sapsystem.go
```diff
@@ -77,6 +78,16 @@ type CustomCommand func(name string, arg ...string) *exec.Cmd
 
 var customExecCommand CustomCommand = exec.Command
 
+func (sl SAPSystemsList) GetSIDsString() string {
```
arbulu89, line 81: As a convention, I would put this method after the `NewSAPSystemsList` function

web/hosts_test.go
```diff
@@ -98,12 +98,28 @@ func TestHostsListHandler(t *testing.T) {
...
 				"trento-sap-environment": "env2",
 				"trento-sap-landscape":   "land2",
-				"trento-sap-system":      "sys2",
+				"trento-sap-systems":     "sys3",
 				"trento-cloud-provider":  "gcp",
 			},
 		},
 	}
 
+	// sapSystemMapFoo := map[string]interface{}{
```
arbulu89, line 107: Is this comment here on purpose or some leftover?

web/hosts_test.go
```diff
@@ -170,19 +186,29 @@ func TestHostsListHandler(t *testing.T) {
 
 	kv.On("ListMap", consul.KvEnvironmentsPath, consul.KvEnvironmentsPath).Return(filters, nil)
 
+	// sapsystemPathFoo := "trento/v0/hosts/foo/sapsystems/"
```
arbulu89, line 189: Is this comment here on purpose or some leftover?

arbulu89 (dismissed):
Thanks @rtorrero ,
Now it looks great.
I have put some minor comments if you want to have a look

rtorrero:
Sorry again... fixed

arbulu89 (approved):

PR #161, opened 2021-07-21
Add debug target to avoid always stripping debug symbols

rtorrero:
This PR adds a new debug target which behaves slightly different from the `default` target:

 - It does not run the tests
 - It doesn't use `-trimpath` which causes some issues while debugging remotely with `delve`
 - It doesn't use `-s -w` in the `LDFLAGS` which cause thinner binaries but strips symbols away needed for debugging

Note that this is only useful for development and should provide identical binaries as before when running the `default` target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,16 @@
 VERSION ?= $(shell hack/get_version_from_git.sh)
 LDFLAGS = -X github.com/trento-project/trento/cmd.version="$(VERSION)"
-GO_BUILD = CGO_ENABLED=0 go build -ldflags "$(LDFLAGS)"
 ARCHS ?= amd64 arm64 ppc64le s390x
+DEBUG ?= 0
 
-default: LDFLAGS += -s -w
-default: GO_BUILD += -trimpath
+ifeq ($(DEBUG), 0)
+	LDFLAGS += -s -w
+	GO_BUILD = CGO_ENABLED=0 go build -ldflags "$(LDFLAGS)" -trimpath
+else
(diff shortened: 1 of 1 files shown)


stefanotorresi (commented):
I'm still not sure that an additional make target is the way to go. The basic target to compile is `make build`. I think we should just make that more flexible.
Have you considered something like `DEBUG=true make`? This way the build flags would be changed no matter what target the user is selecting, be that `make build` or `make cross-compile` or whatever: otherwise, one is forced to run all the things that are in the new `debug` target, and is left with no ability to fine-tune the build flags when using all the other targets.

rtorrero:
Yes, now that I think about it it makes much more sense to use a conditional in this case. Let me know if it's what you meant; I'll squash & merge :+1:

arbulu89:
@stefanotorresi I think this is waiting for you. Honestly, I don't find myself enough capable to take this decision, and as you are the one with more experience with go packaging

Makefile
```diff
@@ -1,8 +1,17 @@
...
+
+ifeq ($(DEBUG), 0)
+	LDFLAGS += -s -w
+	GO_BUILD = CGO_ENABLED=0 go build -ldflags "$(LDFLAGS)" -trimpath
+else
+	GO_BUILD = CGO_ENABLED=0 go build -ldflags "$(LDFLAGS)"
+endif
 
 default: clean mod-tidy fmt vet-check test build
+debug: clean mod-tidy fmt vet-check build
```
stefanotorresi, line 14: I guess this target can be left out now, right?

stefanotorresi (changes requested):

stefanotorresi (approved):

PR #162, opened 2021-07-21
Ansible for checking and remediation

brett060102:
This is a WP:
Issues:
1: all cases where shell is being used for checking need to be updated to make sure
    ansible change status get correctly set on --check. Is initial code status was set to no change, since the code
    was not making a change, but for integration with ara we want change to be set to indicate that remediation
    would result in a change.

2: Not all tests have meta.yml files

3: Existing tests need to be validated against current examples/generic-azure.yaml since that has changed since
    since we started.


ansible/ansible_local.yml.sample
```diff
@@ -0,0 +1,5 @@
+target:
```
arbulu89, line 1: We can remove this file, right?
It looks like some local testing utility

ansible/ansible_remote.yml.sample
```diff
@@ -0,0 +1,10 @@
+target:
```
arbulu89, line 1: The same here,
We can remove this file, right?

ansible/README.md
```diff
@@ -0,0 +1,3 @@
+# trento/ansible
```
arbulu89, line 1: Most probably we need to upgrade this documentation.
For other occasion I guess

ansible/ha-meta.yml
```diff
@@ -0,0 +1,230 @@
+# This playbook gathers meta data facts for SAP/HANA checker.
+
+- hosts: all
```
arbulu89, line 3: We don't need to run this in `all` hosts.
We have to run it in `localhost`, and remove all of this `delegate_to: localhost` from all the roles. it is redundant

ansible/ha-meta.yml
```diff
@@ -0,0 +1,230 @@
...
+      import_role:
+        name: load_facts
+      delegate_to: localhost
+      run_once: yes
+
+    - name: include 1.1.1 meta
+      import_role:
+        name: '1.1.1'
+        tasks_from: meta
+      tags:
```
arbulu89, line 16: Together with the other propose change, I would remove the `tags` and `run_once`. They don't need to be there

ansible/ha-remediation.yml
```diff
@@ -0,0 +1,220 @@
...
+    - name: include load_facts
+      import_role:
+        name: load_facts
+      tags:
+        - always
+
+    - name: include 1.1.1 remediate
+      import_role:
+        name: '1.1.1'
+        tasks_from: main
```
arbulu89, line 15: Do we need this `main` thing? Is it not redundant?

arbulu89 (changes requested):
Hi @brett060102 ,
A lot of things hehe
- I find a lot of unnecessary files in the PR. I think we should remove all the not needed files
- I don't know what we will do with the `vars` files. I don't know if we will provide the option to change them to the users, so maybe initially we should just hardcode them. Just an opinion, this needs a lot of discussion
- It looks like there is some mess with the `main` and `meta` files. Some of the times they are mixed (1.4.2/tasks/meta.yml` for example).
- Personally I would remove all the `meta.yml` files from this PR, and we could add them tottaly independent in other PR (actually, I have the implementation done).
- I would remove all the remediation tasks that are not finished to have better track on the finished and unfinished things.

Soooo,
I think we need to split all of these in multiple PRs.
1. Create the structure and the common remediation tasks, we could add a couple of tasks initially
2. Create the whole metadata structure (as commented, I would prefer that I take care of this)
3. Go adding the tasks 1 by 1, or in small groups so we can do a good review process and track all the finished tasks

ansible/ha-meta.yml
brett060102: OK.

ansible/ha-meta.yml
brett060102: done

ansible/ha-remediation.yml
brett060102: Done.

brett060102 (commented):
I think I addressed all of the current comments.

stefanotorresi:
Hey @brett060102, this is quite a lot of new code for one batch and it would take really long for us to dig into it.

Since we only need to validate the integration with the Ansible runner right now, could you by any chance resubmit with just one single check?
We can add the rest of them later, possibly in groups.

brett060102:
@stefanotorresi and @arbulu89 
I pre-popualated the meta files from the main files and did some of them so we could test that part, but stopped changing that
when  you said to leave that to you.  I can either delete those, or change the contents to:
---

- name: Set Test ID
  set_fact:
    test_id: 'TEST_ID'

 
I can split, but let me do a separate PR for that. I think that It needs at least four test cases though.
I would propose the following:
1.1.1 (changed config file example)
1.1.2.runtime (all shell example, makes no changes, but does restart cluster)
1.2.1 (all shell example)
1.3.5 (shell example, manual remediation)

So, the new PR would have:
man and meta for cases above
ha-meta.yml for cases above
ha-remediation.yml for cases above
current:
   vars/dev_env.yml
   vars/azure.yml

How does that sound.

brett060102:
Closing this PR. A smaller change with just a few tests will be put up.

PR #164, opened 2021-07-22
Anible infra

brett060102:
Create ansible checker frame work and 4 sample tests.


ansible/checker.yml
```diff
@@ -0,0 +1,46 @@
...
+    - name: include 1.3.5 remediate
+      import_role:
+        name: '1.3.5'
+      tags:
+        - 1.3.5
+
+    - name: set_global_results
+      set_fact:
+        global_results: "{{ hostvars.hana1.tests }}{{ hostvars.hana2.tests }}"
+      run_once: yes
```
brett060102, line 39: I know that this is not right and I am still working on it.  It is an interaction issue ara_record block below. That block must be done on local host, and should be done once to write out all test results, but test results are stored as per host fact.
So, the need to loop over all hosts in inventory merge all test results. Still working on that, but wanted to get the rest out for review.

brett060102: Should be OK now.

brett060102:
Sorry, I did a force update in my branch because I was doing that before the PR and was not careful when using my history.

ansible/ansible_remote.yml.sample
```diff
@@ -0,0 +1,10 @@
+target:
```
arbulu89, line 1: I would prefer to remove this file, but well, if it helps on testing purposes

ansible/checker.yml
```diff
@@ -0,0 +1,48 @@
+# This playbook deploys a simple stand alone checker for SAP/HANS.
```
arbulu89, line 1: Let's remove or update this comment. It looks outdated

arbulu89: I would rename the file name from `checker.yml` to `check.yml`. This runs an action, so an imperative verb looks more appropiate

ansible/checker.yml
```diff
@@ -0,0 +1,48 @@
...
+    - name: include load_facts
+      import_role:
+        name: load_facts
+      tags:
+        - always
+
+    - name: include 1.1.1 remediate
+      import_role:
+        name: '1.1.1'
+      tags:
```
arbulu89, line 15: Let's remove the `tags` by now, as we don't need them.
If we find that they are required in the future, we will see how we name and use them

ansible/checker.yml
```diff
@@ -0,0 +1,48 @@
...
+- hosts: all
+
+  tasks:
+    - name: include load_facts
+      import_role:
+        name: load_facts
+      tags:
+        - always
+
+    - name: include 1.1.1 remediate
```
arbulu89, line 12: Let's name it as `Include 1.1.1`.
We are going to avoid the usage of `remediate` word by now

ansible/checker.yml
```diff
@@ -0,0 +1,48 @@
...
+      import_role:
+        name: '1.3.5'
+      tags:
+        - 1.3.5
+
+    - name: set_global_results
+      set_fact:
+        global_results: "{{ (global_results |  default({})) }}{{ hostvars[item]['tests'] }}"
+      with_inventory_hostnames:
+        - all
```
arbulu89, line 40: I think that this won't scale well. We are not going to have the host names under `all` most probably.
Anyway, we can leave this way by now

ansible/checker.yml
```diff
@@ -0,0 +1,48 @@
...
+
+    - name: Set Ara Global Results
+      ara_record:
+        key: "Test_Restults"
```
arbulu89, line 44: Let's use the `trento-results` key and add the `type: json` after the `value` field

ansible/meta.yml
```diff
@@ -0,0 +1,37 @@
+# This playbook gathers meta data facts for SAP/HANA checker.
+
+- hosts: localhost
+
+  tasks:
+    - name: include load_facts
```
arbulu89, line 6: I don't think we need to load the facts here.
Let's remove it if it is not completely needed

ansible/meta.yml
```diff
@@ -0,0 +1,37 @@
...
+      import_role:
+        name: '1.2.1'
+        tasks_from: meta
+
+    - name: include 1.3.5 meta
+      import_role:
+        name: '1.3.5'
+        tasks_from: meta
+
+    - name: Set Global Meta Data
```
arbulu89, line 30: Do we need this task for something?
I guess we could use the `tests_meta_data` directly

ansible/meta.yml
```diff
@@ -0,0 +1,37 @@
...
+      set_fact:
+        global_meta_data: "{{ (tests_meta_data | default({})) }}"
+
+    - name: Set Ara Global Meta Data
+      ara_record:
+        key: "Test_Meta_Data"
+        value: "{{ global_meta_data | to_nice_json  }}"
```
arbulu89, line 37: Let's calle the key `trento-meta` and add the `type: json` field

ansible/roles/1.1.1.runtime/tasks/main.yml
```diff
@@ -0,0 +1,33 @@
+---
+
+- name: Set Test ID
+  set_fact:
+    test_id: '1.1.1_runtime'
+
+- name: "Set other facts for {{ test_id }}"
```
arbulu89, line 7: This is metadata. It should only go in the `meta.yml` file

ansible/roles/1.1.1.runtime/tasks/main.yml
```diff
@@ -0,0 +1,33 @@
...
+    - config_updated is changed
+    - not ansible_check_mode
+
+- block:
+    - import_role:
+        name: post-results
+  when:
+    - ansible_check_mode
+  vars:
+    status: "{{ config_updated is changed | ternary ('Failed','Passed') }}"
```
arbulu89, line 33: I would prefer to have a simple `true/false` option maybe here.
It is simpler to parse if we agree that `true` means `passed` and `false` means `failed`.
Otherwise we need to know this naming conventions in the code itself

ansible/roles/cluster_start/tasks/main.yml
```diff
@@ -0,0 +1,11 @@
+---
```
arbulu89, line 1: We should better remove this implementations by now.
Let's only focus in the `checks` part

ansible/roles/dump_facts/tasks/main.yml
```diff
@@ -0,0 +1,13 @@
+---
```
arbulu89, line 1: Not used

ansible/vars/azure_env.yml
```diff
@@ -0,0 +1,75 @@
+---
+cloud_type: "azure"
```
arbulu89, line 2: We won't use this `cloud_type` most probably

ansible/vars/azure_env.yml
```diff
@@ -0,0 +1,75 @@
...
+
+expected:
+  "1.1.1": "30000"
+  "1.1.1_runtime": "30000"
+  "1.1.2": "36000"
```
arbulu89, line 7: Let's remove all the not used entries here. Put only the ones that apply to the currently added tests

arbulu89 (changes requested):
Hi @brett060102 ,

Thank you for this smaller PR.
Many things:
- Change the target branch to `runner`
- Let's focus only in the `checks` part by now and remove any remediation code (unless the task is a check and remediation together, of course)
- Let me introduce the `metadata` code myself. I will add this way all the fields and names I need. This means that we need to remove all the references to metadata in the check tasks.
- I find a lot of not used files, test files, etc. If they are test files, create a `test` folder here, and put them there. In fact, I would prefer to remove them if possible

So, my expectation would be:
- Let's move this `ansible` folder to `runner/ansible`. This info is new, but better do it here, in this PR
- Within `ansible` folder, we have the `roles` folder and the `vars` folder. Together with them, we will have the `check.yml` and `meta.yml`. The vars folder will have a unique vars file by now. We will need to have as many as groups vars files later on, but by now it is enough
- We remove any reference or implementation to the remediation parts
- Let's drop the metadata implementation from this PR
- I would prefer to only even have a unique `test`. Let's use that one to do the review. I already see some tests (1.3.5) that are complex enough to deserver a self review

brett060102:
@arbulu89 sorry, I meant to put this up against runner. I don't think I can change this PR to do that. So, I will do the changes I can and them put up a new PR against runner. I don't think that having separate .yml's for check and remediation is the way to go, since some much code is shared between the checking and remediation.

brett060102:
I am closing this PR. A new one will be put up against  the runner branch.

PR #165, opened 2021-07-23
Ansible runner improvements - Run meta playbook

arbulu89:
Improvements on the Ansible runner component.

Major changes:
- Now, rather than running one ansible playbook in a loop, the runner runs an initial metadata playbook, which will upload the metadata to ARA for later consumption
- Some major changes in how the `AnsibleRunner` is used, to provide more flexibility, as we need customization in each of the ansible calls. The `RunPlaybook` function covers more use cases (like running without any inventory)
- Add some UT to check the most important parts of the code

**The playbook names might change after we have something established about the whole Ansible playbooks structure**


runner/runner_test.go
```diff
@@ -0,0 +1,101 @@
...
+		cmdCallback,
+	)
+	mockCommand.On("Execute", "python3", "-m", "ara.setup.action_plugins").Return(
+		cmdAction,
+	)
+
+	a, err := NewAnsibleMetaRunner(cfg)
+
+	expectedMetaRunner := &AnsibleRunner{
+		Playbook: "../test/ansible_test/ansible/meta.yml",
```
rtorrero, line 48: It may be better to try to declare constants on top of the file with the Ansible playbooks directories and file names to avoid repeating them several times. No big deal though.

rtorrero (approved):
Appart from the small comment; everything looks fine to me. Since it's just a matter of "style" (thus possibly opinionated), I'm approving anyway. Great job :-)!

runner/runner_test.go
arbulu89: Improved in: https://github.com/trento-project/trento/pull/165/commits/0b54d0ca6fabbbec0f7ed5ab6468d492c123871a

PR #166, opened 2021-07-23
[PoC] Service layer & Consul long polling projector

fabriziosestito:
This PoC introduces:
- A persistence access layer
-  A model layer shared between the persistence layer and the controllers
- A service/respository layer, which is consumed by the handlers and deals with the business logic of the application,
- A projector layer which projects the data from consul by long polling consul resources (blocking query) 

To demonstrate the abstraction, the Hosts list page is redesigned by using the relational model and the service layer.
Projectors are introduced to process and store the data from the consul catalog.

Tests are added, i suggest starting from them top down (handlers, service, projectors) to review the code.

`ProjectorHandlers` deal with the projection behavior (query consul, project the data), while the `Projector` type runs a goroutine which calls the handlers `Query` and `Project` functions from the handler and updates the index of the last seen update which is then persisted in transaction. This basically means that if the server reboots it knows where to start projecting and that if the transaction fails the last seen index is not updated.

Some notes:
for the scope of this PoC, a go implementation of sqlite3 is used: https://gitlab.com/cznic/sqlite
This is _only for testing_, since the standard driver needs _CGO_ enabled, and I wasn't able to compile a static binary which worked on SLES (also, the binary size gets way bigger).
Due to this the concurrency of the database access is limited to 1 (this is an actual limitation also of the standard sqlite3 golang driver) and doesn't compile in all the architectures we are planning to support (this is why the CI is red, but the tests actually run on amd64). For this reasons, I would switch to postgre if we are going to follow this path.


arbulu89 (commented):
Hi @fabriziosestito ,
The PoC looks valuable.
I have to confess that unless we move the KV usage from consul to this sqlite3 database (or any other sql based db), the effort looks just partial.
The hosts information is just a tiny part of the whole dataset as we have commented many times.
I don't know how I would proceed, but in my opinion, if we only intend to implement this part, the added complexity doesn't return back that much.

I find the PoC as valid, but we need to continue thinking about how are we going to move the whole dataset to this database, and even start thinking about removing consul at all.
I don't want to move to this change without having all the cards on top of the table, because it would be other intermediate movement (even though we could absolutely reuse many things, like the database itself and the models, the projector is the part that makes things complex)

stefanotorresi:
We need to pick this back up, guys!

PR #168, opened 2021-07-26
Display trento version UI

fabriziosestito:
Added the trento version to the UI.
Had to move the Version injected variable to a separate package (was in `cmd` before).
Version links to the github release, only the tag is displayed (without `+git.yabba.yabba`) for UX reasons (text was too long, looks bad).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 VERSION ?= $(shell hack/get_version_from_git.sh)
-GO_BUILD = CGO_ENABLED=0 go build -trimpath -ldflags "-s -w -X github.com/trento-project/trento/cmd.version=$(VERSION)"
+GO_BUILD = CGO_ENABLED=0 go build -trimpath -ldflags "-s -w -X github.com/trento-project/trento/version.Version=$(VERSION)"
 ARCHS ?= amd64 arm64 ppc64le s390x
 
 default: clean mod-tidy fmt vet-check test build
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -5,19 +5,18 @@ import (
 	"runtime"
 
 	"github.com/spf13/cobra"
+	"github.com/trento-project/trento/version"
 )
 
 func init() {
 	rootCmd.AddCommand(versionCmd)
 }
 
-var version string
-
 var versionCmd = &cobra.Command{
 	Use:   "version",
 	Short: "Print the version number of Trento",
 	Long:  `All software has versions. This is Trento's`,
 	Run: func(cmd *cobra.Command, args []string) {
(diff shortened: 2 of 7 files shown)


arbulu89:
Hi @fabriziosestito ,
Without doing any review, a screenshot would be appreciated (as far as I can see, the version comes in the sidebar, so we should take care of the collapsed and not collapsed options)

fabriziosestito:
> Hi @fabriziosestito ,
> Without doing any review, a screenshot would be appreciated (as far as I can see, the version comes in the sidebar, so we should take care of the collapsed and not collapsed options)

@arbulu89 
image

image

fabriziosestito:
@arbulu89 
image

image

arbulu89 (dismissed):
Thanks @fabriziosestito ,
The code looks good.
In the other hand, I'm not that convinced of the output in the web, it looks a bit out of place (sepcially in the collapsed sidebar).
I think it is worthy to comment this with the UX team.
What do you thin @en3sis ?

fabriziosestito:
@arbulu89
Yes you are right.
Also, i find a bit weird that the collapsed menu shows the full footer while the expanded menu shows the collapsed one.

arbulu89:
> @arbulu89
> Yes you are right.
> Also, i find a bit weird that the collapsed menu shows the full footer while the expanded menu shows the collapsed one.

This was inherited from the 1st blue-horizon implementation.
We can change it. Let's wait to see what UX proposes.
Maybe just adding some icon to the version output in the sidebar option would make things more coherent, but I don't know it this is a good practice

en3sis:
> Thanks @fabriziosestito ,
> The code looks good.
> In the other hand, I'm not that convinced of the output in the web, it looks a bit out of place (sepcially in the collapsed sidebar).
> I think it is worthy to comment this with the UX team.
> What do you thin @en3sis ?

Hey!

If all the information has to be visible (2nd screenshot) you could use only the page footer and not the menu one at all =D
You can refer to https://productbrand.suse.com/layout/footer for more information (and please let me know if any of that information it's confusing or it's not clear so we can re-visit and improve it)

fabriziosestito:
@arbulu89 @en3sis I think we should go just with the extended footer (second screenshot), as it looks better imho.
Where would you position the version?

arbulu89:
> @arbulu89 @en3sis I think we should go just with the extended footer (second screenshot), as it looks better imho.
> Where would you position the version?

Agree. I would only put it in the extended footer, and I like it personally the location and the way it looks

rtorrero (approved):
LGTM @fabriziosestito, thanks!

arbulu89 (approved):

PR #170, opened 2021-07-26
Fix sap instance status color code

### LINKED ISSUE #163: SAP Instance row in the single host view reports incorrect color code for the status
Looking at:
https://github.com/trento-project/trento/blob/01731e96d8d41d0cdff2df9f76c5746865e8f8a0/web/templates/blocks/sap_instance.html.tmpl#L32

We cannot make the assumption that `Running` is the only green status and all the others are red. For example, I saw that for some type of instances status would actually be `All processes running`.

I doubt we'll ever get to know all the possible values for this field, so I propose to remove the badge for the time being, or at most only color code the ones that are known to represent failures/warnings.

fabriziosestito:
Resolves #163

### DIFF
--- a/web/templates/blocks/sap_instance.html.tmpl
+++ b/web/templates/blocks/sap_instance.html.tmpl
@@ -28,8 +28,9 @@
                     <td>{{ .Elapsedtime }}</td>
                     <td>{{ .Pid }}</td>
                     <td>
-                        {{- $Textstatus := .Textstatus }}
-                        <span class='badge badge-pill badge-{{ if eq $Textstatus "Running" }}primary{{ else }}danger{{ end }}'>{{ $Textstatus }}</span>
+                        <span class='badge badge-pill badge-{{ if eq .Dispstatus "SAPControl-GREEN" }}primary{{ else if eq .Dispstatus "SAPControl-YELLOW" }}warning{{ else if eq .Dispstatus "SAPControl-RED" }}danger{{ else if eq .Dispstatus "SAPControl-GRAY" }}secondary{{ end }}'>
+                            {{ .Textstatus }}
+                        </span>
                     </td>
                 </tr>
             {{- else }}
--- a/web/hosts_test.go
+++ b/web/hosts_test.go
@@ -272,10 +272,32 @@ func TestHostHandler(t *testing.T) {
 								"Value": "HDB00",
 							},
 							"SAPSYSTEMNAME": map[string]interface{}{
-								"Value": "sys1",
+								"Value": "PRD",
 							},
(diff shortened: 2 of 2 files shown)


web/templates/blocks/sap_instance.html.tmpl
```diff
@@ -28,8 +28,11 @@
                     <td>{{ .Elapsedtime }}</td>
                     <td>{{ .Pid }}</td>
                     <td>
-                        {{- $Textstatus := .Textstatus }}
-                        <span class='badge badge-pill badge-{{ if eq $Textstatus "Running" }}primary{{ else }}danger{{ end }}'>{{ $Textstatus }}</span>
+                        {{- $Dispstatus := .Dispstatus }}
+                        {{ .Dispstatus }}
+                        <span class='badge badge-pill badge-{{ if eq $Dispstatus "SAPControl-GREEN" }}primary{{ else if eq $Dispstatus "SAPControl-YELLOW" }}warning{{ else if eq $Dispstatus "SAPControl-RED" }}danger{{ else if eq $Dispstatus "SAPControl-GRAY" }}secondary{{ end }}'>
+                            {{ $Dispstatus }}
```
arbulu89, line 34: I don't think we should print `$Dispstatus`, this is an internal reference. I still think that we should print the `$Textstatus`, which is the user friendly output (and as @stefanotorresi commented in the issue, the Dispstatus is a wrapper code to know the state, but each GREEN state can have different user outpus)

arbulu89 (changes requested):
Hi @fabriziosestito ,
Proposed a change. I think that we should print the `$Textstatus`, even using the `$Dispstatus` to get the badge type (which is a good idea)

web/templates/blocks/sap_instance.html.tmpl
```diff
@@ -28,8 +28,10 @@
                     <td>{{ .Elapsedtime }}</td>
                     <td>{{ .Pid }}</td>
                     <td>
-                        {{- $Textstatus := .Textstatus }}
-                        <span class='badge badge-pill badge-{{ if eq $Textstatus "Running" }}primary{{ else }}danger{{ end }}'>{{ $Textstatus }}</span>
+                        {{- $Dispstatus := .Dispstatus }}
```
arbulu89, line 31: I'm not sure if we need this assignment at the beginning here. It doesn't look like improve anything

arbulu89 (approved):
Thank you @fabriziosestito ,
I guess you checked the result and it makes sense (and the shown badge type and text are meaningful enough)

arbulu89 (approved):

stefanotorresi (approved):
gj!

PR #171, opened 2021-07-26
Display SID instead of pacemaker status

fabriziosestito:
Remove "Pacemaker status" from the cluster details view and replace it with SID


arbulu89 (approved):
Looks good

PR #172, opened 2021-07-27
Add the ansible playbooks metadata structure

arbulu89:
Implementation of the ansible playbook to store the checks metadata.
The description is added in the README file.

This playbook is executed only in localhost (the machine where the runner is being executing), so it doesn't involve any operations in the nodes themselves.

The execution will generate a ARA record, which will be consumed by trento.
The records looks like this:

image

@brett060102 could you have a look?


runner/ansible/roles/post-metadata/tasks/post.yml
```diff
@@ -0,0 +1,10 @@
+- name: Post metadata to ARA
+  ara_record:
+    key: trento-metadata
+    value: '{{ metadata }}'
+    type: json
+  register: record
+
+- name: Print metadata
+  debug:
+    msg: '{{ record.value }}'
```
brett060102, line 10: I would add:
verbosity: 1
here, so the debug message is only done when a "-v" or more is provided.

brett060102 (approved):
Looks great and it works. I have minor note on the debug message, but fine with the code as is.

runner/ansible/roles/post-metadata/tasks/post.yml
arbulu89: Thanks @brett060102,
Yes, it is a nice touch

PR #173, opened 2021-07-27
Add in rest of 1.1 tests

brett060102:
new file:   roles/1.1.2.runtime/tasks/main.yml
	new file:   roles/1.1.2/tasks/main.yml
	new file:   roles/1.1.3.runtime/tasks/main.yml
	new file:   roles/1.1.3/tasks/main.yml
	new file:   roles/1.1.4.runtime/tasks/main.yml
	new file:   roles/1.1.4/tasks/main.yml
	new file:   roles/1.1.5.runtime/tasks/main.yml
	new file:   roles/1.1.5/tasks/main.yml
	new file:   roles/1.1.6/tasks/main.yml
	new file:   roles/1.1.7/tasks/main.yml
	new file:   roles/1.1.8.runtime/tasks/main.yml
	new file:   roles/1.1.8/tasks/main.yml


runner/ansible/vars/azure_env.yml
```diff
@@ -4,3 +4,11 @@ expected:
   "1.1.1": "30000"
   "1.2.1": "true"
   "1.3.5": "60"
+  "1.1.2": "36000"
```
arbulu89, line 7: It think we should have the `ids` in numerical order, to make the reading easier, like it is in the `dev` vars file (I mean to put `1.1.2` before `1.2.1`

runner/ansible/vars/azure_env.yml
```diff
@@ -4,3 +4,11 @@ expected:
...
+  "1.1.3": "20"
+  "1.1.3_runtime": "20"
```
arbulu89, line 9: Do we need this `1.1.3_runtime` variable? It looks like a leftover

runner/ansible/check.yml
```diff
@@ -1,6 +1,12 @@
 - hosts: all
 
   tasks:
+    - name: require check mode
```
arbulu89, line 4: Nice touch! I like it

runner/ansible/roles/1.1.2.runtime/tasks/main.yml
```diff
@@ -0,0 +1,18 @@
+---
+- name: Set Test ID
+  set_fact:
+    test_id: '1.1.2_runtime'
```
arbulu89, line 4: I think we should have here `1.1.2.runtime` (see the `.` instead of the `_`), to have the same id of the role folder.
This should be changed in all the runtimes (including 1.1.1.runtime, which I didn't find in the previous review)

arbulu89 (changes requested):
Hi @brett060102 ,
It looks great in general.
I commented some items that we should fix.
Besides that, It is 99% ready to go

runner/ansible/vars/azure_env.yml
brett060102: Correct, will remove.

runner/ansible/vars/azure_env.yml
brett060102: agreed and I did mean to do that. thanks.

arbulu89 (approved):
Hi @brett060102 ,
Now it looks great!
Thanks. 
Ready to merge

PD: I wonder if there is some more ansiblenic way of running the regular expressions that is not `grep`... Anyway, I guess that most our users are comfortable enough to understand how to use grep hehe

PR #174, opened 2021-07-29
Add 1.2.2 and remaining 1.3 tests

brett060102:
This adds:
1.2.2
1.3.1
1.3.2
1.3.3
1.3.4
1.3.6
1.3.7

1.3.7 is  little funky. The counting method use on the sample rule can't be handled by the jinja2 handling in ansible. So, I needed to come up with a different  way to get the count.


arbulu89 (changes requested):
Hi @brett060102 ,
Redirect the PR to `runner`. You can do this clicking the `edit` button.
**Ok, did it myself**

runner/ansible/roles/1.2.2/tasks/main.yml
```diff
@@ -0,0 +1,30 @@
+---
+- name: Set Test ID
+  set_fact:
+    test_id: '1.2.2'
+
+- name: "{{ test_id }}.check"
+  shell: |
+   timeout=$(crm_attribute -t crm_config -G -n stonith-timeout --quiet)
+          if [[cibadmin -Q --xpath "//primitive[@type='fence_azure_arm']/@type" > /dev/null 2>&1 ]]; then
```
arbulu89, line 9: Please, indent properly the script code, it is difficult to read otherwise
(try to do this in all the new shell code tasks)

runner/ansible/roles/1.2.2/tasks/main.yml
```diff
@@ -0,0 +1,30 @@
...
+            if [[ "${timeout}" = "900" ]]; then
```
arbulu89, line 10: Here, in order of having `900`, this should go in the expected values map. `900` is the expected value for the scenario where we have the `fence_azure_arm`, and `144` for sbd.
We would need most probably 2 expected values, 1 for `fence_azure_arm` and other for `sbd`. Later on, you can just compare the `timeout` value. You don't need to do this `correct` outputting. This was done this way due a limitation of benchmark

runner/ansible/roles/1.3.1/tasks/main.yml
```diff
@@ -0,0 +1,19 @@
+---
+
+- name: Set Test ID
+  set_fact:
+    test_id: '1.3.1'
+
+- name: "{{ test_id }}.check"
+  shell: 'grep SBD_PACEMAKER /etc/sysconfig/sbd'
```
arbulu89, line 8: We should go back to the `replace` option, like in `1.1.1` test, as this is a file checking

runner/ansible/roles/1.3.2/tasks/main.yml
```diff
@@ -0,0 +1,19 @@
+---
+
+- name: Set Test ID
+  set_fact:
+    test_id: '1.3.2'
+
+- name: "{{ test_id }}.check"
+  shell: 'grep SBD_STARTMODE /etc/sysconfig/sbd'
```
arbulu89, line 8: The same as the previous one

runner/ansible/roles/1.3.3/tasks/main.yml
```diff
@@ -0,0 +1,19 @@
+---
+
+- name: Set Test ID
+  set_fact:
+    test_id: '1.3.3'
+
+- name: "{{ test_id }}.check"
+  shell: 'systemctl is-enabled sbd || true'
```
arbulu89, line 8: Could we use the `systemd` module and the `check` mode here?
https://docs.ansible.com/ansible/latest/collections/ansible/builtin/systemd_module.html

runner/ansible/roles/1.3.4/tasks/main.yml
```diff
@@ -0,0 +1,23 @@
+---
+
+- name: Set Test ID
+  set_fact:
+    test_id: '1.3.4'
+
+- name: "{{ test_id }}.check"
+  shell: |
+    echo -n "SBD_COUNT="
```
arbulu89, line 9: What is this first `echo` for?
Here the purpose is to test if we have 3 (expected value) SBD devices configured. 
For that, it checks the `/etc/sysconfig/sbd` file, and withing this the `SBD_DEVICE` entry.
This entry can be like this:
`SBD_DEVICE=/dev/vdb;/dev/vdc;/dev/ddd` (the value can go even within quotes, and have spaces after `;` value).
So in this case, we would count 3 (split the value with `;`)

runner/ansible/roles/1.3.6/tasks/main.yml
```diff
@@ -0,0 +1,32 @@
+---
+
+- name: Set Test ID
```
arbulu89, line 3: This test must be revisited.
Check the original one:
https://github.com/trento-project/trento/blob/main/examples/generic-azure.yaml

This check doesn't have an expected value. It checks if the `wdtimeout` has the double value of `msgwait`.
Maybe the test was updated since you implemented this, but it looks really different.
**In fact, it looks like you implemented here the 1.3.5 test*

runner/ansible/roles/1.3.7/tasks/main.yml
```diff
@@ -0,0 +1,33 @@
+---
+
+- name: Set Test ID
```
arbulu89, line 3: This looks outdated too:
Check: https://github.com/trento-project/trento/blob/main/examples/generic-azure.yaml#L496

arbulu89 (changes requested):
Hi @brett060102 ,
Comments added

runner/ansible/roles/1.2.2/tasks/main.yml
brett060102: will do.

runner/ansible/roles/1.2.2/tasks/main.yml
brett060102: Will work this one out with you in chat.

runner/ansible/roles/1.3.1/tasks/main.yml
brett060102: I think we need to go with line in file for this one. But will make that change for 1.3.1 and 1.3.2

runner/ansible/roles/1.3.3/tasks/main.yml
brett060102: good idea. I will give it a shot.

runner/ansible/roles/1.3.4/tasks/main.yml
brett060102: The audit for this one from the example is:
        audit: |
          sbdarray=$(grep -E '^SBD_DEVICE=' /etc/sysconfig/sbd  | grep -oP 'SBD_DEVICE=\K[^.]+' | sed 's/\"//g')
          IFS=';' sbdarray=( $sbdarray )
          echo "SBD_COUNT=${#sbdarray[@]}"
        tests:
          test_items:
            - flag: "SBD_COUNT"
              compare:
                op: eq
                value: 3

So this translated to test for 
"SBD_COUNT=3"
which is what is in the vars/azure_env.yml 

I will need to use the counting hack in 1.3.7 here to get the count right. "${#" just messes up jinja in  ansible.

runner/ansible/roles/1.3.6/tasks/main.yml
brett060102: it needs the following:
DEF_MSGWAIT={{ expected[test_id] }}

expected 1.3.5 is 60 and   expected for 1.3.6 is 120  but it look like 1.3.6 is always checking msgwait and 1.3.5 watchdog.

runner/ansible/roles/1.3.7/tasks/main.yml
brett060102: should be OK now

runner/ansible/roles/1.3.7/tasks/main.yml
```diff
@@ -0,0 +1,39 @@
...
+  set_fact:
+    test_id: '1.3.7'
+
+- name: "{{ test_id }}.check"
+  shell: |
+    if [[ $(crm_node -l | wc -l) != "2" ]]; then
+      echo passed
```
arbulu89, line 10: I would prefer to remove this `passed/failed` echos and use the `rc` to check the result.
This way, we could remove the expected value `passed`, which doesn't provide any utility.
We could put an empty string in the vars file, or remove it all together, as this test doesn't have a changeable expected value.
(I would vote for removing)

arbulu89 (changes requested):
Hi @brett060102 ,
Nice, we are much closer.
I commented more smaller things that are worth to fix or discuss at least

runner/ansible/roles/1.2.2/tasks/main.yml
```diff
@@ -0,0 +1,25 @@
...
+   timeout=$(crm_attribute -t crm_config -G -n stonith-timeout --quiet)
+   if [[cibadmin -Q --xpath "//primitive[@type='fence_azure_arm']/@type" > /dev/null 2>&1 ]]; then
+     exit $([[ "${timeout}" = "{{ expected['1.2.2.fence_azure_arm'] }}" ]])
+   else
+     exit $([[ "${timeout}" = "{{ expected['1.2.2.sbd'] }}" ]])
+   fi
+  check_mode: no
+  register: config_updated
+  changed_when: config_updated.rc != 0
+  failed_when: config_updated.rc > 1
```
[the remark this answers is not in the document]
brett060102: I guess that  cosmetics depends on your point of view. Without failed_when: ansible will report a failed state and by default will terminate the run. We can ingore the "failure" which will let the play continue. But our convention currently is for failing tests to set changed status.

runner/ansible/roles/1.2.2/tasks/main.yml
```diff
@@ -0,0 +1,25 @@
+---
+- name: Set Test ID
+  set_fact:
+    test_id: '1.2.2'
+
+- name: "{{ test_id }}.check"
+  shell: |
+   timeout=$(crm_attribute -t crm_config -G -n stonith-timeout --quiet)
+   if [[cibadmin -Q --xpath "//primitive[@type='fence_azure_arm']/@type" > /dev/null 2>&1 ]]; then
+     exit $([[ "${timeout}" = "{{ expected['1.2.2.fence_azure_arm'] }}" ]])
```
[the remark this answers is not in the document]
brett060102: Thank you. I banged my head against the wall on this one for while before giving up.

runner/ansible/roles/1.3.3/tasks/main.yml
```diff
@@ -0,0 +1,21 @@
+---
+
+- name: Set Test ID
+  set_fact:
+    test_id: '1.3.3'
+
+- name: "{{ test_id }}.check"
+  systemd:
```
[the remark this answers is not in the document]
brett060102: yes, it does work.

runner/ansible/roles/1.3.4/tasks/main.yml
```diff
@@ -0,0 +1,26 @@
...
+
+- name: Set Test ID
+  set_fact:
+    test_id: '1.3.4'
+
+- name: "{{ test_id }}.check"
+  shell: |
+    sbdarray=$(grep -E '^SBD_DEVICE=' /etc/sysconfig/sbd  | grep -oP 'SBD_DEVICE=\K[^.]+' | sed 's/\"//g')
+    IFS=';' sbdarray=( $sbdarray )
+    # convoluted, but normal count method does not work with jinja2
```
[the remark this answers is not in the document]
brett060102: This is bug report on this issue that is where I got the work around. 
https://github.com/ansible/ansible/issues/16968
I don't know what the root issue is but I could not find another solution. Using {% raw %} did not help.

runner/ansible/roles/1.3.4/tasks/main.yml
```diff
@@ -0,0 +1,26 @@
...
+    temp_ar=(${!sbdarray[@]});  device_count=`expr ${temp_ar[-1]} + 1`
+    echo "SBD_COUNT=$device_count"
```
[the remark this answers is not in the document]
brett060102: yes

runner/ansible/roles/1.3.6/tasks/main.yml
```diff
@@ -0,0 +1,32 @@
...
+    DEF_MSGWAIT={{ expected[test_id] }}
+    result_msgwait=${DEF_MSGWAIT}
+    sbdarray=$(grep -E '^SBD_DEVICE=' /etc/sysconfig/sbd  | grep -oP 'SBD_DEVICE=\K[^.]+' | sed 's/\"//g')
+    IFS=';' sbdarray=( $sbdarray )
+    for i in "${sbdarray[@]}"
+    do
+      msgwait=$(/usr/sbin/sbd -d ${i} dump | grep -oP 'Timeout \(msgwait\)  *: \K\d+')|| echo ""
+      if [[ "${msgwait}" -ne "${DEF_MSGWAIT}" ]]; then
+        result_msgwait="${msgwait}"
+        fi
```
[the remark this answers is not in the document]
brett060102: sorry, old style habits die hard.

runner/ansible/vars/dev_env.yml
```diff
@@ -10,4 +10,12 @@ expected:
   "1.1.7": "2"
   "1.1.8": "1"
   "1.2.1": "true"
+  "1.2.2.sbd": ""
```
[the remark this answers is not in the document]
brett060102: It will cause the test to fail in our dev env.

runner/ansible/roles/1.2.2/tasks/main.yml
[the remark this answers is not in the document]
arbulu89: Ok, so it is not cosmetics. Got it.
So what will happen when we have `config_updated.rc > 1` ?
Because this is a possible scenario, when the check fails

runner/ansible/roles/1.3.4/tasks/main.yml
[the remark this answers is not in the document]
arbulu89: Could we put this issue link as comment there?

runner/ansible/vars/azure_env.yml
```diff
@@ -10,4 +10,11 @@ expected:
   "1.1.7": "2"
   "1.1.8": "1"
   "1.2.1": "true"
+  "1.2.2.sbd": "144"
+  "1.2.2.fence_azure_arm": "900"
+  "1.3.1": "yes"
+  "1.3.2": "always"
+  "1.3.3": "enabled"
```
arbulu89, line 17: We don't really need this `1.3.3` expected value, right?

runner/ansible/vars/dev_env.yml
```diff
@@ -10,4 +10,11 @@ expected:
...
+  "1.2.2.fence_azure_arm": ""
+  "1.3.1": "yes"
+  "1.3.2": "always"
+  "1.3.3": "enabled"
```
arbulu89, line 17: Same here with this 1.3.3 value

arbulu89 (changes requested):
Hey @brett060102 
Almost there, I requested some small changes

runner/ansible/vars/azure_env.yml
brett060102: No we do not.

runner/ansible/roles/1.3.1/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
+---
+
+- name: Set Test ID
+  set_fact:
+    test_id: '1.3.1'
+
+- name: "{{ test_id }}.check"
+  lineinfile:
```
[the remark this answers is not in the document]
brett060102: yes  and it will insert line if not found.

runner/ansible/vars/dev_env.yml
brett060102: yep

arbulu89 (approved):
Thanks @brett060102 
Great work

PR #177, opened 2021-07-29
Make it compatible with node_modules obs service

aleksei-burlakov:
DRAFT!

To enable compiling trento in the obs I introduced two changes.
1) The obs requires the package.json file in the root of the source folder. So I copied it from "./web/frontend" to "./". This is obviously repeating and not good.
2) I have updated the package-lock.json file. (using $ npm install)

OBS Ref: https://api.opensuse.org/package/show/devel:sap:trento/trento


rtorrero:
Thanks @aleksei-burlakov !

2 questions:

Would it work if we leave the `package.json` file in the original location and also use the OBS `download_service` to fetch it?

I saw you are using the `download_service` to fetch the `package-lock.json`. That looks good to me, but wouldn't this mean that we can remove the `package-lock.json` from the OBS repo? (since it's downloaded everytime from here?)

aleksei-burlakov:
> Thanks @aleksei-burlakov !
> 
> 2 questions:
> 
> Would it work if we leave the `package.json` file in the original location and also use the OBS `download_service` to fetch it?

The download_service downloads files into the root of the project not into the root of the source (the tarball). I think it's possible to do it, like download the source from git, download the package.json in the source root, and then recompress the source, but I don't know.

> I saw you are using the `download_service` to fetch the `package-lock.json`. That looks good to me, but wouldn't this mean that we can remove the `package-lock.json` from the OBS repo? (since it's downloaded everytime from here?)

That's a good idea. I thought since the package-lock.json is listed in the sources list of the spec file it must exist in the project, but actually not. The system didn't complain.

aleksei-burlakov:
I have changed the PR. There are 2 commits and they propose two different approaches, i.e. the second commit doesn't accomplish the first, but reverts the first and gives another solution.

The **first** solution (imperative) is to copy the package.json from the root to the web/frontend and do all as before.
The **second** solution (declarative) is to use .npmrc that should eventually put the node_modules in the web/frontend. It's better than the first solution, we shouldn't move around the package.json file, but it doesn't work. I'd be glad to here what's wrong with it))

aleksei-burlakov:
I'm closing the issue. The trento obs project was changed so that and there is no in changing the trento sources.

PR #179, opened 2021-07-30
SAP Systems list revamp

fabriziosestito:
SAP Systems list revamp, now we only list the application sap systems and the related application instances
Still missing: Health container & filters (under discussion), tags

A new viewmodel sap systems table was created to simplify the KV store data processing.

### DIFF
--- a/web/environments.go
+++ b/web/environments.go
@@ -1,11 +1,15 @@
 package web
 
 import (
+	"fmt"
 	"net/http"
 
 	"github.com/gin-gonic/gin"
+	"github.com/trento-project/trento/internal"
 	"github.com/trento-project/trento/internal/consul"
 	"github.com/trento-project/trento/internal/environments"
+	"github.com/trento-project/trento/internal/hosts"
+	"github.com/trento-project/trento/internal/sapsystem"
 )
 
 func NewEnvironmentListHandler(client consul.Client) gin.HandlerFunc {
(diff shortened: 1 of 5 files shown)


fabriziosestito:
image

web/environments.go
```diff
@@ -102,29 +106,140 @@ func NewLandscapeHandler(client consul.Client) gin.HandlerFunc {
...
+	}
+
+	for _, row := range rowsBySID {
+		sapSystemsTable = append(sapSystemsTable, row)
+	}
+
+	return sapSystemsTable
+}
+
+func (t SAPSystemsTable) filter(sid []string) SAPSystemsTable {
```
arbulu89, line 184: Maybe `filterBySid` would be a better name

web/templates/sapsystems.html.tmpl
```diff
@@ -1,10 +1,94 @@
...
+                                <th scope='col'>Cluster name</th>
+                                <th scope='col'>Host</th>
+                            </tr>
+                            </thead>
+                            <tbody>
+                            {{- range .InstancesTable }}
+                            <tr>
+                                <td class="row-status">{{ template "health_icon" "unknown" }}</td>
+                                <td>
+                                    <a href="/sapsystems/{{ .SID }}?environment=default&landscape=default">{{ .SID }}</a>
```
arbulu89, line 76: We will need to be careful with this `SID` link (in the inner table)
This SID could be HANA installation SID, which is not a SAP system. 
We would need to see what to do. 
Maybe, we should remove the link from the internal table

arbulu89 (approved):
Good work!

web/environments.go
fabriziosestito: we will need to filter by health and possibly tag in the near future, i would leave it as `filter` also to match the `ClustersTable` filter function

web/templates/sapsystems.html.tmpl
fabriziosestito: if we remove the link and the HANA SID is different from the outer row, we would not have a link to the HANA SID

arbulu89: Yes, I know. What this page should redirect to SAP systems, and a HANA installation is not a full SAP system.
I don't know if we will have pages for individual SAP instances or databases honestly

fabriziosestito: Ok I was missing this context, i will remove the link in the inner rows then

arbulu89 (approved):

PR #180, opened 2021-07-30
Add 1.4.*, 1.5.* and 1.1.6 tests for the ansible runner

brett060102:
add 1.4.* tests
add 1.5.* tests
add 1.1.6.runtime test
1.2.1 change task name to be .check to match other tests.


runner/ansible/roles/1.4.2/tasks/main.yml
```diff
@@ -0,0 +1,30 @@
...
+- name: Set Test ID
+  set_fact:
+    test_id: '1.4.2'
+
+- name: "{{ test_id }}.check"
+  shell: |
+    cf_status=$(crm_attribute -t crm_config -G -n concurrent-fencing --quiet)
+    if cibadmin -Q --xpath "//primitive[@type='fence_azure_arm']/@type" > /dev/null 2>&1; then
+        if [[ ${cf_status} = "true" ]]; then
+            echo "correct"
```
arbulu89, line 12: I would prefer to use return codes here.
The `correct` word is just a hardcoded value, that doesn't add any value.
In this case, this would be a correct return code, the next `else` would be an error one, and the 3rd one a correct one again

runner/ansible/vars/azure_env.yml
```diff
@@ -17,3 +17,13 @@ expected:
   "1.3.4": "3"
   "1.3.5": "60"
   "1.3.6": "120"
+  "1.4.1.monitor_interval": "3600"
+  "1.4.1.monitor_timeout": "120"
+  "1.4.1.pcmk_action_limit": "3"
+  "1.4.1.pcmk_monitor_retries": "4"
+  "1.4.1.pcmk_reboot_timeout": "900"
+  "1.4.1.power_timeout": "240"
+  "1.4.2": "correct OR fence_azure_arm not configured"
```
arbulu89, line 26: I would remove this `1.4.2` as it doesn't add any value, and we don't want the user changing this value, as it is custom predefined value

runner/ansible/vars/azure_env.yml
```diff
@@ -17,3 +17,13 @@ expected:
...
+  "1.5.1": "pass"
```
arbulu89, line 27: We don't need this `1.5.1` entry, we don't use it

runner/ansible/roles/1.5.2/tasks/main.yml
```diff
@@ -0,0 +1,39 @@
+---
+
+- name: Set Test ID
+  set_fact:
+    test_id: '1.5.2'
+
+- name: "{{ test_id }}.check"
+  shell: |
```
arbulu89, line 8: Cannot we use the https://docs.ansible.com/ansible/latest/collections/ansible/builtin/user_module.html module to test this?
Basically, we want to check if the `hacluster` user doesn't have `linux` as password, as this is a signal of a security hole

runner/ansible/roles/1.5.3/tasks/main.yml
```diff
@@ -0,0 +1,27 @@
...
+
+- name: Set Test ID
+  set_fact:
+    test_id: '1.5.3'
+
+- name: "{{ test_id }}.check"
+  shell: |
+          # The third field of ntpq -p is the stratum and if it's 16
+          # the sync is disabled. Make sure it's not 16
+          if ! sudo ntpq -pn | grep "No association ID's returned" \
```
arbulu89, line 11: Do we need to have `sudo` here?
I thought ansible would execute this already with root rights

runner/ansible/vars/azure_env.yml
```diff
@@ -17,3 +17,13 @@ expected:
...
+  "1.5.2": "blocked"
```
arbulu89, line 28: In fact, we don't need any of these `1.4.2`, `1.5.1`, `1.5.2` and `1.5.3` results here

runner/ansible/roles/1.4.1.monitor_interval/tasks/main.yml
```diff
@@ -0,0 +1,25 @@
+---
+
+- name: Set Test ID
+  set_fact:
+    test_id: '1.4.1.monitor_interval'
+
+- name: "{{ test_id }}.check"
+  shell: |
+    if cibadmin -Q --xpath "//primitive[@type='fence_azure_arm']/@type" > /dev/null 2>&1; then
```
arbulu89, line 9: Cannot we do something to have all these `1.4.1` tests as an unique test?
This will make things really hard to work with, because they would be treated as individual tests, something we don't really like to do.
I would try to run all the commands here and compare them with the expected results or something.
It should be doable with shell,
Maybe, store the return code of the comparisons in an array, and check eventually if there is some incorrect value.

runner/ansible/roles/1.4.1.monitor_interval/tasks/main.yml
```diff
@@ -0,0 +1,25 @@
...
+        cibadmin -Q --xpath "//primitive[@type='fence_azure_arm']/operations/op [@name='monitor']" | grep -oP 'interval="\K[^"]+' || echo ""
+    else
+        # Azure Fence Agent is not configured. We echo expected values just to fulfill the check
+        echo 3600
```
arbulu89, line 13: This `3600` should we `expected[test_id]` I guess

arbulu89 (changes requested):
Hi @brett060102 ,
My comments below.
In general, I don't like the approach of splitting the 1.4.1 test is subtests, it makes things harder to understand how to pack the result in the frontend. I think that we can put them all at one test.
Unless we are fine making them as independent tests, but we would need to change the identifier for something better (maybe adding a letter at the end). What do you think @stefanotorresi ?

runner/ansible/roles/1.4.2/tasks/main.yml
aleksei-burlakov: To my mind, it can as well be the "correct" string as long as it's consistent. We use however different string literals in the checks, so maybe, it'll be easier to implement with numeric codes.

runner/ansible/roles/1.5.2/tasks/main.yml
aleksei-burlakov: > Cannot we use the https://docs.ansible.com/ansible/latest/collections/ansible/builtin/user_module.html module to test this?
It looks to me like if we could, but let's wait for an answer from an ansible expert.

runner/ansible/roles/1.5.3/tasks/main.yml
aleksei-burlakov: If it's all run as root, there is no need in sudo here of course.

runner/ansible/roles/1.4.2/tasks/main.yml
brett060102: I can go with a return code here.

runner/ansible/roles/1.4.1.monitor_interval/tasks/main.yml
brett060102: As per our talk. I will cllaps all 1,4.1 tests into a single test with pass/fail return.

runner/ansible/roles/1.5.3/tasks/main.yml
brett060102: We cannot/should not depend on this being run as the root user.

runner/ansible/vars/azure_env.yml
brett060102: Correct, I should have pulled that line.

runner/ansible/vars/azure_env.yml
brett060102: OK. Will we work 1.4.2 to use exit status and fence_azure_arm not configured will be a pass.

runner/ansible/roles/1.4.2/tasks/main.yml
brett060102: Have moved to an RC for this one.

runner/ansible/roles/1.4.1/tasks/main.yml
```diff
@@ -0,0 +1,30 @@
...
+    test_id: '1.4.1'
+
+- name: "{{ test_id }}.check"
+  shell: |
+    # if no Azure Fence Agent exit, with a pass
+    cibadmin -Q --xpath "//primitive[@type='fence_azure_arm']/@type" > /dev/null 2>&1 || exit 0
+    # Check that following match expected, exit with error if any do not
+    [[ $(cibadmin -Q --xpath "//primitive[@type='fence_azure_arm']/instance_attributes/nvpair[@name='pcmk_monitor_retries']" | grep -oP 'value="\K[^"]+') != "{{ expected[test_id + '.pcmk_monitor_retries'] }}" ]] && exit 1
+    [[ $(cibadmin -Q --xpath "//primitive[@type='fence_azure_arm']/instance_attributes/nvpair[@name='pcmk_action_limit']" | grep -oP 'value="\K[^"]+') != "{{ expected[test_id + '.pcmk_action_limit'] }}" ]] && exit 1
+    [[ $(cibadmin -Q --xpath "//git ststusprimitive[@type='fence_azure_arm']/instance_attributes/nvpair[@name='power_timeout']" | grep -oP 'value="\K[^"]+') != "{{ expected[test_id + '.power_timeout'] }}" ]] && exit 1
```
diegoakechi, line 14: I believe this `git ststus` was miss-typed here.

arbulu89: Besides this typo, the check looks good though

runner/ansible/vars/azure_env.yml
```diff
@@ -17,3 +17,10 @@ expected:
   "1.3.4": "3"
   "1.3.5": "60"
   "1.3.6": "120"
+  "1.4.1.monitor_interval": "3600"
+  "1.4.1.monitor_timeout": "120"
+  "1.4.1.pcmk_action_limit": "3"
+  "1.4.1.pcmk_monitor_retries": "4"
+  "1.4.1.pcmk_reboot_timeout": "900"
+  "1.4.1.power_timeout": "240"
+  "1.4.2": "true"
```
arbulu89, line 26: We can remove this `1.4.2`. It is not used

arbulu89 (commented):
Hi @brett060102 ,

The check `1.4.1` looks good now (except the type).
Let's remove the test `1.5.3` by now, so we can close this PR as soon as everything else is fixed

runner/ansible/roles/1.5.2/tasks/main.yml
arbulu89: @brett060102 I have done some research and this option looks a bit nice:
(source: https://askubuntu.com/questions/611580/how-to-check-the-password-entered-is-a-valid-password-for-this-user)

```
salt=$(sudo getent shadow hacluster | cut -d$ -f3)
epassword=$(sudo getent shadow hacluster | cut -d: -f2)
match=$(python3 -c 'import crypt; print(crypt.crypt("linux", "$6$'${salt}'"))')
exit [ ${match} == ${epassword} ]
```

Could you give a try?
PD:
crypt is a native package in python3: https://docs.python.org/3.6/library/crypt.html

runner/ansible/roles/1.4.1/tasks/main.yml
brett060102: Thanks

runner/ansible/vars/azure_env.yml
brett060102: No, it is used now:
    [[ $(crm_attribute -t crm_config -G -n concurrent-fencing --quiet) = "{{ expected[test_id] }}" && exit 0

runner/ansible/roles/1.5.2/tasks/main.yml
brett060102: OK, I new get what you are trying to do. I will take a look.

runner/ansible/roles/1.4.2/tasks/main.yml
```diff
@@ -0,0 +1,24 @@
+---
+
+- name: Set Test ID
+  set_fact:
+    test_id: '1.4.2'
+
+- name: "{{ test_id }}.check"
+  shell: |
+    cibadmin -Q --xpath "//primitive[@type='fence_azure_arm']/@type" > /dev/null 2>&1 || exit 0
+    [[ $(crm_attribute -t crm_config -G -n concurrent-fencing --quiet) = "{{ expected[test_id] }}" && exit 0
```
arbulu89, line 10: Is this working? I miss the closing brackets (`]]`)

runner/ansible/roles/1.5.2/tasks/main.yml
arbulu89: Nice!

arbulu89 (changes requested):
Hi @brett060102 ,
Good work.
We need to fix the `1.4.2` and the missing `]]`.
Once this is fixed we can merge

runner/ansible/roles/1.4.2/tasks/main.yml
brett060102: fixed and really can't imagine how I toasted that one. should have stopped earlier.

arbulu89 (approved):
Thanks @brett060102 
Ready to go!

PR #181, opened 2021-08-02
Implement the cluster settings modal for checks selection

arbulu89:
Check selection modal for the cluster.

This modal allows the user to store the checks that are performed for this particular cluster.
The checks catalog is coming from ARA, from the playbook executed with the `meta` label, and the ARA record stored by this.

Within the modal, the user can select these checks, that will be stored in the KV storage once the `Save` button is clicked.
This selection will modify the ansible inventory to enable only this checks.

In order to store the checks selection, I have moved the discovered cluster data to `discovered_data` sub directory in the KV value, otherwise everything within the cluster was being removed every discovery loop, and we need to have persistance storage for the checks selection.

Besides that, this PR brings a Flash alerting system. Check this commit to review it independently: https://github.com/trento-project/trento/pull/181/commits/99703d3ab1e4a54fae94c388001353b2d8c38312

So summarazing:
- Add the new modal
- Checks catalog is coming from ARA, from the metadata playbook
- Check selected checks are stored in the KV under `clusters/$clustername/user_settings/checks` as a string (`1.1.1,1.1.2`, means that the checks 1.1.1 and 1.1.2 are enabled)
- The selected checks are used to modify the ansible inventory to add these selection as a variable, which is used to filter the executed checks per cluster
- Alerting system using the classical flash messages, and the POST-REDIRECT mechanism

A short demo:
cluster_modal

KV storage:
image


web/alert.go
```diff
@@ -0,0 +1,67 @@
...
+	var alerts []*Alert
+
+	for _, alertI := range f {
+		alert, _ := alertStrToStruct(alertI.(string))
+		alerts = append(alerts, alert)
+	}
+	session.Save()
+
+	return alerts
+}
```
fabriziosestito, line 67: We can avoid the string encoding and splitting by using the struct directly in `session.AddFlash`,
you just need to register it:

``` go
func StoreAlert(c *gin.Context, a *Alert) {
	gob.Register(Alert{})

	session := sessions.Default(c)
	session.AddFlash(a, AlertsKey)
	session.Save()
}

func GetAlerts(c *gin.Context) []*Alert {
	gob.Register(Alert{})

	session := sessions.Default(c)
	f := session.Flashes(AlertsKey)
	var alerts []*Alert

	for _, alertI := range f {
		alert, _ := alertI.(Alert)
		fmt.Println(alert)
		alerts = append(alerts, &alert)
	}
	session.Save()

	return alerts
}
```

web/clusters.go
```diff
@@ -549,12 +502,30 @@ func NewClusterHandler(client consul.Client) gin.HandlerFunc {
...
 		})
 	}
 }
+
+type checkSelectionForm struct {
+	Ids []string `form:"ids[]"`
+}
+
+func NewPostClusterHandler(client consul.Client) gin.HandlerFunc {
+	return func(c *gin.Context) {
```
fabriziosestito, line 521: Maybe i would call it SaveChecks handler

web/app.go
```diff
@@ -70,7 +70,8 @@ func NewAppWithDeps(host string, port int, deps Dependencies) (*App, error) {
 	engine.GET("/hosts/:name/ha-checks", NewHAChecksHandler(deps.consul))
 	engine.GET("/catalog", NewChecksCatalogHandler(deps.checksService))
 	engine.GET("/clusters", NewClusterListHandler(deps.consul))
-	engine.GET("/clusters/:id", NewClusterHandler(deps.consul))
+	engine.GET("/clusters/:id", NewClusterHandler(deps.consul, deps.checksService))
+	engine.POST("/clusters/:id", NewPostClusterHandler(deps.consul))
```
fabriziosestito, line 79: maybe i would change the path to `/clusters/:id/checks`

fabriziosestito (approved):
GJ!

web/alert.go
arbulu89: Nice! Man, I was trying to do this almost the whole day, and finally ended up creating this encoding system..
Changing right away!

web/clusters.go
arbulu89: Ok, I like it

web/app.go
arbulu89: Yes, better. We might have other posts most probably for this cluster

arbulu89:
@fabriziosestito Issues fixed in the last 2 commits:
https://github.com/trento-project/trento/pull/181/commits/d54777c46e014f991b75ef49095d1e075058f56b
https://github.com/trento-project/trento/pull/181/commits/946df4e2754da7b976adc9de04562248885d4837

Thank you for the review!

internal/cluster/check_selection_kvstore.go
```diff
@@ -0,0 +1,48 @@
...
+
+	"github.com/pkg/errors"
+
+	"github.com/trento-project/trento/internal/consul"
+)
+
+func GetCheckSelection(client consul.Client, clusterId string) (string, error) {
+	err := client.WaitLock(consul.KvClustersPath)
+	if err != nil {
+		return "", errors.Wrap(err, "error waiting for the lock for clusters")
```
aleksei-burlakov, line 14: Empty strings? 😏

arbulu89: What should I return?
It we have an error the error message will come from the `err` parameter.
If you are concerned about the usage should be like this: https://github.com/trento-project/trento/blob/cbd750177af363763d8b7aee7b4d0d9606d8c857/agent/discovery/cloud.go#L37

I don't think it is needed. The error message is already coming in the error in your case, so it is actually duplicated

fabriziosestito: This is pretty idiomatic since the empty string is the blank value of the string type, if the err is not nil the string is going to be ignored from the code that handles the error

PR #182, opened 2021-08-02
CI: fetch tags before building trento

### LINKED ISSUE #178: Demo environment does not display the version
The script assumes that tags are fetched, while the CI action does not fetch them, so the version number is not injected.
see: https://github.com/actions/checkout/issues/206

fabriziosestito:
Resolves #178


stefanotorresi (approved):
lgtm

PR #184, opened 2021-08-02
Add deployment to netweaver nodes

fabriziosestito:
Add netweaver nodes to the CI.
The job was repeated with the netweaver nodes tags, until we find a better solution.


rtorrero (approved):
looks amazing to me

PR #191, opened 2021-08-05
Add new documentation about the runner component

arbulu89:
New entry in the README file about the usage of the trento runner.

I would appreciate if anyway follows the quickstart to see if it works fine


rtorrero:
Hey @arbulu89, I think you forgot to update the index on the top of the document :-) Appart from that, lgtm

rtorrero (dismissed):
We should probably add a index entry in line 29

README.md
```diff
@@ -170,12 +170,42 @@ To start the trento agent:
...
+Trento Runner is responsible for running the health checks. It is based on [Ansible](https://docs.ansible.com/ansible/latest/index.html) and [ARA](https://ara.recordsansible.org/).
+This component can be executed in the same machine as the Web UI, but it is not mandatory, it can be executed in any other machine that has network access to the agents.
+
+In order to start it, some packages must be installed and started. Here a quick go through:
+```shell
+# Install ansible. Tested with version 2.11.2
+pip install ansible
+# Install ARA
+pip install --user ansible "ara[server]"
+# Setup ARA database
```
fabriziosestito, line 184: Is installing ansible again in the user directory needed?

fabriziosestito (commented):
LGTM, I left just some comments about the ansible/ara installation process.

Also, I tried to follow the steps to setup a k3s /helm chart installation of the runner and it looks good!

README.md
```diff
@@ -170,12 +170,42 @@ To start the trento agent:
 > recommendations on cluster settings and state for many HA components. New rules
 > can be created and tuned to adjust to different requirements.
 
+## Trento Runner
+
```
fabriziosestito, line 174: The section assumes that user is going to run the runner and ARA on the same machine, this is not always true. When I tried to replicate the installation on k3s i realized that my trento runner container was missing the python3 `ara.setup.*` dependencies. Could we split the section in two, one for the ARA setup and one for the trento runner? Also, is installing the `ara[server]` dep needed on the trento runner machine/container, or we can install just the `ara` module without the server deps?

README.md
```diff
@@ -170,12 +170,42 @@ To start the trento agent:
...
+# Install ARA
+pip install --user ansible "ara[server]"
+# Setup ARA database
+ara-manage migrate
+# Start ARA server. This process can be started in background or in other shell terminal
+ara-manage runserver ip:port
+
+# Enable the SSH connections to the agents machines
+# Update the /etc/hosts file to identify the agents hosts
+# Authorize SSH connection for all the agents hosts
```
fabriziosestito, line 191: I am a bit confused about this part, even if it works fine by adding the entries to the /etc/hosts files I wouldn't ask the user to do so. We already have a service discovery in place, we might want to use that. The consul server DNS should be  possibly used to resolve the nodes.

README.md
arbulu89: Let me check this. I followed the ARA installation guides, but maybe this repeats the installation twice.
What I thought it does is to install ara under the ansible user, which is something required by it.
I will double check

arbulu89: Double checked and you were right @fabriziosestito , the installation is redundant. The source of the documentation by the way: https://ara.readthedocs.io/en/latest/getting-started.html

README.md
arbulu89: I will add new content to make clear that the ARA server can go in another machine.
And I will have a look about the ARA dependencies when the 2 components are running in different locations

README.md
arbulu89: The code updated to work without editing the host file.
I will update the comments here to make sure that the unique mandatory part is the ssh authorization

arbulu89:
@rtorrero @fabriziosestito Content updated with your suggestions

arbulu89:
@fabriziosestito @rtorrero This is ready to be reviewed too!

arbulu89:
@fabriziosestito @rtorrero @stefanotorresi 
Pinging again! Needing a review here

README.md
```diff
@@ -182,32 +182,41 @@ In order to start them, some packages must be installed and started. Here a quic
...
+```
+2021-09-02 07:13:48,715 ERROR django.security.DisallowedHost: Invalid HTTP_HOST header: '10.74.1.5:8000'. You may need to add '10.74.1.5' to ALLOWED_HOSTS.
+2021-09-02 07:13:48,732 WARNING django.request: Bad Request: /api/
+```
+
+To fix it run:
+
+```
+export ARA_ALLOWED_HOSTS="['10.74.1.5']"
+```
```
fabriziosestito, line 203: this would block localhost connection to the ara server tho.
Maybe we could just be lazy with `ARA_ALLOWED_HOSTS=['*']` ?

fabriziosestito (approved):

README.md
arbulu89: Ok, I will add a new line with this option for lazy admins XD

PR #193, opened 2021-08-09
Add grouped cib resources

### LINKED ISSUE #153: Virtual IP of an host doesn't show on Azure
Original Issue: #141 

In the current implementation the Virtual IP is not showing on Azure.

See demo env screenshot:
image

The virtual ip resource seems to be grouped on Azure, but the CIB exporter is not  parsing groups, so we currently don't have this information available.
https://github.com/ClusterLabs/ha_cluster_exporter/blob/2b282af270f3b6bbc7880a1d9c8644bd570b0a26/collector/pacemaker/cib/data.go#L18

[cib.xml.txt](https://github.com/trento-project/trento/files/6841747/cib.xml.txt) (In failover status)

On libvirt installation, it works as expected since the resource is not grouped.

fabriziosestito:
This PR incorporates the CIB struct and parsing logic from https://github.com/ClusterLabs/ha_cluster_exporter/
and adds grouped primitives parsing which is needed to  retrieve the virtual ip address on azure.

Resolves #153


arbulu89 (commented):
@fabriziosestito 
Once started, wouldn't it make sense to bring the `crmmon` parser as well?, otherwise it looks strange still have the `ha_cluster_exporter` import over there.
I would vote to bring both. If we just bring it partially, it would even make sense to just patch the error in the `ha_cluster_exporter`. (In fact, we should even fix in the `ha_cluster_exporter` project).

fabriziosestito:
> @fabriziosestito
> Once started, wouldn't it make sense to bring the `crmmon` parser as well?, otherwise it looks strange still have the `ha_cluster_exporter` import over there.
> I would vote to bring both. If we just bring it partially, it would even make sense to just patch the error in the `ha_cluster_exporter`. (In fact, we should even fix in the `ha_cluster_exporter` project).

@arbulu89 Yes, it makes a lot of sense, should I add it to this PR?

arbulu89:
> > @fabriziosestito
> > Once started, wouldn't it make sense to bring the `crmmon` parser as well?, otherwise it looks strange still have the `ha_cluster_exporter` import over there.
> > I would vote to bring both. If we just bring it partially, it would even make sense to just patch the error in the `ha_cluster_exporter`. (In fact, we should even fix in the `ha_cluster_exporter` project).
> 
> @arbulu89 Yes, it makes a lot of sense, should I add it to this PR?

Yes, I would add it here as it is the same migration after all

internal/cluster/cluster.go
```diff
@@ -6,9 +6,10 @@ import (
 	"strings"
 
 	// Reusing the Prometheus Ha Exporter cibadmin xml parser here
```
arbulu89, line 8: This comment looks a leftover. If we want to have a referent I would point it out with the url of the project just in top of the 2 imports

arbulu89 (approved):
Nice!!!
I left a little comment about one of the comments that stays in the code, if you want to have a look

arbulu89 (approved):

PR #195, opened 2021-08-12
Add sap systems, clusters and hosts tags

fabriziosestito:
This PR enables the user to tag sapsystems, clusters and hosts.

- Select2 plugin is used for the frontend, tags are stored in the kv store outside the discovery tree.
- Filtering by tag was added in the list pages.
- Added API for creating and deleting tags via ajax, 
- Added `gin-swagger` and the  generated documentation.
- The `swag` command line tool is added to the tracked tools as well.

- [BUGFIX]  the data coming from the `Load` functions of the kv store models is now sorted with a sort function, this is needed because we were relying on the map ordering which is not guaranteed in golang. This was causing several quirks, to mention a major one the pagination was non-deterministic

tags


internal/hosts/hosts.go
```diff
@@ -141,20 +143,42 @@ func CreateFilterMetaQuery(query map[string][]string) string {
 	return strings.Join(filters, " and ")
 }
 
-func Load(client consul.Client, query_filter string, health_filter []string) (HostList, error) {
+func Load(client consul.Client, queryFilter string, healthFilter []string, tagsFilter []string) (HostList, error) {
 	var hosts = HostList{}
 
-	query := &consulApi.QueryOptions{Filter: query_filter}
-	consul_nodes, _, err := client.Catalog().Nodes(query)
+	query := &consulApi.QueryOptions{Filter: queryFilter}
```
arbulu89, line 149: It's a pity that we cannot remove this filtering, but I guess that it is used to get the host list in the cluster and sap system pages

internal/tags/tags.go
```diff
@@ -0,0 +1,64 @@
...
+	var tags []string
+	for tag := range tagsMap {
+		tags = append(tags, tag)
+	}
+
+	return tags, nil
+}
+
+func (t *Tags) Create(tag string) error {
+	path := t.getKvTagsPath() + tag + "/"
```
arbulu89, line 47: I guess the `fmt.Sprintf` option would be cleaner

web/clusters.go
```diff
@@ -46,7 +48,7 @@ func detectClusterType(c *cluster.Cluster) string {
 
 	for _, c := range c.Crmmon.Clones {
 		for _, r := range c.Resources {
-			switch r.Agent {
+			switch r.Agent { /*  */
```
arbulu89, line 51: What is this comment for?

web/templates/blocks/clusters_table.html.tmpl
```diff
@@ -4,26 +4,36 @@
...
-                        {{- if .HasDuplicatedName }}<i class="eos-icons eos-18 text-warning" data-toggle="tooltip" data-original-title="This cluster has a duplicated name">warning</i>{{- end }}{{ .Name }}
+                        {{- if .HasDuplicatedName }}<i class="eos-icons eos-18 text-warning" data-toggle="tooltip" data-original-title="This cluster has a duplicated name">warning</i>{{- end }}<a href="/clusters/{{ .Id }}">{{ .Name }}</a>
                     </td>
-                    <td>{{ .Id }}</td>
+                    <td><a href="/clusters/{{ .Id }}">{{ .Id }}</a></td>
                     <td>{{ .Type }}</td>
+                    <td>{{- range $i, $v := .SIDs }}{{- if $i }},{{- end }} {{ . }}{{- end }}</td>
                     <td>{{ .HostsNumber }}</td>
                     <td>{{ .ResourcesNumber }}</td>
+                    <td>
```
arbulu89, line 28: This looks like something that could go in other template, passing the maybe data-url as param.
I guess you already checked and has some drawback hehe

internal/tags/tags.go
```diff
@@ -0,0 +1,64 @@
+package tags
+
+import (
+	"fmt"
+
+	"github.com/pkg/errors"
+	"github.com/trento-project/trento/internal/consul"
+)
+
+const kvTagsPath string = "trento/v0/tags/%s/%s/"
```
arbulu89, line 10: We have the rest of the KV paths stored in the `consul/kv.go`. I don't know if we could move it there, for consistency, as the package is already imported

internal/tags/tags.go
```diff
@@ -0,0 +1,64 @@
...
+func NewTags(client consul.Client, resource string, id string) *Tags {
+	return &Tags{
+		client:   client,
+		resource: resource,
+		id:       id,
+	}
+}
+
+func (t *Tags) getKvTagsPath() string {
+	return fmt.Sprintf(kvTagsPath, t.resource, t.id)
```
arbulu89, line 27: I think it would be better to use `path.Join` here

internal/tags/tags.go
```diff
@@ -0,0 +1,64 @@
...
+		tags = append(tags, tag)
+	}
+
+	return tags, nil
+}
+
+func (t *Tags) Create(tag string) error {
+	path := t.getKvTagsPath() + tag + "/"
+
+	if err := t.client.KV().PutMap(path, nil); err != nil {
```
arbulu89, line 49: I don't really get it here. You are posting a map without values, and the real value comes in the path itself. Shouldn't we have better a key/value format and put the tag as value?

internal/tags/tags.go
```diff
@@ -0,0 +1,64 @@
...
+		return errors.Wrap(err, "Error storing a host tags")
```
arbulu89, line 50: This error message looks wrong, we should have the `host` word, right?

web/clusters.go
```diff
@@ -333,34 +342,60 @@ func NewClustersTable(clusters map[string]*cluster.Cluster, hostList hosts.HostL
 		}
 	}
 
-	return clusterTable
+	sort.Slice(clusterTable, func(i, j int) bool {
```
arbulu89, line 345: Is this piece of code fixing the mentioned bug?

web/clusters.go
```diff
@@ -369,24 +404,44 @@ func (t ClustersTable) GetAllSIDs() []string {
...
+				set[sid] = struct{}{}
+				sids = append(sids, sid)
 			}
 		}
 	}
 
 	return sids
 }
 
+func (t ClustersTable) GetAllTags() []string {
```
arbulu89, line 423: It's a pity that we need that much code to simply remove duplicates...

web/hosts.go
```diff
@@ -52,9 +67,11 @@ func NewHostListHandler(client consul.Client) gin.HandlerFunc {
 		c.HTML(http.StatusOK, "hosts.html.tmpl", gin.H{
 			"Hosts":           hostList[firstElem:lastElem],
 			"SIDs":            getAllSIDs(hostList),
+			"Tags":            getAllTags(hostsTags),
```
arbulu89, line 70: What's the difference between `Tags` and `HostsTags` here?

arbulu89 (changes requested):
Hi @fabriziosestito and @rtorrero ,
I tried my best on reviewing the PR, but it is pretty big and I might overlook something.
Anyway, I have written down some comments that are worth to have a look.
Let me know what do you think

web/clusters.go
fabriziosestito: I blame copilot here 🤖

web/clusters.go
fabriziosestito: Yes, it sorts the slice so the results are deterministic

web/hosts.go
fabriziosestito: Tags is the set of all the de-duplicated tags of all the hosts, and it is used to fill the options of the filter

fabriziosestito: `HostsTags` is a map with the hostname as key and the tags of that host as value

web/templates/blocks/clusters_table.html.tmpl
fabriziosestito: yes i tried but since there is no way that i know to pass variables directly to a nested template, i decided to repeat it, since i did not want to pass them from the backend.

arbulu89 (approved):
Thank you for addressing my comments!
I did my best reviewing, but it is difficult in a such a long PR.

Great job @rtorrero and @fabriziosestito

PR #201, opened 2021-08-23
Fix: Remediation text for the MCAST/UCAST check 1.1.6

aleksei-burlakov:
It improves the remediation text for the MCAST/UCAST check 1.1.6.

### DIFF
--- a/examples/generic-azure.yaml
+++ b/examples/generic-azure.yaml
@@ -164,15 +164,42 @@ groups:
             - flag: "udpu"
         remediation: |
           ## Remediation
-          It will be difficult to change the corosync MCAST transport to UCAST
-          on an existing cluster. Besides adding the 'transport: udpu' parameter one
-          would have to configure each node separatelly. The easiest way to
-          make the cluster UCAST it to bootstrap it. When creating the cluster
-          you would have to add the key -u, for example
-          $ crm cluster init -u ...
+          To change the corosync MCAST transport to UCAST edit the /etc/corosync/corosync.conf
+          as in the example
+          ```
+              max_messages: 20
+              interface {
+                  ringnumber: 0
+          -       bindnetaddr: 10.162.32.167
+          -       mcastaddr: 239.11.100.41
+                  mcastport: 5405
+                  ttl: 1
+              }
+          +   transport: udpu
+          ...
+          +nodelist {
+          +       node {
+          +               ring0_addr: 10.162.32.167
+          +               nodeid: 1
+          +       }
+          +
+          +       node {
+          +               ring0_addr: 10.162.32.89
+          +               nodeid: 2
+          +       }
+          +
+          +}
+          ```
+          1) stop the already running cluster by using **systemctl stop pacemaker**
+          2) In the totem section, in the interface subsection remove the
+          keys-value pairs **bindnetaddr** and **mcastaddr**.
+          3) In the totem section add key-value pair **transport: udpu**
+          4) Add section nodelist and subsections node for each nodes of the
+          cluster, where the **ring0_addr** is the IP address of the node.
 
           ## References
           - https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/high-availability-guide-suse-pacemaker
+          - section 9.1.3 in https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-sr-guide-PerfOpt-15/
         scored: true
       - id: 1.1.6.runtime
         description: "Corosync transport parameter is set to udpu (runtime)"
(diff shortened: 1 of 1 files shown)


aleksei-burlakov:
@lee-martin could you please have a look?

examples/generic-azure.yaml
```diff
@@ -164,12 +164,37 @@ groups:
             - flag: "udpu"
         remediation: |
           ## Remediation
-          It will be difficult to change the corosync MCAST transport to UCAST
-          on an existing cluster. Besides adding the 'transport: udpu' parameter one
-          would have to configure each node separatelly. The easiest way to
-          make the cluster UCAST it to bootstrap it. When creating the cluster
-          you would have to add the key -u, for example
-          $ crm cluster init -u ...
+          To change the corosync MCAST transport to UCAST edit the /etc/corosync/corosync.conf
```
lee-martin, line 167: So test id 1.1.6 checks if "Corosync transport parameter is set to udpu" , so how does that relate to documenting how to fix the MCAST/UCAST setting? It seems there are actually two separate tests at hand here:
1) Is udpu set correctly, and if not, how to remediate
2) Is UCAST set, and if not how to remediate.   The MCAST/UCAST text is improved, but that remediation belongs in the UCAST check (which I can't find in the xaml file, maybe needs creating?

examples/generic-azure.yaml
```diff
@@ -164,12 +164,37 @@ groups:
...
+          +       }
+          +
+          +       node {
+          +               ring0_addr: 10.162.32.89
+          +               nodeid: 2
+          +       }
+          +
+          +}
+          ```
+          1) In the totem section, in the interface subsection remove the
```
lee-martin, line 193: Reading the HANA HA setup guide, it seems the first step is  "stop the already running cluster by using systemctl stop pacemaker" ? See section 9.1.3 of https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-sr-guide-PerfOpt-15/

examples/generic-azure.yaml
```diff
@@ -164,12 +164,37 @@ groups:
...
+          keys-value pairs **bindnetaddr** and **mcastaddr**.
+          2) In the totem section add key-value pair **transport: udpu**
+          3) Add section nodelist and subsections node for each nodes of the
```
lee-martin, line 196: Does the nodelist section apply to udpu, or only UCAST remediation?
As mentioned above, it seems two separate checks (udpu and UCAST)  got merged here.

examples/generic-azure.yaml
```diff
@@ -164,12 +164,37 @@ groups:
...
+          cluster, where the **ring0_addr** is the IP address of the node.
 
           ## References
           - https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/high-availability-guide-suse-pacemaker
```
lee-martin, line 200: Please always also reference SUSE guides where they exist.
In this case section 9.1.3 of https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-sr-guide-PerfOpt-15/

Moreover I do not even find the UCAST recommendation at: https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/high-availability-guide-suse-pacemaker
Actually that MSFT guide does not even seem to mention udpu explicitly either. Can you have a second look to see if my eyes are missing these details, or if you looks a a different MSFT link? We really should avoid making incorrect claims about what sources recommend, unless I missed something or it was changed of course.

lee-martin:
> @lee-martin could you please have a look?

@aleksei-burlakov  Sure, please see my comments above. Please comment me directly when you are done since I don't yet have a mail filter which sees when I'm being requested to review. This is still new to me :)

lee-martin:
@aleksei-burlakov  thanks, this looks pretty good now. The only missing thing I can see is that the reference is " section 9.1.3 " specifically and not the general " section 9.1 ", but other than that, I'm good.

@arbulu89 & @stefanotorresi  following our conversation, it seems all we are missing here is the minor correction in the reference mentioned above, once we have that we are good to merge from my side

lee-martin (approved):
looks fine from my side now. thanks.

aleksei-burlakov:
It fails, I think because the benchcommon checkers were removed.

arbulu89:
> It fails, I think because the benchcommon checkers were removed.

`benchcommon` has only be removed in the `runner` branch. This `main` branch should unaffected

arbulu89:
@lee-martin changes done for the ansible implementation too: https://github.com/trento-project/trento/pull/225

arbulu89:
@aleksei-burlakov @lee-martin 
Closing this, as the text has been already ported to the ansible optoin

PR #202, opened 2021-08-23
Add 2.1.* - 2.2.* tests

brett060102:
Add the 2.1.* and 2.2.* tests.

Also added support to run all tests.


runner/ansible/roles/2.1.1/tasks/main.yml~
```diff
@@ -0,0 +1,30 @@
+---
```
arbulu89, line 1: This file looks some kind of old backup. We should remove it, right?

runner/ansible/roles/2.1.3/tasks/main.yml
```diff
@@ -0,0 +1,39 @@
...
+    test_id: '2.1.3'
+
+- name: "{{ test_id }}.check"
+  shell: |
+    # SAP HANA Resource Agent is not configured, then pass
+    cibadmin -Q --xpath "//primitive[@type='SAPHana']/@type" || exit 0
+    # if configured and not msl, then fail
+    cibadmin -Q --xpath "//master/primitive[@type='SAPHana']/@type" || exit 1
+    # check the following values, if any do not match expected fail.
+    [[ $(cibadmin -Q --xpath "//primitive[@type='SAPHana']/operations/op [@name='monitor'] [@role='Master']" | grep -oP 'interval="\K[^"]+') != "{{ expected[test_id + '.master_interval'] }}" ]] && exit 1
```
arbulu89, line 14: Could we put the `//primitive[@type='SAPHana']/operations/op` string in a variable? I see it repeated multiple times

runner/ansible/roles/2.1.3/tasks/main.yml
```diff
@@ -0,0 +1,39 @@
...
+    [[ $(cibadmin -Q --xpath "//primitive[@type='SAPHana']/operations/op [@name='start']" | grep -oP 'timeout="\K[^"]+') != "{{ expected[test_id + '.start_timeout'] }}" ]] && exit 1
+    [[ $(cibadmin -Q --xpath "//primitive[@type='SAPHana']/operations/op [@name='stop']" | grep -oP 'interval="\K[^"]+') != "{{ expected[test_id + '.stop_interval'] }}" ]] && exit 1
+    [[ $(cibadmin -Q --xpath "//primitive[@type='SAPHana']/operations/op [@name='stop']" | grep -oP 'timeout="\K[^"]+') != "{{ expected[test_id + '.stop_timeout'] }}" ]] && exit 1
+    [[ $(cibadmin -Q --xpath "//primitive[@type='SAPHana']/operations/op [@name='promote']" | grep -oP 'interval="\K[^"]+') != "{{ expected[test_id + '.promote_interval'] }}" ]] && exit 1
+    [[ $(cibadmin -Q --xpath "//primitive[@type='SAPHana']/operations/op [@name='promote']" | grep -oP 'timeout="\K[^"]+') != "{{ expected[test_id + '.promote_timeout'] }}" ]] && exit 1
+    [[ $(cibadmin -Q --xpath "//primitive[@type='SAPHana']/instance_attributes/nvpair [@name='PREFER_SITE_TAKEOVER']" | grep -oP 'value="\K[^"]+' | tr '[:upper:]' '[:lower:]') != "{{ expected[test_id + '.nvpair_PREFER_SITE_TAKEOVER'] }}" ]] && exit 1
```
arbulu89, line 24: The same for `//primitive[@type='SAPHana']/instance_attributes/nvpair `, it is repeated many times

runner/ansible/roles/2.1.2/tasks/main.yml
```diff
@@ -0,0 +1,32 @@
...
+    test_id: '2.1.2'
+
+- name: "{{ test_id }}.check"
+  shell: |
+    # SAPHanaTopology is not configured, then pass
+    cibadmin -Q --xpath "//primitive[@type='SAPHanaTopology']/@type" || exit 0
+    # if configured and not clone, then fail
+    cibadmin -Q --xpath "//clone/primitive[@type='SAPHanaTopology']/@type" || exit 1
+    # check the following values, if any do not match expected fail.
+    [[ $(cibadmin -Q --xpath "//primitive[@type='SAPHanaTopology']/operations/op [@name='monitor']" | grep -oP 'interval="\K[^"]+') != "{{ expected[test_id + '.monitor_interval'] }}" ]] && exit 1
```
arbulu89, line 14: `//primitive[@type='SAPHanaTopology']/operations/op` is repeated many times. We could put it in a variable to reduce the amount of text

runner/ansible/roles/2.1.4/tasks/main.yml
```diff
@@ -0,0 +1,28 @@
...
+- name: Set Test ID
+  set_fact:
+    test_id: '2.1.4'
+
+- name: "{{ test_id }}.check"
+  shell: |
+    # if virtIP not configured, exit with fail
+    cibadmin -Q --xpath "//primitive[@type='IPaddr2']/@type" || exit 1
+    # if not grouped, exit with fail
+    cibadmin -Q --xpath "//group/primitive[@type='IPaddr2']/@type" || exit 1
```
arbulu89, line 12: This grouped check might be risky, as it only belongs to azure. Could we check here if the cloud provider is azure to run this check (only the `//group/primitive[@type='IPaddr2']/@type` one)

runner/ansible/roles/2.1.5/tasks/main.yml
```diff
@@ -0,0 +1,27 @@
+---
+
+- name: Set Test ID
```
arbulu89, line 3: This is an only azure test. We could put an additional `when` condition to run in the imported roles list

runner/ansible/roles/2.1.6/tasks/main.yml
```diff
@@ -0,0 +1,35 @@
+---
+
+- name: Set Test ID
+  set_fact:
+    test_id: '2.1.6'
+
+- name: "{{ test_id }}.check"
+  shell: |
+    #   if rsc_colocation constraint is not configured, then fail
+    cibadmin -Q --xpath "//constraints/rsc_colocation" || exit 1
```
arbulu89, line 10: Multiple times repeated strings here too `//constraints/rsc_colocation` and `//constraints/rsc_order`

runner/ansible/roles/2.1.7/tasks/main.yml
```diff
@@ -0,0 +1,36 @@
...
+    # get ha_dr_provider_SAPHanaSR section from config file
+    ha_dr_provider_SAPHanaSR=$(sed -n -e '/ha_dr_provider_SAPHanaSR\]/,/\[/ p; /trace\]/,/\[/ p' /usr/sap/${sid}/SYS/global/hdb/custom/config/global.ini)
+    # if following not found, fail
+    [[ ${ha_dr_provider_SAPHanaSR} != "{{ expected[test_id + '.provider'] }}" ]] && exit 1
+    [[ ${ha_dr_provider_SAPHanaSR} != "{{ expected[test_id + '.path'] }}" ]] && exit 1
+    [[ ${ha_dr_provider_SAPHanaSR} != "{{ expected[test_id + '.execution_order'] }}" ]] && exit 1
+    [[ ${ha_dr_provider_SAPHanaSR} != "{{ expected[test_id + '.ha_dr_saphanasr'] }}" ]] && exit 1
+    # if hook not found, exit with fail
+    share_path=$(cat /hana/shared/PRD/global/hdb/custom/config/global.ini | grep "path =" | sed "s/.*= //")
+    [[ -s "$path/SAPHanaSR.py" ]] || exit 1
```
arbulu89, line 23: Where is this `$path` coming from? Maybe `share_path`

runner/ansible/roles/2.1.7/tasks/main.yml
```diff
@@ -0,0 +1,36 @@
...
+  shell: |
+    #  if no sid, then fail
+    sid=$(crm configure show | grep -m1 SID= | sed -e "s/.*SID=\(...\).*/\1/")
+    [[ $sid == "" ]] && exit 1
+    # if config file does not exist, exit fail
+    [[ -e /usr/sap/${sid}/SYS/global/hdb/custom/config/global.ini ]] || exit 1
+    # get ha_dr_provider_SAPHanaSR section from config file
+    ha_dr_provider_SAPHanaSR=$(sed -n -e '/ha_dr_provider_SAPHanaSR\]/,/\[/ p; /trace\]/,/\[/ p' /usr/sap/${sid}/SYS/global/hdb/custom/config/global.ini)
+    # if following not found, fail
+    [[ ${ha_dr_provider_SAPHanaSR} != "{{ expected[test_id + '.provider'] }}" ]] && exit 1
```
arbulu89, line 17: All these 4 expected values are constants in fact. I think we should remove them from the variables files. The user shouldn't never touch them. As far as I know.
But well, in a second thought, let's keep in them where they are, maybe it is easier to find them

runner/ansible/roles/2.1.8/tasks/main.yml
```diff
@@ -0,0 +1,28 @@
+---
+
+- name: Set Test ID
+  set_fact:
+    test_id: '2.1.8'
+
+- name: "{{ test_id }}.check"
+  shell: |
+    # make sure sudoers is set up
+    grep -E "[a-z][a-z0-9]{2}adm +ALL=\(ALL\) +NOPASSWD: +/usr/sbin/crm_attribute +-n +hana_[a-z][a-z0-9]{2}_site_srHook_\*" /etc/sudoers && exit 0
```
arbulu89, line 10: A lot of duplicated texts. Let's try to move them to variables

runner/ansible/vars/azure_env.yml
```diff
@@ -24,3 +24,50 @@ expected:
...
+  "2.1.5.nvpair_resource_stickiness": "0"
+  "2.1.6.rsc_colocation_score": "4000"
+  "2.1.6.rsc_colocation_rsc": "g_ip"
+  "2.1.6.rsc_colocation_rsc_role": "Started"
+  "2.1.6.rsc_colocation_with_rsc": "msl_SAPHana"
+  "2.1.6.rsc_colocation_with_rsc_role": "Master"
+  "2.1.6.rsc_order_kind": "Optional"
+  "2.1.6.rsc_order_first": "cln_SAPHanaTopology"
+  "2.1.6.rsc_order_then": "msl_SAPHana"
+  "2.1.7.provider": "provider = SAPHanaSR"
```
arbulu89, line 59: Can we change the next for lines to only contain the value, and not the key?
It requires to change the test of course

runner/ansible/roles/2.2.2/tasks/main.yml
```diff
@@ -0,0 +1,24 @@
+---
+
+- name: Set Test ID
```
arbulu89, line 3: I don't know if we should have this check, that only passes if the SLE version is 15.
@lee-martin should we keep it?
Maybe we could check if the version is greater than 15 at most

runner/ansible/roles/2.2.3.exclude/tasks/main.yml
```diff
@@ -0,0 +1,24 @@
+---
+
+- name: Set Test ID
```
arbulu89, line 3: Can we run this checks with ansible native methods? https://docs.ansible.com/ansible/latest/collections/ansible/builtin/package_module.html

runner/ansible/roles/2.2.8/tasks/main.yml
```diff
@@ -0,0 +1,31 @@
+---
+
+- name: Set Test ID
+  set_fact:
+    test_id: '2.2.8'
+
+- name: "{{ test_id }}.check"
+  shell: |
+    #  HANA and SPS versions are compatible
+    sid=$(sudo crm configure show | grep -m1 SID= | sed -e "s/.*SID=\(...\).*/\1/" | tr '[:upper:]' '[:lower:]')
```
arbulu89, line 10: Careful with this `sudo`, we shouldn't have it here if possible

runner/ansible/roles/2.2.8/tasks/main.yml
```diff
@@ -0,0 +1,31 @@
...
+    full_version=$(sudo -i -u ${sid}adm HDB version | grep "version:" | sed -e "s/^.*:[\ ]*//")
+    hana_version=$(echo $full_version | cut -d. -f1)
+    sps_version=$(echo $full_version | cut -d. -f3)
+    sps_revision=$(echo $full_version | cut -d. -f4)
+    [[ "$hana_version" = "{{ expected[test_id] }}" ]] && exit 0
+    [[ "$hana_version" != "1" ]] && exit 1
```
arbulu89, line 16: This `"$hana_version" != "1"` looks wrong. Actually, almost everyone will have hana version 2.xx.
This if condition is used to determine the next 2 if conditions, which I guess should have the values (`110` and `03` in the vars file)

arbulu89 (changes requested):
Hi @brett060102 
Comments added

examples/hana-scale-up-perf-optimized-azure.yaml
```diff
@@ -296,7 +300,7 @@ groups:
             - flag: "true5"
         remediation: |
           ## Remediation
-            Set up the sudoers entry to allow <sidadm> in /etc/sudoers to use the srHook.
+            Set up the sudo ers entry to allow <sidadm> in /etc/sudoers to use the srHook.
```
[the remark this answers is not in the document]
brett060102: that change should not be there. I did not mean to make any changes outside of ansible space.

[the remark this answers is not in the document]
brett060102: All changed backed out

runner/ansible/check.yml
```diff
@@ -18,137 +18,278 @@
     - name: include 1.1.1
       import_role:
         name: '1.1.1'
-      when: "'1.1.1' in cluster_selected_checks"
+      when: ('1.1.1' in cluster_selected_checks) or
+            (cluster_selected_checks | lower == "all")
```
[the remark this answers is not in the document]
brett060102: I am OK with that being removed.

runner/ansible/roles/2.1.1/tasks/main.yml~
brett060102: yep

runner/ansible/roles/2.1.2/tasks/main.yml
brett060102: Agreed, but I can't make that change from this system. I played with that when doing these and it is a bit tricky. And I can't test it until I get back home. I am cool with your making this change though.

runner/ansible/roles/2.1.3/tasks/main.yml
brett060102: I think so, but I can't do that from where I am. I am fine you giving this a shot.

runner/ansible/roles/2.1.3/tasks/main.yml
brett060102: Same reply as other cases.

runner/ansible/roles/2.1.4/tasks/main.yml
brett060102: I am cool with making that change, but as with the others, I can't do it at this time.

runner/ansible/roles/2.1.5/tasks/main.yml
brett060102: Yes, we can do that.

runner/ansible/roles/2.1.6/tasks/main.yml
brett060102: Agreed, could be set as shell var, but I couldn't test the result.

runner/ansible/roles/2.1.7/tasks/main.yml
brett060102: Let's talk about this one. Not quite sure what you mean.

runner/ansible/roles/2.1.7/tasks/main.yml
brett060102: yes, I should have updated that when I copied the code from example.

runner/ansible/roles/2.1.8/tasks/main.yml
brett060102: Give it a shot. I would need to be on my home machine to do that.

brett060102 (commented):
I have changed what I could.

runner/ansible/check.yml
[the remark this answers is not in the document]
brett060102: removed

runner/ansible/roles/2.1.1/tasks/main.yml~
brett060102: removed

runner/ansible/roles/2.2.8/tasks/main.yml
brett060102: The section of code from the example this replicates is:
          # SAP HANA 2.0 is supported with all known SPS versions.
          if [[ "$hana_version" = "2" ]]; then echo passed; fi
          # For SAP HANA 1.0 you need version SPS10 rev3, SPS11 or newer.
          if [[ "$hana_version" = "1" ]]; then
            if [[ "$sps_version" -ge "110" ]]; then echo passed; fi
            if [[ "$sps_version" = "100" && "$sps_revision" = "03" ]]; then echo passed; fi
          fi
What should be happening is that if the hanna_version matches expected, then we exit with a pass. So, the test != 1 test only happens if the version is not 2. We could put 110 and 3 into the vars file, but if that won't ever change I am not sure we need to put it in the vars file.

runner/ansible/roles/2.2.8/tasks/main.yml
brett060102: Agreed, but I think that command does need root privs. We can look at https://docs.ansible.com/ansible/2.4/become.html for this kind of thing, but I would need to be on my primary system to look into to that.

runner/ansible/roles/2.2.3.exclude/tasks/main.yml
brett060102: yes, we should be able to do that/

runner/ansible/roles/2.1.6/tasks/main.yml
arbulu89: We could have a couple of variables for these 2 xpath items.
But well, not a bid deal

runner/ansible/roles/2.1.7/tasks/main.yml
arbulu89: I mean that the values on the expected dictionary `.provider`, `.path`, `.execution_order`, etc, are constants, so they never will change, I would vote to put them directly here rather than in the variables file

runner/ansible/roles/2.2.4/tasks/main.yml
```diff
@@ -0,0 +1,24 @@
+---
+
+- name: Set Test ID
+  set_fact:
+    test_id: '2.2.4'
+
+- name: "{{ test_id }}.check"
+  shell: |
+    # Check the pacemaker version IS
+    PACKAGE=$(rpm -qv corosync)
```
arbulu89, line 10: I think we should use `sed` as in the pacemaker example here, to not have the whole `corosync` string.
And even use the same nomenclature like `VERSION_NUMBER` etc

arbulu89 (changes requested):
Hey @brett060102 ,
Some new comments added.
I'm not still sure about the 2.2 chapter.
@diegoakechi @lee-martin @stefanotorresi Are we sure we want to include these checks now and here? Keep in mind that we cannot tune the expected values, so "constant" package and os versions will tend to fail in many cases.
Opinions?

runner/ansible/roles/2.1.3/tasks/main.yml
brett060102: Done

runner/ansible/roles/2.1.3/tasks/main.yml
brett060102: Done

runner/ansible/roles/2.2.2/tasks/main.yml
lee-martin: @arbulu89 This looks a little strange, but this is the first time I look at our Ansible checks, so I might be missing something.
So the original check ID 2.2.2 is in https://github.com/trento-project/trento/blob/main/examples/hana-scale-up-perf-optimized-azure.yaml  and that checks if the OS is 15 SP1 or newer (using the value 151).  To me this reads like somehow the test ID get mixed up with the target value we are actually checking for.

I also miss the "remediation" section, which most importantly contains the reason why we are even performing the check. Where does one find the remediation text in the ansible runner?

lee-martin: > Hey @brett060102 ,
> Some new comments added.
> I'm not still sure about the 2.2 chapter.
> @diegoakechi @lee-martin @stefanotorresi Are we sure we want to include these checks now and here? Keep in mind that we cannot tune the expected values, so "constant" package and os versions will tend to fail in many cases.
> Opinions?

@arbulu89  I agree constant values are highly suspicious and should be reviewed. Also, all these checks seem to back the remediation text which references the reason why we are checking something. Where does one find the remediation text in the ansible runner? This would then justify why each check does what it does.
Looking at the where these checks were ported from ( https://github.com/trento-project/trento/blob/main/examples/hana-scale-up-perf-optimized-azure.yaml ) it seems many didn't have a reference and my guess is they originate from a spreadsheet we used early in Trento, but it did not have references.  We need to review all checks which do not have references anyway, and this process should then also uncover unrealistic checks, e.g. overly specific version checks of packages etc.  Can we open issues on a check by check basis to do this? I would not make this a stopper for implementing ainsible runner right now, we can fix these checks after the transition step by step.

arbulu89: The remediation texts come in an additional file. We will add them in other PR. 
We just want to have here the check implementation.
The remediation will be copy/paste from the current bench common checks anyway

arbulu89: My fault on this test. It actually checks if the version is **greater** than 151.
I guess we can move on with this by now

runner/ansible/roles/2.1.3/tasks/main.yml
```diff
@@ -0,0 +1,41 @@
+---
+
+- name: Set Test ID
+  set_fact:
+    test_id: '2.1.3'
+
+- name: "{{ test_id }}.check"
+  shell: |
+    # SAP HANA Resource Agent is not configured, then pass
+    cibadmin -Q --xpath "//primitive[@type='SAPHana']/@type" || exit 0
```
arbulu89, line 10: This should report a `exit 1`. If this primitive doesn't exist we have a missconfigured cluster.
(Update the comment above too)

runner/ansible/roles/2.1.3/tasks/main.yml
```diff
@@ -0,0 +1,41 @@
...
+    [[ $(cibadmin -Q --xpath "${XPATH} [@name='monitor'] [@role='Slave']" | grep -oP 'timeout="\K[^"]+') != "{{ expected[test_id + '.slave_timeout'] }}" ]] && exit 1
+    [[ $(cibadmin -Q --xpath "${XPATH} [@name='start']" | grep -oP 'interval="\K[^"]+') != "{{ expected[test_id + '.start_interval'] }}" ]] && exit 1
+    [[ $(cibadmin -Q --xpath "${XPATH} [@name='start']" | grep -oP 'timeout="\K[^"]+') != "{{ expected[test_id + '.start_timeout'] }}" ]] && exit 1
+    [[ $(cibadmin -Q --xpath "${XPATH} [@name='stop']" | grep -oP 'interval="\K[^"]+') != "{{ expected[test_id + '.stop_interval'] }}" ]] && exit 1
+    [[ $(cibadmin -Q --xpath "${XPATH} [@name='stop']" | grep -oP 'timeout="\K[^"]+') != "{{ expected[test_id + '.stop_timeout'] }}" ]] && exit 1
+    [[ $(cibadmin -Q --xpath "${XPATH} [@name='promote']" | grep -oP 'interval="\K[^"]+') != "{{ expected[test_id + '.promote_interval'] }}" ]] && exit 1
+    [[ $(cibadmin -Q --xpath "${XPATH} [@name='promote']" | grep -oP 'timeout="\K[^"]+') != "{{ expected[test_id + '.promote_timeout'] }}" ]] && exit 1
+    XPATH="//primitive[@type='SAPHana']/instance_attributes/nvpair"
```
arbulu89, line 25: I would add a new line here, to add visibility

runner/ansible/roles/2.1.5/tasks/main.yml
```diff
@@ -0,0 +1,27 @@
...
+  set_fact:
+    test_id: '2.1.5'
+
+- name: "{{ test_id }}.check"
+  shell: |
+    #  if azure-lb RA is not configured, then fail
+    cibadmin -Q --xpath "//primitive[@type='azure-lb']/@type" || exit 1
+    # if not grouped, exit with fail
+    [[ $(cibadmin -Q --xpath "//group/primitive[@type='azure-lb']/@type") != "0" ]] && exit 1
```
arbulu89, line 12: We need to change this call. I would do the same that you did in the 2.1.4 check.

```
cibadmin -Q --xpath "//group/primitive[@type='azure-lb']/@type" || exit 1
```

runner/ansible/roles/2.1.7/tasks/main.yml
```diff
@@ -0,0 +1,36 @@
+---
+
+- name: Set Test ID
+  set_fact:
+    test_id: '2.1.7'
+
+- name: "{{ test_id }}.check"
```
arbulu89, line 7: I find this test quite wrong.
Here what the check should do:
1. FInd the sid
2. With the sid we can get the /usr/sap/${sid}/SYS/global/hdb/custom/config/global.ini file
3. This file is a INI file format, which has a [ha_dr_provider_SAPHanaSR] header
4. Within this header, we need to check the content of provider, path and execution order

So, could we split the check in 2 tasks, the first to get the file path and register the value.
Once we have the value, we could use the ini files task type to check the content.
What do you think?

runner/ansible/roles/2.2.3/tasks/main.yml
```diff
@@ -0,0 +1,24 @@
...
+
+- name: Set Test ID
+  set_fact:
+    test_id: '2.2.3'
+
+- name: "{{ test_id }}.check"
+  shell: |
+    # Check the pacemaker version IS
+    VERSION_ID=$(rpm -qv pacemaker | sed -e "s/pacemaker-\(.*\)+.*/\1/")
+    [[ $VERSION_ID -ge "{{ expected[test_id] }}" ]] && exit 0
```
arbulu89, line 11: This comparison doesn't work

runner/ansible/roles/2.2.4/tasks/main.yml
```diff
@@ -0,0 +1,24 @@
+---
+
+- name: Set Test ID
+  set_fact:
+    test_id: '2.2.4'
+
+- name: "{{ test_id }}.check"
+  shell: |
+    # Check the pacemaker version IS
```
arbulu89, line 9: Typo in the comment `pacemaker -> corosync`

arbulu89 (changes requested):
Hey @brett060102 ,
Here many other comments.
As far as I have seen these are the checks that we need to fix:
2.1.2
2.1.3
2.1.5
2.1.7
2.2.3

runner/ansible/roles/2.1.2/tasks/main.yml
```diff
@@ -0,0 +1,33 @@
+---
+
+- name: Set Test ID
+  set_fact:
+    test_id: '2.1.2'
+
+- name: "{{ test_id }}.check"
+  shell: |
+    # SAPHanaTopology is not configured, then pass
+    cibadmin -Q --xpath "//primitive[@type='SAPHanaTopology']/@type" || exit 0
```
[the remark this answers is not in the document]
brett060102: fixed.

runner/ansible/roles/2.1.3/tasks/main.yml
brett060102: fixed

runner/ansible/roles/2.1.3/tasks/main.yml
brett060102: done

runner/ansible/roles/2.1.5/tasks/main.yml
brett060102: done

runner/ansible/roles/2.2.3/tasks/main.yml
brett060102: Fixed, changed to use:
changed_when: config_updated.stdout is version(expected[test_id], '<')

brett060102:
@arbulu89 fingers crossed, but I think this is good now. We might want to move to use the ansible version compare everywhere we do version checking, but I think we should hold on that for later.

runner/ansible/roles/2.2.4/tasks/main.yml
brett060102: Done

runner/ansible/roles/2.1.7/tasks/main.yml
brett060102: This should be fixed now. Was able to test with the sample file you sent me.
Still follows was was done in example code.

runner/ansible/roles/2.2.3.exclude/tasks/main.yml
```diff
@@ -0,0 +1,25 @@
...
+- name: "{{ test_id }}.check"
+  shell: |
+    # Check the pacemaker version IS NOT
+    # If not installed, exit with error
+    VERSION_ID=$(rpm -q --qf "%{VERSION}\n" pacemaker) || exit 2
+    [[ "$VERSION_ID" != "{{ expected[test_id] }}" ]] && exit 0
+    exit 1
+  check_mode: no
+  register: config_updated
+  changed_when: config_updated.rc != 0
```
arbulu89, line 16: We don't we use the `is version` logic here?
By the way, I don't get why you have started using the `exit 2` here.
`exit 1` doesn't do the job?

runner/ansible/roles/2.2.3/tasks/main.yml
```diff
@@ -0,0 +1,23 @@
...
+
+- name: Set Test ID
+  set_fact:
+    test_id: '2.2.3'
+
+- name: "{{ test_id }}.check"
+  shell: |
+    # Check the pacemaker version IS
+    # If not installed, exit with error
+    rpm -q --qf "%{VERSION}\n" pacemaker || exit 2
```
arbulu89, line 11: I don't think we need the `exit 2` here.
If the package doesn't exist, the `rpm` command will return a `1`, so I think it is neater to have the `failed_when: config_udpated.rc > 0`.

runner/ansible/roles/2.2.4/tasks/main.yml
```diff
@@ -0,0 +1,25 @@
...
+- name: Set Test ID
+  set_fact:
+    test_id: '2.2.4'
+
+- name: "{{ test_id }}.check"
+  shell: |
+    # Check the corosync version IS
+    # If not installed, exit with error
+    PACKAGE=$(rpm -qv corosync) || exit 2
+    [[ "$PACKAGE" == "{{ expected[test_id] }}" ]] && exit 0
```
arbulu89, line 12: The same here, why don't we use the `is version` logic?

arbulu89 (changes requested):
Hi @brett060102 ,
The changes look good in general.
As a last request, could we use the `is version` logic in all of the checks where we do version comparison. It looks the neater option, and having the same way in all the "same kind of check" items looks a better option

runner/ansible/roles/2.2.4/tasks/main.yml
```diff
@@ -8,12 +8,10 @@
...
     # Check the corosync version IS
     # If not installed, exit with error
-    PACKAGE=$(rpm -qv corosync) || exit 2
-    [[ "$PACKAGE" == "{{ expected[test_id] }}" ]] && exit 0
-    exit 1
+    rpm -q --qf "%{VERSION}\n" corosync || exit 2
   check_mode: no
   register: config_updated
-  changed_when: config_updated.rc != 0
+  changed_when: config_updated.stdout is version(expected[test_id], '<')
```
arbulu89, line 14: This should be `<=`. shouldn't it?
We want to have a `pass` when the version matches the expected value

runner/ansible/roles/2.2.5/tasks/main.yml
```diff
@@ -8,12 +8,10 @@
...
     # Check the sbd version IS
     # If not installed, exit with error
-    VERSION_ID=$(rpm -q --qf "%{VERSION}\n" sbd) || exit 2
-    [[ "$VERION_ID" == "{{ expected[test_id] }}" ]] && exit 0
-    exit 1
+    rpm -q --qf "%{VERSION}\n" sbd || exit 2
   check_mode: no
   register: config_updated
-  changed_when: config_updated.rc != 0
+  changed_when: config_updated.stdout is version(expected[test_id], '<')
```
arbulu89, line 14: This should be `<=` (expected value should be a pass)

runner/ansible/roles/2.2.7/tasks/main.yml
```diff
@@ -7,13 +7,13 @@
...
-    VERSION=$(python3 --version)
-    [[ "{{ expected[test_id] }}" =~ $VERSION ]] && exit 0
-    exit 1
+    python3 --version | cut -d" " -f2
   check_mode: no
   register: config_updated
-  changed_when: config_updated.rc != 0
-  failed_when: config_updated.rc > 1
+  changed_when: >
+    config_updated.stdout is version(expected[test_id + '.low'], '<') or
```
arbulu89, line 14: Let's test only the `.low` value here. I think it is good enough. The higher ending checking will create issues

runner/ansible/roles/2.2.8/tasks/main.yml
```diff
@@ -9,18 +9,15 @@
...
+    if [[ "$full_version" = "" ]]; then
+      echo "0"
+    else
+      echo "$full_version"
+    fi
   check_mode: no
   register: config_updated
-  changed_when: config_updated.rc != 0
-  failed_when: config_updated.rc > 1
+  changed_when: config_updated.stdout is version(expected[test_id], '<')
```
arbulu89, line 19: This should be `<=`

arbulu89 (changes requested):
Hi @brett060102 ,
We are almost! There, only some super minor changes required

runner/ansible/roles/2.2.4/tasks/main.yml
arbulu89: My bad! This works fine

runner/ansible/roles/2.2.5/tasks/main.yml
arbulu89: My bad! This works fine

runner/ansible/roles/2.2.8/tasks/main.yml
arbulu89: My bad! This works fine

runner/ansible/roles/2.1.6/tasks/main.yml
```diff
@@ -0,0 +1,35 @@
+---
```
arbulu89, line 1: Hey @brett060102 ,
I have tested this `2.1.6`, and I would say it is broken, as the queried lines return multi line outputs, so the comparisons always fail
Here my outputs using the command that we have right now:
```
vmhana01:/home/cloudadmin # cibadmin -Q --xpath "//constraints/rsc_colocation" | grep -oP 'score="\K[^"]+'
4000
-INFINITY
vmhana01:/home/cloudadmin # cibadmin -Q --xpath "//constraints/rsc_colocation" | grep -oP 'rsc="\K[^"]+' | grep -o g_ip
g_ip
vmhana01:/home/cloudadmin # cibadmin -Q --xpath "//constraints/rsc_colocation" | grep -oP 'rsc-role="\K[^"]+'
Started
Master
Started
Slave
vmhana01:/home/cloudadmin # cibadmin -Q --xpath "//constraints/rsc_colocation" | grep -oP 'with-rsc="\K[^"]+' | grep -o msl_SAPHana
msl_SAPHana
msl_SAPHana
vmhana01:/home/cloudadmin # cibadmin -Q --xpath "//constraints/rsc_colocation" | grep -oP 'with-rsc-role="\K[^"]+'
Master
Slave
vmhana01:/home/cloudadmin # cibadmin -Q --xpath "//constraints/rsc_order" | grep -oP 'kind="\K[^"]+'
Optional
vmhana01:/home/cloudadmin # cibadmin -Q --xpath "//constraints/rsc_order" | grep -oP 'first="\K[^"]+' | grep -o cln_SAPHanaTopology
cln_SAPHanaTopology
vmhana01:/home/cloudadmin # cibadmin -Q --xpath "//constraints/rsc_order" | grep -oP 'then="\K[^"]+' | grep -o msl_SAPHana
msl_SAPHana
vmhana01:/home/cloudadmin # 

```

arbulu89 (commented):
Other issue. The `2.1.6` check doesn't work properly

runner/ansible/roles/2.1.7/tasks/main.yml
```diff
@@ -0,0 +1,36 @@
...
+    [[ -e /usr/sap/${sid}/SYS/global/hdb/custom/config/global.ini ]] || exit 1
+    # get ha_dr_provider_SAPHanaSR section from config file
+    ha_dr_provider_SAPHanaSR=$(sed -n -e '/ha_dr_provider_SAPHanaSR\]/,/\[/ p; /trace\]/,/\[/ p' /usr/sap/${sid}/SYS/global/hdb/custom/config/global.ini)
+    # if following not found, fail
+    echo ${ha_dr_provider_SAPHanaSR} | grep "provider = {{ expected[test_id + '.provider'] }}" || exit 1
+    echo ${ha_dr_provider_SAPHanaSR} | grep "path = {{ expected[test_id + '.path'] }}" || exit 1
+    echo ${ha_dr_provider_SAPHanaSR} | grep "execution_order = {{ expected[test_id + '.execution_order'] }}" || exit 1
+    echo ${ha_dr_provider_SAPHanaSR} | grep "ha_dr_saphanasr = {{ expected[test_id + '.ha_dr_saphanasr'] }}" || exit 1
+    # if hook not found, exit with fail
+    share_path=$(cat /hana/shared/PRD/global/hdb/custom/config/global.ini | grep "path =" | sed "s/.*= //")
```
arbulu89, line 22: Other failure, this `PRD` here must be replaced by `${sid}`

arbulu89 (commented):
Error in test `2.1.7`

arbulu89:
Hey @brett060102,
I took the liberty to fix the mentioned issues. Find the commit here: https://github.com/trento-project/trento/pull/202/commits/4bd8b835e72cfd6f7be7672f8112a84a2a0d849a

runner/ansible/roles/2.1.7/tasks/main.yml
brett060102: yes and thank you.

runner/ansible/roles/2.2.7/tasks/main.yml
brett060102: Ok.

runner/ansible/roles/2.1.6/tasks/main.yml
brett060102: Thanks, I see that you added a space to the initial grep search string. Not sure how this worked in the example code though.

brett060102:
Look over you changes and all look fine to me. Glad this is in. Thank you.

PR #204, opened 2021-08-26
Add ansible_host connection option to ansibleHostsTemplate

fabriziosestito:
This PR adds the `ansible_host` connection option to the generated ansible inventory by using the node discovered IP address.
See: https://docs.ansible.com/ansible/latest/reference_appendices/special_variables.html#connection-variables

This is required in order to let ansible know the ip of the hosts without resolving them, so we don't ask the user to modify `/etc/hosts` or alter `resolv.conf/k8s core dns or pod dns policies.

See: https://github.com/trento-project/trento/pull/191#discussion_r691861673


arbulu89 (approved):
Nice! Thank you for checking out this

PR #206, opened 2021-08-27
Trento server helm chart

fabriziosestito:
This is an initial implementation of a Trento server helm chart, featuring the web  and the runner component.
It also contains an helm chart for the ARA dependency, which uses the official container image.
The official consul helm chart was added as dependency.
For the moment I'm hosting the images on my personal GitHub container registry, we could trigger the image building process from the CI once this has been tested.

The runner component documentation was followed from this PR: https://github.com/trento-project/trento/pull/191
A Docker image was built by using the branch from this PR: https://github.com/trento-project/trento/pull/189
with the modifications described in this PR: https://github.com/trento-project/trento/pull/204
The docker image used in the following example can be found here: https://github.com/users/fabriziosestito/packages/container/trento/6942033?tag=check5

What works:
- Agents outside kuberentes can connect to the consul server inside kubernetes, this is a scenario provided by the consul helm chart directly
- Runners and Web components seems to work correctly in a real SAP cluster

Needed improvements:
- We miss a readyness/live probe for the trento runner component
- ARA persistence layer as volume or db dependency
- Docker images built from are CI
- Curl/bash Installer script that follows the example below (this is WIP)

_NOTE:  Due to a known regression (https://github.com/k3s-io/k3s/issues/3704) this helm chart does not work with k3s `v1.21.3`, so make sure to have the latest version of k3s installed._

### Install or upgrade k3s
If installing as root:
`curl -sfL https://get.k3s.io | sh`

If installing as non root user:
`curl -sfL https://get.k3s.io | sh -s - --write-kubeconfig-mode 644`

### Export KUBECONFIG env variable
`export KUBECONFIG=/etc/rancher/k3s/k3s.yaml`

### Install or upgrade helm
`curl https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3 | bash`

### Add hashicorp helm repo
`helm repo add hashicorp https://helm.releases.hashicorp.com`

`helm repo update`

### Install trento server
`cd helm/trento-server/`

`helm dependency update`

`helm install trento . --set trento-web.tag="check5" --set trento-runner.tag="check5" --set-file trento-runner.privateKey=id_rsa_runner`


helm/trento-server/templates/NOTES.txt
```diff
@@ -0,0 +1,26 @@
...
+                 | .   |_   _   _   _   _   _| ,   |
+                 |    .| |_| |_| |_| |_| |_| |  .  |
+                 | ,   | .    .     .      . |    .|
+                 |   . |  .     . .   .  ,   |.    |
+     ___----_____| .   |.   ,  _______   .   |   , |---~_____
+_---~            |     |  .   /+++++++\    . | .   |         ~---_
+                 |.    | .    |+++++++| .    |   . |              ~-_
+              __ |   . |   ,  |+++++++|.  . _|__   |                 ~-_
+     ____--`~    '--~~__ .    |++++ __|----~    ~`---,              ___^~-__
+-~--~                   ~---__|,--~'                  ~~----_____-~'   `~----~
```
stefanotorresi, line 15: ahahah I love this!!!

stefanotorresi (changes requested):
In relation to https://github.com/trento-project/trento/pull/212#pullrequestreview-747830645 , we should probably move this stuff into `packaging/helm` or something like that.

Other than this, this looks very good!

Keep up the good work and, as mentioned previously, try to consolidate the installation script with the one related to the Agent component so that we provide complementary entry-points that are consistent with each other as far as the UX goes.

fabriziosestito:
> In relation to [#212 (review)](https://github.com/trento-project/trento/pull/212#pullrequestreview-747830645) , we should probably move this stuff into `packaging/helm` or something like that.
> 
> Other than this, this looks very good!
> 
> Keep up the good work and, as mentioned previously, try to consolidate the installation script with the one related to the Agent component so that we provide complementary entry-points that are consistent with each other as far as the UX goes.

@stefanotorresi thanks for the feedbacks, I will move forward in this direction and try to provide a script for the installation.

One thing I am still missing would be the containers registry. Should we use the trento organization github registry or should we move in the OBS direction? If we choose the latter, I would need some support  with obs.

stefanotorresi:
> Should we use the trento organization github registry [...]?
yes. :)

.gitignore
```diff
@@ -2,3 +2,4 @@
 /build
 /web/frontend/assets
 /web/frontend/node_modules
+/helm/trento-server/charts/*.tgz
```
stefanotorresi, line 5: you forgot to change the path here as well :)
```suggestion
/packaging/helm/trento-server/charts/*.tgz
```

stefanotorresi (changes requested):
it's `packaging`, not `packages` 😉

stefanotorresi (approved):

PR #207, opened 2021-08-30
Track obs files

aleksei-burlakov:
The files
-   obs/_service
-   obs/trento.changes
-   obs/trento.spec

will be fetched and extracted by obs_tar_scm. We do it, to make the version control in only one system, only in the github.


aleksei-burlakov:
It's too incomplete.

Issue #186, opened 2021-08-03
Cluster View: Present complete information on the Role Section for HANA


diegoakechi:
The cluster monitors HANA Application and collects the cluster attribute `role` as a delimited string containing different pieces of information about the HANA State, like if the node is primary or secondary, if the landscape is in a healthy state, among others. 

On the screen, this information should be visualized in an easier way by the user, where he immediately identifies the HANA status collected by the cluster. Currently, a partial view of the information is presented, what can be misleading, like on the screenshot below where there are 2 HANA masters, while the correct is possible that both nodes are running the HANA Nameserver process as master. 

![image](https://user-images.githubusercontent.com/8420718/128009007-9309089c-ba00-4f95-9658-2b58b3198537.png)

A full description of each field can be found on the SAPHanaSR main pages. I am pasting an extracted session here for reference:

```
roles - actual and configured roles for HANA on that node

The roles field in the Hosts section has six sub-fields for scale-up (A:B:C:D:E:F). It has four sub-fields for
scale-out (C:D:E:F). The sub-fields C:D:E:F are primarily useful for HANA scale-out. For scale-up that
four fields C:D:E:F should be master1:master:worker:master.

Field A: return code of landscapeHostConfiguration.py
Value: [ 4 | 3 | 2 | 1 | 0 ]
This field contains the return code of landscapHostConfiguration.py. The parameter does not tell you if the
secondary system is ready for a takeover. The meaning is different from common Linux return codes.
4 = OK - Everything looks perfect on the HANA primary.
3 = WARNING - A HANA Host Auto-Failover is taking place.
2 = INFO - The landscape is completely functional, but the actual role of the host differs from the config-
ured role.
1 = DOWN - There are not enough active hosts.
0 = FATAL - Internal script error, the state could not be determined.
See landscapeHostConfiguration.py.

Field B: HANA SR role
Value: [ "P"rimary | "S"econdary ]

Field C: NameServer Config Role
Value: [ master1 | master2 | master3 | worker | slave | standby | shtdown ]

Field D: NameServer Actual Role
Value: [ master | slave | standby | shtdown ]

Field E: IndexServer Config Role
Value: [ master1 | master2 | master3 | worker | slave | standby | shtdown ]

Field F: IndexServer Actual Role
Value: [ master | worker | slave | standby | shtdown ]
```

fabriziosestito:
@lee-martin @diegoakechi how do we want to show this information to the user?
[...]

PR #210, opened 2021-09-01
Remove HANA role badge

fabriziosestito:
This PR removes the HANA role badge from the site/node view of the cluster scale-up detail view, as mentioned in #186


arbulu89 (approved):
Nice!

PR #211, opened 2021-09-02
Show the cluster health based on runner checks

arbulu89:
Use the runner check results to show the cluster health instead of the current consul health.

image
image


fabriziosestito (approved):
LGTM! Thanks!

web/services/checks.go
```diff
@@ -12,11 +12,35 @@ import (
...
+}
+
+func (a *AggregatedCheckData) String() string {
+	if a.CriticalCount > 0 {
+		return CheckFailing
+	} else if a.PassingCount > 0 {
+		return CheckPassing
+	}
+
+	return ""
```
dottorblaster, line 34: In order to make illegal states unrepresentable, do we think returning an empty string is semantically valid? Does/should this have a default value in our domain?

Thoughts? :D

arbulu89: Good point!
I think I will change it to `CheckUndefined` at least, which can still have an empty string.
We mostly use this value in https://github.com/trento-project/trento/blob/main/web/templates/blocks/health_icon.html.tmpl, which would use the `else` statement at the end.

I will try to put the `undefined` string as value, and see if it doesn't break anything (I will need to do some monkey testing, I don't trust much our UT to this level of details)

arbulu89: Improved at: https://github.com/trento-project/trento/pull/211/commits/f47001a0cdc64b4dda2e4af9364efebb9056c90c

Does it pass the filter? XD

dottorblaster: Thank you so much for the explanation! Much needed by a newbie like me 😅
Yes it does pass the filter for sure 😆

PR #212, opened 2021-09-03
Trigger rebuilding of trento in obs

aleksei-burlakov:
(EDITED)
The first commit adds a GitHub action that calls obs services to make a package and then submits it to the OBS.
The second commit is to correctly rebase the PR over the main. (Lately the example/*.yaml cheks were removed)

### DIFF
--- a/.github/workflows/submit-to-obs.yml
+++ b/.github/workflows/submit-to-obs.yml
@@ -1,0 +1,74 @@
+name: Rebuild trento in obs
+on: [push]
+env:
+  PACKAGE_NAME: trento
+  OBS_USER: ${{ secrets.OBS_USER }}
+  OBS_PASS: ${{ secrets.OBS_PASS }}
+  OBS_PROJECT: ${{ secrets.OBS_PROJECT}}
+  TARGET_PROJECT: ${{ secrets.TARGET_PROJECT}}
+  FOLDER: packaging/suse
+  REPOSITORY: ${{ github.repository }}
+jobs:
+  delivery:
+    runs-on: ubuntu-18.04
+    if: github.ref == 'refs/heads/main'
+    container:
+      image: shap/continuous_deliver
+      env:
+        OBS_UNSTABLE: https://download.opensuse.org/repositories/OBS:/Server:/Unstable/SLE_15_SP3/OBS:Server:Unstable.repo
+        OBS_UNSTABLE_KEY: https://download.opensuse.org/repositories/OBS:/Server:/Unstable/SLE_15_SP3/repodata/repomd.xml.key
+        GITHUB_OAUTH_TOKEN:  ${{ secrets.GITHUB_TOKEN }}
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+    - name: configure OSC
+    # OSC credentials must be configured beforehand as the HOME variables cannot be changed from /github/home
+    # that is used to run osc commands
+      run: |
+        /scripts/init_osc_creds.sh
+        mkdir -p $HOME/.config/osc
+        cp /root/.config/osc/oscrc $HOME/.config/osc
+    - name: update go and install obs-service-node_modules
+      run: |
+        rpm --import ${OBS_UNSTABLE_KEY}
+        zypper ar ${OBS_UNSTABLE}
+        zypper ref
+        zypper in -y obs-service-node_modules go
+        go version
+    - name: prepare tranto.changes file
+      run: |
+        VERSION=$(hack/get_version_from_git.sh)
+        TAG=$(echo $VERSION | cut -f1 -d+)
+        .ci/gh_release_to_obs_changeset.py $REPOSITORY -a shap-staff@suse.de -t $TAG -f $FOLDER/trento.changes
+    - name: prepare _service file
+      run: |
+        VERSION=$(hack/get_version_from_git.sh)
+        sed -i 's~%%REVISION%%~${{ github.sha }}~' $FOLDER/_service && \
+        sed -i 's~%%REPOSITORY%%~${{ github.repository }}~' $FOLDER/_service && \
+        sed -i 's~%%VERSION%%~'"${VERSION}"'~' $FOLDER/_service
+    - name: deliver package
+      run: cp $FOLDER/_service . && /scripts/upload.sh
+  submit:
+    needs: delivery
+    runs-on: ubuntu-18.04
+    if: ${{ github.event_name != 'pull_request' }}
+    container:
+      image: shap/continuous_deliver
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+    - name: configure OSC
+      run: |
+        /scripts/init_osc_creds.sh
+        mkdir -p $HOME/.config/osc
+        cp /root/.config/osc/oscrc $HOME/.config/osc
+    - name: prepare _service file
+      run: |
+        VERSION=$(hack/get_version_from_git.sh)
+        sed -i 's~%%REVISION%%~${{ github.sha }}~' $FOLDER/_service && \
+        sed -i 's~%%REPOSITORY%%~${{ github.repository }}~' $FOLDER/_service && \
+        sed -i 's~%%VERSION%%~'"${VERSION}"'~' $FOLDER/_service
+    - name: submit package
+      run: cp $FOLDER/_service . && /scripts/submit.sh
--- a/packaging/suse/_service
+++ b/packaging/suse/_service
@@ -1,0 +1,27 @@
+<services>
+    <service name="tar_scm" mode="disabled">
+        <param name="url">https://github.com/%%REPOSITORY%%.git</param>
+        <param name="scm">git</param>
+        <param name="revision">%%REVISION%%</param>
+        <param name="exclude">.git</param>
+        <param name="exclude">.github</param>
+        <param name="extract">web/frontend/package.json</param>
+        <param name="extract">web/frontend/package-lock.json</param>
+        <param name="exclude">web/frontend/package-lock.json</param>
+        <param name="versionformat">%%VERSION%%</param>
+        <param name="filename">trento</param>
+    </service>
+    <service name="set_version" mode="disabled">
+        <param name="file">trento.spec</param>
+    </service>
+    <service name="recompress" mode="disabled">
+        <param name="file">*.tar</param>
+        <param name="compression">gz</param>
+    </service>
+    <service name="node_modules" mode="disabled">
+        <param name="cpio">node_modules.obscpio</param>
+        <param name="output">node_modules.spec.inc</param>
+        <param name="source-offset">10000</param>
+    </service>
+    <service name="go_modules" mode="disabled" />
+</services>
--- a/packaging/suse/trento.spec
+++ b/packaging/suse/trento.spec
@@ -1,0 +1,114 @@
+#
+# spec file for package trento
+#
+# Copyright (c) 2021 SUSE LLC
+#
+# All modifications and additions to the file contributed by third parties
+# remain the property of their copyright owners, unless otherwise agreed
+# upon. The license for this file, and modifications and additions to the
+# file, is the same license as for the pristine package itself (unless the
+# license for the pristine package is not an Open Source License, in which
+# case the license is the MIT License). An "Open Source License" is a
+# license that conforms to the Open Source Definition (Version 1.9)
+# published by the Open Source Initiative   .
+
+# Please submit bugfixes or comments via https://bugs.opensuse.org/
+#
+
+
+Name:           trento
+# Version will be processed via set_version source service
+Version:        0
+Release:        0
+License:        Apache-2.0
+Summary:        An open cloud-native web console improving on the life of SAP Applications administrators .
+Group:          System/Monitoring
+URL:            https://github.com/trento-project/trento
+Source:         %{name}-%{version}.tar.gz
+Source1:        vendor.tar.gz
+Source2:        node_modules.spec.inc
+Source3:	package.json
+%include %_sourcedir/node_modules.spec.inc
+ExclusiveArch:  aarch64 x86_64 ppc64le s390x
+BuildRoot:      %{_tmppath}/%{name}-%{version}-build
+BuildRequires:  golang-packaging
+BuildRequires:  golang(API) = 1.16
+BuildRequires:  npm
+BuildRequires:  local-npm-registry
+Provides:       trento = %{version}-%{release}
+
+%{go_nostrip}
+
+%description
+An open cloud-native web console improving on the life of SAP Applications administrators.
+
+Trento is a city on the Adige River in Trentino-Alto Adige/Suedtirol in Italy. [...] It is one of the nation''s wealthiest and most prosperous cities, [...] often ranking highly among Italian cities for quality of life, standard of living, and business and job opportunities. (source)
+
+This project is a reboot of the "SUSE Console for SAP Applications", also known as the Blue Horizon for SAP prototype, which is focused on automated infrastructure deployment and provisioning for SAP Applications.
+
+As opposed to that first iteration, this new one will focus more on operations of existing clusters, rather than deploying new one.
+
+%prep
+%setup -q            # unpack project sources
+%setup -q -T -D -a 1 # unpack go dependencies in vendor.tar.gz, which was prepared by the source services
+
+cp %SOURCE3 .
+local-npm-registry %{_sourcedir} install --with=dev
+
+%define shortname trento
+
+%build
+
+mv node_modules web/frontend/
+make build
+
+%install
+
+# Install the binary.
+install -D -m 0755 %{shortname} "%{buildroot}%{_bindir}/%{shortname}"
+
+# Install the systemd unit
+install -D -m 0644 hack/trento-agent.service %{buildroot}%{_unitdir}/trento-agent.service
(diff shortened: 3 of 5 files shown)


stefanotorresi (changes requested):
@aleksei-burlakov to be honest, I'm having a lot of second thoughts about going through with the Ubuntu approach.

Copy-pasting the external code of the OBS source services is a major pitfall which I would strongly oppose to, and I just don't see the advantages, when we already have a [dedicated Docker image](https://github.com/arbulu89/continuous-delivery) that solved this problem for us.

By reusing the same approach we already have in place in the other projects, the entire workflow would be reduced just 3 steps:
- checkout
- initialize credentials
- execute OSC via already provided scripts

You wouldn't need to install any dependencies and we would be more in control of them, since they would be pinned and fixed, instead of potentially changing at every run of the CI job.

One last minor remark: in all the other projects, we usually put all this stuff under `packaging/suse` so that we allow for different packaging mechanism for other distros in the future.

stefanotorresi:
I also have one more question: how would we generate the OBS-compliant changelog?

obs/oscrc
```diff
@@ -0,0 +1,6 @@
+[general]
+apiurl = https://api.opensuse.org
+
+[https://api.opensuse.org]
+user=shapbot
```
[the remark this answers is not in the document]
aleksei-burlakov: I reuse the /scripts/init_osc_creds.sh in the docker image as described here https://confluence.suse.com/display/haslesforsap/Automatic+OBS+packages+updates+with+GitHub+and+GitHub+Actions

obs/specs/_service
```diff
@@ -0,0 +1,27 @@
...
+    <service name="tar_scm" mode="disabled">
+        <param name="url">https://github.com/trento-project/trento.git</param>
+        <param name="scm">git</param>
+        <param name="revision">main</param>
+        <param name="exclude">.git</param>
+        <param name="exclude">.github</param>
+        <param name="extract">web/frontend/package.json</param>
+        <param name="extract">web/frontend/package-lock.json</param>
+        <param name="exclude">web/frontend/package-lock.json</param>
+        <param name="versionformat">0.4.0+%ct.%h</param>
```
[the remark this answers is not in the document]
aleksei-burlakov: It's sedded as in https://confluence.suse.com/display/haslesforsap/Automatic+OBS+packages+updates+with+GitHub+and+GitHub+Actions  now

aleksei-burlakov:
> I also have one more question: how would we generate the OBS-compliant changelog?

By a shell script, I guess. I think we should address it in the next PR.

aleksei-burlakov:
@here I have refactored the PR wrt to https://confluence.suse.com/display/haslesforsap/Automatic+OBS+packages+updates+with+GitHub+and+GitHub+Actions

stefanotorresi:
> > I also have one more question: how would we generate the OBS-compliant changelog?
> 
> By a shell script, I guess. I think we should address it in the next PR.

hey Aleksei, WRT this, maybe you can copy this:
https://github.com/SUSE/sap_host_exporter/blob/master/.ci/gh_release_to_obs_changeset.py

obs/specs/_service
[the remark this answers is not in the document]
aleksei-burlakov: > > > I also have one more question: how would we generate the OBS-compliant changelog?
> > 
> > 
> > By a shell script, I guess. I think we should address it in the next PR.
> 
> hey Aleksei, WRT this, maybe you can copy this:
> https://github.com/SUSE/sap_host_exporter/blob/master/.ci/gh_release_to_obs_changeset.py
> 
> Rest of the changes mostly LGTM, I will give it a closer look as soon as I can.

The problem with tockens is that the **$ obs token -o runservice** will fail, because the obs cannot download the vendors from outside.

aleksei-burlakov:
> > > I also have one more question: how would we generate the OBS-compliant changelog?
> > 
> > 
> > By a shell script, I guess. I think we should address it in the next PR.
> 
> hey Aleksei, WRT this, maybe you can copy this:
> https://github.com/SUSE/sap_host_exporter/blob/master/.ci/gh_release_to_obs_changeset.py
> 
> Rest of the changes mostly LGTM, I will give it a closer look as soon as I can.

The problem with tockens is that the **$ obs token -o runservice** will fail, because the obs cannot download the vendors from outside.

stefanotorresi:
> The problem with tockens is that the **$ obs token -o runservice** will fail, because the obs cannot download the vendors from outside.

OBS tokens are not related to the changelog script I linked above: that uses GitHub tokens, which are available within GitHub actions jobs!

.github/workflows/submit-to-obs.yml
```diff
@@ -0,0 +1,74 @@
...
+  OBS_PROJECT: ${{ secrets.OBS_PROJECT}}
+  TARGET_PROJECT: ${{ secrets.TARGET_PROJECT}}
+  FOLDER: packaging/suse
+jobs:
+  validation:
+    runs-on: ubuntu-18.04
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
```
stefanotorresi, line 16: I guess this is not needed, right?

.github/workflows/submit-to-obs.yml
```diff
@@ -0,0 +1,74 @@
...
+
+  delivery:
+    needs: validation
+    runs-on: ubuntu-18.04
+    if: ${{ github.event_name != 'pull_request' }}
```
stefanotorresi, line 21: this should actually only happen on the main branch:
```suggestion
    if: github.ref == 'refs/heads/main'
```

.github/workflows/submit-to-obs.yml
```diff
@@ -0,0 +1,74 @@
...
+jobs:
+  validation:
+    runs-on: ubuntu-18.04
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+  delivery:
+    needs: validation
```
stefanotorresi, line 19: we don't really need any validation step

.github/workflows/submit-to-obs.yml
```diff
@@ -0,0 +1,74 @@
...
+      run: |
+        /scripts/init_osc_creds.sh
+        mkdir -p $HOME/.config/osc
+        cp /root/.config/osc/oscrc $HOME/.config/osc
+    - name: update go and install obs-service-node_modules
+      run: |
+        rpm --import ${OBS_UNSTABLE_KEY}
+        zypper ar ${OBS_UNSTABLE}
+        zypper ref
+        zypper in -y obs-service-node_modules go
```
stefanotorresi, line 43: Ideally we wouldn't need to do any of this and we would instead update the Docker image, but we can do this another time.
That said, we need to pin to a specific go version, otherwise this is going to blow up in our faces sooner or later.

.github/workflows/submit-to-obs.yml
```diff
@@ -0,0 +1,74 @@
...
+        go version
+    - name: deliver package
+      run: |
+        echo "github.sha = ${{ github.sha }}"
+        echo "github.repository = ${{ github.repository }}"
```
stefanotorresi, line 48: I guess you put these there just for debugging?

.github/workflows/submit-to-obs.yml
```diff
@@ -0,0 +1,74 @@
...
+        sed -i 's~%%VERSION%%~${{ github.sha }}~' $FOLDER/_service && \
```
stefanotorresi, line 49: this should not be `VERSION`, it should be `REVISION` instead
```suggestion
        sed -i 's~%%REVISION%%~${{ github.sha }}~' $FOLDER/_service && \
```

.github/workflows/submit-to-obs.yml
```diff
@@ -0,0 +1,74 @@
...
+      with:
+        fetch-depth: 0
+    - name: configure OSC
+      run: |
+        /scripts/init_osc_creds.sh
+        mkdir -p $HOME/.config/osc
+        cp /root/.config/osc/oscrc $HOME/.config/osc
+    - name: submit package
+      run: |
+       sed -i 's~%%VERSION%%~${{ github.sha }}~' $FOLDER/_service && \
```
stefanotorresi, line 71: this should not be `VERSION`, it should be `REVISION` instead
```suggestion
        sed -i 's~%%REVISION%%~${{ github.sha }}~' $FOLDER/_service && \
```

packaging/suse/_service
```diff
@@ -0,0 +1,27 @@
+<services>
+    <service name="tar_scm" mode="disabled">
+        <param name="url">https://github.com/%%REPOSITORY%%.git</param>
+        <param name="scm">git</param>
+        <param name="revision">%%VERSION%%</param>
```
stefanotorresi, line 5: this should not be `VERSION`, it should be `REVISION` instead
```suggestion
        <param name="revision">%%REVISION%%</param>
```

packaging/suse/_service
```diff
@@ -0,0 +1,27 @@
...
+        <param name="exclude">.git</param>
+        <param name="exclude">.github</param>
+        <param name="extract">web/frontend/package.json</param>
+        <param name="extract">web/frontend/package-lock.json</param>
+        <param name="exclude">web/frontend/package-lock.json</param>
+        <param name="versionformat">0.4.0+%ct.%h</param>
```
stefanotorresi, line 11: This is still hard-coded, and this is the actual `%%VERSION%%` parameter. 😉 
It should probably be sedded as well, like this: https://github.com/SUSE/sap_host_exporter/blob/aaa32755fedd8b687dfbc6373a95dc6224cf18f6/Makefile#L74

packaging/suse/trento.spec
```diff
@@ -0,0 +1,114 @@
...
+# license that conforms to the Open Source Definition (Version 1.9)
+# published by the Open Source Initiative   .
+
+# Please submit bugfixes or comments via https://bugs.opensuse.org/
+#
+
+
+Name:           trento
+# Version will be processed via set_version source service
+Version:        0.4.0+1630322269.39b75d7
```
stefanotorresi, line 21: I think we can leave this to `0` to avoid confusion with the comment.

stefanotorresi (changes requested):
Not quite there yet 🛠️

aleksei-burlakov:
> > The problem with tockens is that the **$ obs token -o runservice** will fail, because the obs cannot download the vendors from outside.
> 
> OBS tokens are not related to the changelog script I linked above: that uses GitHub tokens, which are available within GitHub actions jobs!

Ah, ok, I thought it was about that we wanted to make the whole thing by tokens.

.github/workflows/submit-to-obs.yml
aleksei-burlakov: Nope, it's there just in case we come up with some ideas about validation.

.github/workflows/submit-to-obs.yml
aleksei-burlakov: :+1:

.github/workflows/submit-to-obs.yml
aleksei-burlakov: Ok

.github/workflows/submit-to-obs.yml
aleksei-burlakov: :+1:

.github/workflows/submit-to-obs.yml
aleksei-burlakov: Yes, I will remove them.

aleksei-burlakov:
@stefanotorresi thank you for reviewing, it's all correct that you say about the version and revision. I have changed in another branch and forgot to bring it to the PR branch.

.github/workflows/submit-to-obs.yml
aleksei-burlakov: Correct

.github/workflows/submit-to-obs.yml
aleksei-burlakov: Correct

packaging/suse/_service
aleksei-burlakov: Correct

packaging/suse/_service
aleksei-burlakov: Correct

packaging/suse/trento.spec
aleksei-burlakov: Ок

aleksei-burlakov:
@stefanotorresi I have refactored the PR wrt to the comments.

- `gh_release_to_obs_changeset.py` generated the `trento.changes`
- `VERSION` and `REVISION` are used correctly in the `_service`
- `Version: 0` in the trento.spec (it's defaulted to 0, and later set by the obs services)
- `echo ...` removed from the github action
- `valudation`-step is removed from the github action.

Besides, I have removed the yaml checks from the trento.spec in the second commit.

packaging/suse/trento.spec
```diff
@@ -0,0 +1,106 @@
...
+# Please submit bugfixes or comments via https://bugs.opensuse.org/
+#
+
+
+Name:           trento
+# Version will be processed via set_version source service
+Version:        0
+Release:        0
+License:        Apache-2.0
+Summary:        An open cloud-native web console improving on the life of SAP Applications administrators .
```
arbulu89, line 24: It looks like there is an extra space here before the dot

packaging/suse/trento.spec
```diff
@@ -0,0 +1,106 @@
...
+Group:          System/Monitoring
+URL:            https://github.com/trento-project/trento
+Source:         %{name}-%{version}.tar.gz
+Source1:        vendor.tar.gz
```
arbulu89, line 28: Where are these source files coming from?
Are they created during the CI process?
If they are pushed manually to the repo, I would at least put a comment in top of them

packaging/suse/trento.spec
```diff
@@ -0,0 +1,106 @@
...
+BuildRequires:  npm
+BuildRequires:  local-npm-registry
+Provides:       trento = %{version}-%{release}
+
+%{go_nostrip}
+
+%description
+An open cloud-native web console improving on the life of SAP Applications administrators.
+
+Trento is a city on the Adige River in Trentino-Alto Adige/Suedtirol in Italy. [...] It is one of the nation''s wealthiest and most prosperous cities, [...] often ranking highly among Italian cities for quality of life, standard of living, and business and job opportunities. (source)
```
arbulu89, line 45: Double quote here in `nation` word (`''`)

packaging/suse/trento.spec
```diff
@@ -0,0 +1,106 @@
...
+%service_del_preun trento-agent.service
+
+%postun
+%service_del_postun trento-agent.service
+
+%files
+%defattr(-,root,root)
+%doc *.md
+%doc docs/*.md
+%if 0%{?suse_version} >= 1500
```
arbulu89, line 97: This `license` split will bring issues in the future. I would vote to only use the `license` tag` and remove the `doc` options, as this is only used in previous versions to SLE12SP3

packaging/suse/trento.spec
```diff
@@ -0,0 +1,106 @@
...
+%license LICENSE
+%else
+%doc LICENSE
+%endif
+%{_bindir}/%{shortname}
+%{_unitdir}/trento-agent.service
+%{_sbindir}/rc%{name}
```
arbulu89, line 104: What is this line `rc%{name}` doing?

packaging/suse/trento.spec
```diff
@@ -0,0 +1,106 @@
...
+make build
+
+%install
+
+# Install the binary.
+install -D -m 0755 %{shortname} "%{buildroot}%{_bindir}/%{shortname}"
+
+# Install the systemd unit
+install -D -m 0644 hack/trento-agent.service %{buildroot}%{_unitdir}/trento-agent.service
+sed -i "s+/srv/trento/trento agent start+trento agent start+g" %{buildroot}%{_unitdir}/trento-agent.service
```
arbulu89, line 72: I don't quite understand this line.
If we need to do such a thing, wouldn't be better to have the correct `trento-agent.service` after all?
It can be changed in the future though

arbulu89 (commented):
@aleksei-burlakov I have not checked fully the CI file. I guess that you already worked on this with @stefanotorresi .
In any case, where are we submitting the package after uploading it to our devel repo? To tumbleweed factory?

I have put some minor comments in the spec file

packaging/suse/trento.spec
aleksei-burlakov: Good catch)

packaging/suse/trento.spec
aleksei-burlakov: They are created by the obs services, when calling `$ osc service dr` command. We run this command in the CI, but it could be run manually the same.

packaging/suse/trento.spec
aleksei-burlakov: It's a rudiment. Removed.

packaging/suse/trento.spec
aleksei-burlakov: Correct. I will change the trento-agent.service as well. Besides, as discussed with @fabriziosestito I will move the `trento-agent.service` file from the `hack` folder to the root of the project. I think it's no bad to do it right in this PR.

aleksei-burlakov: @fabriziosestito @arbulu89 the `trento-agent.service` needs another improvement. Since there are no more bench-common yaml checks, the trento-agent service will fail to start.

packaging/suse/trento.spec
arbulu89: Nice!

arbulu89 (approved):

PR #213, opened 2021-09-06
Ansible records usage improvements

arbulu89:
This PR fixes 3 things:
- Use `become` option to provide root privileges to ansible to run the queries. Otherwise, the checks are executed with the user used in the connection, which might not have the rights to run our checks
- Send the empty ARA record at the beginning only if all the hosts are unreachable. Otherwise, it is being sent always, and this makes the Web show an empty health results during the ansible execution
- Fix the annoying critical health container padding...

Now the critical counter has the same padding than the 2 others
image

### DIFF
--- a/runner/ansible/check.yml
+++ b/runner/ansible/check.yml
@@ -1,5 +1,6 @@
 - hosts: all
   gather_facts: false
+  become: yes
 
   vars:
     ara_playbook_labels:
--- a/runner/ansible/check.yml
+++ b/runner/ansible/check.yml
@@ -1,5 +1,6 @@
 - hosts: all
   gather_facts: false
+  ignore_errors: yes
   become: yes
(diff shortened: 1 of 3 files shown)


brett060102 (approved):
The changes look good to me. So, long as you are OK with all playbooks being run as root.
The more normal convention is to use become in the play that needs to be run as root.
So, even plays run on the local host will be run as root.

arbulu89:
> The changes look good to me. So, long as you are OK with all playbooks being run as root.
> The more normal convention is to use become in the play that needs to be run as root.
> So, even plays run on the local host will be run as root.

Actually, that's a good point. Maybe, we will need to make the `become` more gradular at some point, but by now, let's go this way. If get some further issue we will make it task specific

PR #217, opened 2021-09-07
Add cluster connection settings user input in cluster modal

arbulu89:
Implement the option to customize the ansible ssh connection settings, specifically the user that is used during the ssh connections.
This is done through the cluster settings modal.

It follows the next logic:
- User stored data goes to: `trento/v0/clusters/mycluster/user_settings/connection/` as map
- If the data is not stored, the code tries to use the Azure metadata information to get the OsProfile admin user name
- Ansible inventory is changed based on this logic, adding the `ansible_user` value

Demo:

connection_settings

Follow up issues:
https://github.com/trento-project/trento/issues/218
https://github.com/trento-project/trento/issues/219

### DIFF
--- a/web/clusters_test.go
+++ b/web/clusters_test.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"net/url"
+	"path"
 	"regexp"
 	"strconv"
 	"strings"
@@ -15,6 +16,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/tdewolff/minify/v2"
 	"github.com/tdewolff/minify/v2/html"
+	"github.com/trento-project/trento/internal/cloud"
 	"github.com/trento-project/trento/internal/consul"
 	"github.com/trento-project/trento/internal/consul/mocks"
 	"github.com/trento-project/trento/web/models"
@@ -70,6 +72,14 @@ func clustersListMap() map[string]interface{} {
 									},
 								},
 							},
+						},
+					},
+					"Nodes": []interface{}{
+						map[string]interface{}{
+							"Name": "test_node_1",
+						},
+						map[string]interface{}{
+							"Name": "test_node_2",
 						},
 					},
 					"NodeAttributes": map[string]interface{}{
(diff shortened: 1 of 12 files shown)


arbulu89:
@fabriziosestito @rtorrero @aleksei-burlakov @dottorblaster @stefanotorresi 
I need a review here to complete all the changes for the 0.4.0 release

web/clusters_test.go
```diff
@@ -372,6 +382,19 @@ func checksResultByHost() map[string]*services.AggregatedCheckData {
...
+		"provider": cloud.Azure,
+		"metadata": &cloud.AzureMetadata{
+			Compute: cloud.Compute{
+				OsProfile: cloud.OsProfile{
+					AdminUserName: fmt.Sprintf("defuser%d", userIndex),
+				},
+			},
+		},
+	}
+}
```
dottorblaster, line 396: Very declarative, I like it!

dottorblaster (approved):

fabriziosestito (commented):
@arbulu89  I tried to run the PR on gollum, the default user is an empty column in the modal.
Is this desired or we want to show "root" in the case we are not running on Azure?

image

P.S: Please don't laugh at my cluster petname

arbulu89:
> @arbulu89 I tried to run the PR on gollum, the default user is an empty column in the modal.
> Is this desired or we want to show "root" in the case we are not running on Azure?
> 
> image
> 
> P.S: Please don't laugh at my cluster petname

This value is coming from the cloud metadata, so in libvirt we don't have it. Behind the scenes, if it is not being shown there `root` will be used.
The thing is that there is not any way of getting this default value, so using `root` would be a hardcoded value on the code.
We could put it there to have at least some value in the web ui

fabriziosestito:
> > @arbulu89 I tried to run the PR on gollum, the default user is an empty column in the modal.
> > Is this desired or we want to show "root" in the case we are not running on Azure?
> > image
> > P.S: Please don't laugh at my cluster petname
> 
> This value is coming from the cloud metadata, so in libvirt we don't have it. Behind the scenes, if it is not being shown there `root` will be used.
> The thing is that there is not any way of getting this default value, so using `root` would be a hardcoded value on the code.
> We could put it there to have at least some value in the web ui

Oh ok, yes I would go with the hardcoded one in case we are not on azure, just to be consistent with the UI

arbulu89:
> > > @arbulu89 I tried to run the PR on gollum, the default user is an empty column in the modal.
> > > Is this desired or we want to show "root" in the case we are not running on Azure?
> > > image
> > > P.S: Please don't laugh at my cluster petname
> > 
> > 
> > This value is coming from the cloud metadata, so in libvirt we don't have it. Behind the scenes, if it is not being shown there `root` will be used.
> > The thing is that there is not any way of getting this default value, so using `root` would be a hardcoded value on the code.
> > We could put it there to have at least some value in the web ui
> 
> Oh ok, yes I would go with the hardcoded one in case we are not on azure, just to be consistent with the UI

"root" default user now is showed in the modal if there is not any cloud metadata found:
image

fabriziosestito (approved):

PR #221, opened 2021-09-07
Improve tags ux and add autocomplete

fabriziosestito:
This PR improves the tags UX and adds the autocomplete feature.
It also improves the internal tags package and embeds the assets in the assets pipeline instead of referencing them from an external CDN.

Peek 2021-09-07 16-08


arbulu89 (approved):
@fabriziosestito It looks great.
Neat code and the demo looks fantastic.
Nothing to complain about

PR #222, opened 2021-09-08
Remove hardcoded SID prefix (prd) from attributes keys

fabriziosestito:
Fixes broken attribute fetching when `SID != "PRD"`

### DIFF
--- a/web/clusters.go
+++ b/web/clusters.go
@@ -26,6 +26,7 @@ type Node struct {
 	Ip         string
 	Health     string
 	VirtualIps []string
+	SID        string
 }
 
 type Resource struct {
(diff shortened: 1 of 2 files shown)


arbulu89 (changes requested):
@fabriziosestito There are still some changes required in (2 entries at least): https://github.com/trento-project/trento/blob/main/web/templates/cluster_hana.html.tmpl

For example: https://github.com/trento-project/trento/blob/main/web/templates/cluster_hana.html.tmpl#L31

fabriziosestito:
> @fabriziosestito There are still some changes required in (2 entries at least): https://github.com/trento-project/trento/blob/main/web/templates/cluster_hana.html.tmpl
> 
> For example: https://github.com/trento-project/trento/blob/main/web/templates/cluster_hana.html.tmpl#L31

Ah yes i realized, you were too fast to review!

arbulu89 (approved):
Now looks good!
In my workaround, I implemented the new 2 methods for the `Node` struct, as until now the data was always obtained from the 1st node. I guess it doesn't make much difference.

PD: Maybe we could wrap the `strings.ToLower` thing in a `Node` struct method, to not have it duplicated all over the place. Anyway, it doesn't make a major difference neither. Up to you!

arbulu89 (approved):
Great!

PR #224, opened 2021-09-09
Add 1.1.9 and 1.1.9.runtime checks to the ansible runner

arbulu89:
Add missing 1.1.9 and 1.1.9.runtime checks to the ansible runner.
Based on: https://github.com/trento-project/trento/blob/main/examples/generic-azure.yaml#L240

### DIFF
--- a/runner/ansible/roles/1.1.9.runtime/tasks/main.yml
+++ b/runner/ansible/roles/1.1.9.runtime/tasks/main.yml
@@ -6,9 +6,9 @@
 
 - name: "{{ test_id }}.check"
   shell: |
-    corosync-cmapctl | grep totem.interface.0 || exit 1
-    corosync-cmapctl | grep totem.interface.1 || exit 1
-    exit 0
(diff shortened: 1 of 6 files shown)


runner/ansible/roles/1.1.9.runtime/tasks/main.yml
```diff
@@ -0,0 +1,23 @@
+---
+
+- name: Set Test ID
+  set_fact:
+    test_id: '1.1.9.runtime'
+
+- name: "{{ test_id }}.check"
+  shell: |
+    INTERFACE_COUNT=$(corosync-cmapctl | grep totem.interface\\..*\.ttl | wc -l)
```
arbulu89, line 9: @brett060102 Check updated to be more robust

brett060102 (commented):
Changes test clean and look good. Is it OK having the interface count be hard coded? Would users ever want to have it be more than two? If no, this is cool.

brett060102 (approved):

PR #226, opened 2021-09-09
Trento agent one-line installer

fabriziosestito:
Iteration of PR: #152 

Adds the one line `curl-bashable` installer which

- Installs or upgrades dependencies
- Installs or upgrades, configure consul
- Adds consul systemd unit
- Installs or upgrade trento rpm 

Mandatory parameters were kept to the bare minimum and they are asked interactively if the user doesn't provide them as flags or env variables.

Updates the README with the quick installation instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,8 @@ of existing clusters, rather than deploying new one.
   - [Runtime dependencies](#runtime-dependencies)
   - [Build dependencies](#build-dependencies)
   - [Development dependencies](#development-dependencies)
-- [Installation](#installation)
+- [Quick-start installation](#quick-start-installation)
+- [Manual installation](#manual-installation)
 - [Running Trento](#running-trento)
   - [Consul](#consul)
   - [Trento Agents](#trento-agents)
@@ -88,13 +89,73 @@ Additionally, for the development we use:
 
 > See the [Development](#development) section for details on how to install `mockery`.
 
-# Installation
+# Quick-Start Installation
 
-## From binaries
+We provide an installation script to automatically install and update the latest version of Trento.
+
+## Trento Server Installation
+
+T.B.D.
+
+## Trento Agent Installation
+
+You can `curl | bash` if you want to live on the edge.
+
+```
+$ curl -sfL https://raw.githubusercontent.com/trento-project/trento/main/install.sh | sudo bash
+```
+
+Or you can fetch the script, and then execute it manually.
+
+```
+$ curl -O https://raw.githubusercontent.com/trento-project/trento/main/install.sh
+$ chmod 700 install.sh
+$ sudo ./install.sh
+```
+
+The script will ask you for two IP addresses.
+
+- `agent bind IP`: the private address to which the trento-agent should be bound for internal communications.
+  This is an IP address that should be reachable by the other hosts, including the trento server.
+  Note for the Pacemaker users: this IP address _should not be_ a floating IP.
+
+- `trento server IP`: the address where Trento server can be reached.
+
+You can pass these arguments as flags or env variables too:
+
+```
+$ curl -sfL https://raw.githubusercontent.com/trento-project/trento/main/install.sh | sudo bash -s - --agent-bind-ip=192.168.33.10 --server-ip=192.168.33.1
+```
+
+```
+$ AGENT_BIND_IP=192.168.33.10 SERVER_IP=192.168.33.1 sudo ./install.sh
+```
+
+### Start Trento Agent
+
+The installation script does not start the agent automatically.
+
+You can start it by simply:
+
+```
+$ sudo systemctl start trento-agent
+```
+
+Please make sure the server is running before starting the agent
+
+To enable the service execute:
+
+```
+$ sudo systemctl enable trento-agent
+```
+
+# Manual Installation
+
+## Pre-built binaries
 
 Pre-built statically linked binaries are made available via [GitHub releases](https://github.com/trento-project/trento/releases).
 
-## Manual
+## Compile from source
 
 You clone also clone and build it manually:
 
--- a/README.md
+++ b/README.md
@@ -102,15 +102,15 @@ T.B.D.
 You can `curl | bash` if you want to live on the edge.
 
 ```
-$ curl -sfL https://raw.githubusercontent.com/trento-project/trento/main/install.sh | sudo bash
+$ curl -sfL https://raw.githubusercontent.com/trento-project/trento/main/install-agent.sh | sudo bash
 ```
 
 Or you can fetch the script, and then execute it manually.
 
 ```
-$ curl -O https://raw.githubusercontent.com/trento-project/trento/main/install.sh
-$ chmod 700 install.sh
-$ sudo ./install.sh
+$ curl -O https://raw.githubusercontent.com/trento-project/trento/main/install-agent.sh
+$ chmod 700 install-agent.sh
+$ sudo ./install-agent.sh
 ```
 
 The script will ask you for two IP addresses.
@@ -124,11 +124,11 @@ The script will ask you for two IP addresses.
 You can pass these arguments as flags or env variables too:
 
 ```
-$ curl -sfL https://raw.githubusercontent.com/trento-project/trento/main/install.sh | sudo bash -s - --agent-bind-ip=192.168.33.10 --server-ip=192.168.33.1
+$ curl -sfL https://raw.githubusercontent.com/trento-project/trento/main/install-agent.sh | sudo bash -s - --agent-bind-ip=192.168.33.10 --server-ip=192.168.33.1
 ```
 
 ```
-$ AGENT_BIND_IP=192.168.33.10 SERVER_IP=192.168.33.1 sudo ./install.sh
+$ AGENT_BIND_IP=192.168.33.10 SERVER_IP=192.168.33.1 sudo ./install-agent.sh
 ```
 
 ### Start Trento Agent
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,182 @@
+#!/bin/bash
+# The script requires root permissions
+
+set -e
+
+if [ "$EUID" -ne 0 ]; then
+    echo "Please run as root."
+    exit
+fi
+
+function print_help() {
+    cat << END
+This is a trento-agent installer. Trento is a web-based graphical user interface
+that can help you deploy, provision and operate infrastructure for SAP Applications
+
+Usage:
+
+  sudo ./install.sh --agent-bind-ip <192.168.122.10> --server-ip <192.168.122.5>
+
+Arguments:
+  --agent-bind-ip   The private address to which the trento-agent should be bound for internal communications.
+                    This is an IP address that should be reachable by the other nodes, including the trento server.
+  --server-ip       The trento server ip.
+  --help            Print this help.
+END
+}
+
+case "$1" in
+--help)
+    print_help
+    exit 0
+    ;;
+esac
+
+ARGUMENT_LIST=(
+    "agent-bind-ip"
+    "server-ip"
+)
+
+opts=$(
+    getopt \
+        --longoptions "$(printf "%s:," "${ARGUMENT_LIST[@]}")" \
+        --name "$(basename "$0")" \
+        --options "" \
+        -- "$@"
+)
+
+eval set "--$opts"
+
+while [[ $# -gt 0 ]]; do
+    case "$1" in
+    --agent-bind-ip)
+        AGENT_BIND_IP=$2
+        shift 2
+        ;;
+
+    --server-ip)
+        SERVER_IP=$2
+        shift 2
+        ;;
+
+    *)
+        break
+        ;;
+    esac
+done
+
+if [ -z "$AGENT_BIND_IP" ]; then
+    read -rp "Please provide a bind IP for the agent: " AGENT_BIND_IP < /dev/tty
+fi
+if [ -z "$SERVER_IP" ]; then
+    read -rp "Please provide the server IP: " SERVER_IP < /dev/tty
+fi
+if [ -z "$NODE_NAME" ]; then
+    NODE_NAME="$HOSTNAME"
+fi
+
+TRENTO_REPO_KEY=${TRENTO_REPO_KEY:-"https://download.opensuse.org/repositories/devel:/sap:/trento/15.3/repodata/repomd.xml.key"}
+TRENTO_REPO=${TRENTO_REPO:-"https://download.opensuse.org/repositories/devel:/sap:/trento/15.3/devel:sap:trento.repo"}
+
+CONSUL_VERSION=1.9.6
+CONSUL_PATH=/srv/consul
+CONFIG_PATH="$CONSUL_PATH/consul.d"
+CONSUL_HCL_TEMPLATE='data_dir = "/srv/consul/data/"
+log_level = "DEBUG"
+datacenter = "dc1"
+ui = true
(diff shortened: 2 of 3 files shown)


install.sh
```diff
@@ -0,0 +1,182 @@
...
+        zypper in -y trento > /dev/null
+    fi
+}
+
+install_consul
+setup_consul
+install_trento
+
+echo -e "\e[92mDone.\e[97m"
+echo -e "You can now start trento-agent with: \033[1msystemctl start trento-agent\033[0m"
```
aleksei-burlakov, line 182: I would say, "Now you can start the trento-agent with", what @lee-martin would say?

fabriziosestito: yes it sounds better

README.md
```diff
@@ -88,13 +89,71 @@ Additionally, for the development we use:
...
+```
+$ curl -O https://raw.githubusercontent.com/trento-project/trento/main/install.sh
+$ chmod 700 install.sh
+$ sudo ./install.sh
+```
+
+The script will ask you for two IP addresses.
+
+- `agent bind IP`: the private address to which the trento-agent should be bound for internal communications.
+  This is an IP address that should be reachable by the other nodes, including the trento server.
```
aleksei-burlakov, line 119: Maybe "reachable from the other nodes". An exert required #@lee-martin .

README.md
```diff
@@ -88,13 +89,71 @@ Additionally, for the development we use:
...
+
+```
+$ AGENT_BIND_IP=192.168.33.10 SERVER_IP=192.168.33.1 sudo ./install.sh
+```
+
+### Start Trento Agent
+
+The installation script does not start the agent automatically.
+
+You can start it by simply:
```
aleksei-burlakov, line 138: In the script `install.sh:182` we say
`echo -e "You can now start trento-agent with: \033[1msystemctl start trento-agent\033[0m"`
We should be consistent either we start _with_, or we start _by_.
Besides, I would put _by_ in the end, like "You can start it simply by"

README.md
```diff
@@ -88,13 +89,71 @@ Additionally, for the development we use:
...
+
+```
+$ sudo systemctl start trento-agent
+```
+
+To enable the service:
```
aleksei-burlakov, line 144: I would add a verb _do_, "To enable the service _do_"

aleksei-burlakov:
It's almost perfect. I have left some comments about the language, but to be sure, we would need expertise from @lee-martin

lee-martin:
@aleksei-burlakov  & @fabriziosestito  For me the curl downloads don't even work, under https://raw.githubusercontent.com/trento-project/trento/main/install.sh  i get `404: Not Found` when running on my Azure clients.  Can you check the link

README.md
lee-martin: @aleksei-burlakov  @fabriziosestito 

I think I would propose something like the text below.  In general please avoid using the word "node", for Trento we use the word "hostname" to mean that, so we should be consistent.  Also, why should the Trento Agent be reachable by anything other than the Trento server? I'm not currently aware of any need for this.

"
- `agent bind IP`: the local IP address to which the trento-agent should bound to for Trento internal communications.
  This is IP address that must be reachable by the Trento server.
"

fabriziosestito:
> @aleksei-burlakov & @fabriziosestito For me the curl downloads don't even work, under https://raw.githubusercontent.com/trento-project/trento/main/install.sh i get `404: Not Found` when running on my Azure clients. Can you check the link

@lee-martin it is expected to not be working since the README assumes this PR to be merged in the main branch, to test it on your azure clients please use this link https://raw.githubusercontent.com/fabriziosestito/trento/agent-installer/install.sh

README.md
fabriziosestito: > @aleksei-burlakov @fabriziosestito
> 
> I think I would propose something like the text below. In general please avoid using the word "node", for Trento we use the word "hostname" to mean that, so we should be consistent. Also, why should the Trento Agent be reachable by anything other than the Trento server? I'm not currently aware of any need for this.
> 
> "
> 
>     * `agent bind IP`: the local IP address to which the trento-agent should bound to for Trento internal communications.
>       This is IP address that must be reachable by the Trento server.
>       "

@lee-martin not sure we can change the word "node" with the word "hostname" in this context, they have a completely different meaning.  Maybe we could find another word for it, "host" maybe?

About the second point:
Looking at the consul architecture it was my understanding that the agents need to contact each others because of consul gossip protocol, see here: https://www.consul.io/docs/architecture and here: https://learn.hashicorp.com/tutorials/consul/reference-architecture?utm_source=consul.io&utm_medium=docs#network-connectivity-details

stefanotorresi: "host" is the term to use instead of "node".
I agree that "hostname" doesn't fit here, as that's a property of a host.

lee-martin: @fabriziosestito  Yes, absolutely "host" is what my brain was thinking, but my fingers automatically typed hostname. In trento we of course use the word "host" (without name) - my typo, sorry.

lee-martin: > About the second point:
> Looking at the consul architecture it was my understanding that the agents need to contact each others because of consul gossip protocol, see here: https://www.consul.io/docs/architecture and here: https://learn.hashicorp.com/tutorials/consul/reference-architecture?utm_source=consul.io&utm_medium=docs#network-connectivity-details

@fabriziosestito  I understand your point from consul, but this has pretty major security implications for the actual SAP systems we will be managing. They should not all all be able to openly communicate with each other, e.g. prod and dev will frequently be separated. Ideally I would like to proceed that only the tento server speaks to the individual agents, @arbulu89 & @stefanotorresi  - do you see any issue with not allowing the agents to talk to each other? I would really like to avoid the intercommunication if possible.

install.sh
lee-martin: agree on "Now you can start the trento-agent with". thanks.

README.md
lee-martin: See the other comment, we agreed on "Now you can start the trento-agent with".

README.md
lee-martin: In would vote for "To enable the service execute"

README.md
```diff
@@ -88,13 +89,71 @@ Additionally, for the development we use:
...
+- `agent bind IP`: the private address to which the trento-agent should be bound for internal communications.
+  This is an IP address that should be reachable by the other nodes, including the trento server.
+  Note for the Pacemaker users: this IP address _should not be_ a floating IP.
+
+- `trento server IP`: the address where Trento server can be reached.
+
+You can pass these arguments as flags or env variables too:
+
+```
+$ curl -sfL https://raw.githubusercontent.com/trento-project/trento/main/install.sh | sudo bash -s - --agent-bind-ip=192.168.33.10 --server-ip=192.168.33.1
```
lee-martin, line 127: When I tested this on SLES15 SP2 this happened:

```
cloudadmin@vmiscsisrv:~> curl -sfL https://raw.githubusercontent.com/fabriziosestito/trento/agent-installer/install.sh |sudo bash -s - --agent-bind-ip=10.75.1.4 --server-ip=10.0.0.4
Installing trento-agent...
warning: Rebuilding outdated index databases
warning: Generating 18 missing index(es), please wait...
* Installing trento
Done.
You can now start trento-agent with: systemctl start trento-agent

cloudadmin@vmiscsisrv:~> sudo bash
vmiscsisrv:/home/cloudadmin # systemctl start trento-agent
Failed to start trento-agent.service: Unit trento-agent.service is not loaded properly: Exec format error.
See system logs and 'systemctl status trento-agent.service' for details.

vmiscsisrv:/home/cloudadmin # systemctl status trento-agent
● trento-agent.service - Trento agent service
   Loaded: error (Reason: Exec format error)
   Active: inactive (dead)

Sep 10 16:32:27 vmiscsisrv systemd[1]: /usr/lib/systemd/system/trento-agent.service:7: Executable path is not absolute: trento agent start --consul-config-dir=/srv/consul/consul.d /srv/trento/examples/generic-azure.yaml /srv/trento/examples/hana-scale-up-perf-optimized-azure.yaml
Sep 10 16:36:15 vmiscsisrv systemd[1]: /usr/lib/systemd/system/trento-agent.service:7: Executable path is not absolute: trento agent start --consul-config-dir=/srv/consul/consul.d /srv/trento/examples/generic-azure.yaml /srv/trento/examples/hana-scale-up-perf-optimized-azure.yaml
Sep 10 16:36:28 vmiscsisrv systemd[1]: /usr/lib/systemd/system/trento-agent.service:7: Executable path is not absolute: trento agent start --consul-config-dir=/srv/consul/consul.d /srv/trento/examples/generic-azure.yaml /srv/trento/examples/hana-scale-up-perf-optimized-azure.yaml
Sep 10 16:36:33 vmiscsisrv systemd[1]: /usr/lib/systemd/system/trento-agent.service:7: Executable path is not absolute: trento agent start --consul-config-dir=/srv/consul/consul.d /srv/trento/examples/generic-azure.yaml /srv/trento/examples/hana-scale-up-perf-optimized-azure.yaml
Sep 10 16:36:37 vmiscsisrv systemd[1]: /usr/lib/systemd/system/trento-agent.service:7: Executable path is not absolute: trento agent start --consul-config-dir=/srv/consul/consul.d /srv/trento/examples/generic-azure.yaml /srv/trento/examples/hana-scale-up-perf-optimized-azure.yaml

```

Also the extra dash in the sudo command (e.g. `bash -s - ` ) looks like it will be error prone when folks potentially read and type this.  I guess this is redirecting stdin? Maybe add a comment that this is whats happening?

README.md
```diff
@@ -88,13 +89,71 @@ Additionally, for the development we use:
...
+## Trento Server Installation
+
+T.B.D.
+
+## Trento Agent Installation
+
+You can `curl | bash` if you want to live on the edge.
+
+```
+$ curl -sfL https://raw.githubusercontent.com/trento-project/trento/main/install.sh | sudo bash
```
lee-martin, line 105: This installation worked fine on 15 SP3 , unlike https://github.com/trento-project/trento/pull/226/files#r706322513 on 15 SP2.  

On the successful 15 SP3 installation I then started the trento agent service WITHOUT the server running, this left systemd in a long timeout as the agent tried to reach the server.  Is it possible to add some feedback from the service to systemctl to know the agent is retrying its connection?  Or maybe we can reduce the timeout since it seemed a long time?

install.sh
```diff
@@ -0,0 +1,182 @@
...
+. /etc/os-release
+if [[ ! $PRETTY_NAME =~ "SUSE" ]]; then
+    echo "Operating system is not supported. Exiting."
+    exit 1
+fi
+
+echo "Installing trento-agent..."
+
+function install_deps() {
+    echo "* Installing dependencies... "
```
lee-martin, line 122: We should ask the user about installing dependencies before just doing it.

install.sh
```diff
@@ -0,0 +1,182 @@
...
+    fi
+
+    echo "$CONSUL_SERVICE_TEMPLATE" >/usr/lib/systemd/system/$CONSUL_SERVICE_NAME
+    systemctl daemon-reload
+}
+
+function install_trento() {
+    rpm --import "${TRENTO_REPO_KEY}" > /dev/null
+    path=${TRENTO_REPO%/*}/
+    if zypper lr --details | cut -d'|' -f9 | grep "$path" > /dev/null 2>&1; then
```
lee-martin, line 162: We should inform the user that we just added a repo and share the details on screen.

install.sh
lee-martin, line 182: I would add a comment here that the trento server needs starting first.

README.md
fabriziosestito: The systemd unit is provided by the RPM, @aleksei-burlakov  could you check it please?

lee-martin: I also tried the installer syntax `curl -sfL https://raw.githubusercontent.com/trento-project/trento/main/install.sh | sudo bash` on 15 SP2, that also failed:

```
cloudadmin@vmiscsisrv:~> cat /etc/os-release
NAME="SLES"
VERSION="15-SP2"
...

cloudadmin@vmiscsisrv:~> ip a
...
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group default qlen 1000
    link/ether 00:22:48:84:18:28 brd ff:ff:ff:ff:ff:ff
    inet 10.76.1.4/24 brd 10.76.1.255 scope global eth0
...

cloudadmin@vmiscsisrv:~> curl -sfL https://raw.githubusercontent.com/fabriziosestito/trento/agent-installer/install.sh |sudo bash
Please provide a bind IP for the agent: 10.76.1.4   
Please provide the server IP: 10.0.0.4
Installing trento-agent...
warning: Rebuilding outdated index databases
warning: Generating 18 missing index(es), please wait...
* Installing trento
Done.
You can now start trento-agent with: systemctl start trento-agent
cloudadmin@vmiscsisrv:~> sudo systemctl start trento-agent
Failed to start trento-agent.service: Unit trento-agent.service is not loaded properly: Exec format error.
See system logs and 'systemctl status trento-agent.service' for details.
```

So something different on 15 SP2 vs SP3?  The installer should work on all SLE15, so best test on SP1 and SP0 as well.

install.sh
```diff
@@ -0,0 +1,182 @@
...
+
+
+[Install]
+WantedBy=multi-user.target'
+
+. /etc/os-release
+if [[ ! $PRETTY_NAME =~ "SUSE" ]]; then
+    echo "Operating system is not supported. Exiting."
+    exit 1
+fi
```
lee-martin, line 117: Since we are seeing some installation issues even within the SLE15, we should only allow SLE15 installs and not just anything SUSE.

README.md
fabriziosestito: > Also the extra dash in the sudo command (e.g. bash -s - ) looks like it will be error prone when folks potentially read and type this. I guess this is redirecting stdin? Maybe add a comment that this is whats happening?

This is pretty common tho, see:
https://rancher.com/docs/k3s/latest/en/installation/install-options/how-to-flags/

aleksei-burlakov: > The systemd unit is provided by the RPM, @aleksei-burlakov could you check it please?

Yes, of course

install.sh
fabriziosestito: Changed in: we stop the installation if we don't find the tools we need and we ask the user to manually install them, this is a common approach in other installer script (see helm, k3s, etc)

README.md
fabriziosestito: I modified the script and the README to clarify that the server must be started before the client.

I wouldn't mess with the default consul retry timeouts just for this. If the user wants to know more about what a systemd unit is doing,  the obvious way would be to go and read the journal.

README.md
aleksei-burlakov: I've changed the rpm, the trento-agent.service will also start in sle 15.2, 15.1 and 15.0

install.sh
fabriziosestito: Taking granted that the problems with SLE15 were caused by the RPM systemd unit (see comments above) and now fixed, we use the script to install trento also in opensuse and leap for development purposes. 
Do we really want to limit the version to SLE15?

README.md
stefanotorresi: We do require intercommunication across all the Trento Agent peers, at the moment.
We might find other solutions to this, let's follow-up somewhere else.

install.sh
stefanotorresi: It's not required to use SLE: technically, we just need a SUSE distro because the scripts assumes `zypper` is available.

stefanotorresi (approved):

PR #228, opened 2021-09-13
Fix cluster list ordering in case of clusters with the same name

fabriziosestito:
Fix cluster list ordering in case of clusters with the same name


arbulu89 (approved):
One thousand thanks @fabriziosestito

Issue #194, opened 2021-08-12
Cloud Metadata Discovery fails when using dmidecode


diegoakechi:
VMs running on Azure will report `Microsoft Corporation` in case they are running on Hyper-V virtualization, what might not be the case all the time. 

Some users reports the following output on the dmidecode of their VMs that makes the Cloud Discover fail:

```~ # dmidecode -s bios-vendor
American Megatrends Inc.
```

With that, the safest way would be to use the Azure INstance Metada web service to confirm that the machine is really running on an Azure environment.

https://docs.microsoft.com/en-us/azure/virtual-machines/linux/instance-metadata-service?tabs=linux

As this is always a local call inside the VM (even that REST API), it should respond very fast not affecting the discovery time.

That said, it is important to make sure the discovery will not hang waiting for a long timeout in case the machine is not running on Azure, where the metadata API is not available.

petersatsuse:
MSFT provided us the command already to find out if its Azure or not for saptune last year.

To figure out whether it Is Azure VM, you can use below command: 

# dmidecode -s chassis-asset-tag
7783-7084-3265-9085-8269-3286-77

The result is ASCII encoded. When convert to a string, the following string shows up: MSFT AZURE VM

PR #229, opened 2021-09-13
Use the correct dmidecode field to identify cloud provider

dottorblaster:
This is a potential fix for #194, basically we just changed the `dmidecode` field to `chassis-asset-tag`.

### DIFF
--- a/internal/cloud/metadata.go
+++ b/internal/cloud/metadata.go
@@ -8,7 +8,8 @@ import (
 )
 
 const (
-	azure = "azure"
+	azure       = "azure"
+	azureDmiTag = "7783-7084-3265-9085-8269-3286-77"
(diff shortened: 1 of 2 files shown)


fabriziosestito (approved):
LGTM

arbulu89 (approved):
Nice!
Most probably we will need to improve this code when we start adding other cloud providers, but by now it should make the job!
PD: @dottorblaster Were you able to run it in a real azure setup?

dottorblaster:
> Were you able to run it in a real azure setup?
@arbulu89 along with @fabriziosestito we were able to replicate this and look into it using an agent of the demo environment for some minutes :sweat_smile: yeah, we tested it and it went ok with this

PR #234, opened 2021-09-13
Add the discovery-period flag to the agent

arbulu89:
Add the `discovery-period` flag to the agent cmd line execution, to make the discovery loop period configurable. The default value is 2 minutes.

With this, the user can select different periods, as having a low period (1 or 2 minutes) when there are many cluster (even with 3 clusters the performance starts suffereing), the web starts behaving poorly due to the lock mechanism.

This PR tries to be a workaround before having a better solution for the persistance layer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,6 +165,9 @@ To start the trento agent:
 ./trento agent start
 ```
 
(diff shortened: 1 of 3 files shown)


fabriziosestito (approved):

dottorblaster (approved):
LGTM

PR #239, opened 2021-09-14
Helm chart fixes

fabriziosestito:
A set of fixes for the helm chart:
- removes repository directive from local subcharts  deps
- changes the image registry to the trento-project one
- fixes the ingress value in the trento-web subchart
- adds Chart.lock


dottorblaster (approved):
It works on my k3s cluster

Issue #208, opened 2021-09-01
Remove the discovery TTL consul health checks


stefanotorresi:
We are misusing the Consul Health Check functionality to register a bunch of TTL checks which are not actually needed.

We can supersede these with a single `trento-agent` health check reporting the status of the agent itself (instead of the result of the HA Configuration Checks, which is going to be run by the Runner component instead).

Related to #89

fabriziosestito:
@stefanotorresi what should the `Trento Health` check depend on?

stefanotorresi:
> @stefanotorresi what should the `Trento Health` check depend on?

Well, my idea would be that if any of the discovery loops fails, than this single check fails with some output about the specific error that occurred.
IOW I would like all the Consul health checks that we currently have for the discovery loops consolidated under a single one.

That said, if you have other ideas on how to self-check the Trento Agent, I'm all ears!

fabriziosestito:
@stefanotorresi I have a PR ready for this, just holding it after the runner branch merge, i basically did what you just described but changed a little bit when the discoveries return errors (e.g. no error if nothing is discovered, just a negative result)

PR #238, opened 2021-09-14
Rename the "Checks" section in the single host page...

stefanotorresi:
...and remove the `checkId` column, to disambiguate with the HA checker feature and make abundantly clear that these things are about Trento internals.

Before:
Screenshot from 2021-09-14 17-38-17

After:
Screenshot from 2021-09-14 17-34-46

PR #240, opened 2021-09-14
Remove discovery health checks; add trento-agent health check

fabriziosestito:
This PR removes the discovery health checks and adds a single trento-agent health check, which fails in case of discovery error. The check is green as long as the any of the discovery loops doesn't _fail_.

It also changes the domain definition of _failed discovery_, till now we considered a discovery as failed also when the discovery was having a negative result, for example no cloud provider found would trigger a failed check.
Now we save the negative output on the consul check note, but we don't fail in this case.

Closes: #208

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -2,7 +2,9 @@ package agent
 
 import (
 	"context"
+	"fmt"
 	"os"
+	"strings"
 	"sync"
 	"time"
 
@@ -16,6 +18,8 @@ import (
 	"github.com/trento-project/trento/internal/hosts"
 	"github.com/trento-project/trento/version"
 )
+
+const trentoAgentCheckId = "trentoAgent"
 
 type Agent struct {
 	cfg            Config
@@ -105,9 +109,7 @@ func (a *Agent) Start() error {
 	var wg sync.WaitGroup
 
 	wg.Add(1)
-	// The Discover Loop is executing at a much slower pace than the Checker Loop
-	// and will keep namespaces in Key-Value Consul store updated with specific facts
-	// discovered on the node
+
 	go func(wg *sync.WaitGroup) {
 		log.Println("Starting Discover loop...")
 		defer wg.Done()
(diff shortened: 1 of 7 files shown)


agent/agent.go
```diff
@@ -180,28 +168,27 @@ func (a *Agent) registerConsulService() error {
...
-			}
+				result = fmt.Sprintf("Error while running discovery '%s': %s", d.GetId(), err)
+				status = consulApi.HealthCritical
 
-			err = a.consul.Agent().UpdateTTL(d.GetId(), result, status)
-			if err != nil {
-				log.Println("An error occurred while trying to update TTL with Consul:", err)
+				log.Errorln(result)
 			}
+			note = append(note, result)
```
stefanotorresi, line 182: nitpick: `note` is _very_ ambiguous as a variable name here, because the consul `HealthCheck` structure has a `Notes` field, but what we're actually populating later with the `UpdateTTL()` function is the `Output` field.

stefanotorresi (changes requested):

stefanotorresi:
@fabriziosestito you might want to handle the fact that now we only want to show the `trentoAgent` consul health check, so that this remains consistent with #238. Basically, I think we can totally remove the table and only show one single check.

I'm also wondering, how are multiple discovery failures displayed?

agent/agent.go
fabriziosestito: I see! changed to `output`

I was using note because the signature of UpdateTTL() calls it so, but i see the confusion also inside the consul API since the check definition has a note field as well
image

dottorblaster: maybe `resultNote`? `discoveryNote`?

dottorblaster: Oh, `output` is fine

arbulu89 (commented):
@fabriziosestito Besides Stafano's comments, could you attach a screenshot to see how this looks like?

fabriziosestito:
@stefanotorresi @arbulu89 

Here how it looks with the changes requested 

Passing:
image

Error:
image

No Agent Running:
image

stefanotorresi (approved):
👍🏼 not perfect, tbh, but good enough. Having separate health checks actually made the visualization much cleaner, but it was too much code overhead and also an abuse of the Consul feature.
This will have to change when we eventually get rid of Consul, anyway.

arbulu89:
@fabriziosestito You will need to rebase `main` branch and do the same operation with the new subscription health check

fabriziosestito:
> @fabriziosestito You will need to rebase `main` branch and do the same operation with the new subscription health check

@arbulu89 
I managed to rebase with the main branch, but i have a doubt.
Should we consider an error not finding `SUSEConnect`  in path making the check fail on other distribution or we could, for example, write a message that no subscription was discovered?

arbulu89:
> > @fabriziosestito You will need to rebase `main` branch and do the same operation with the new subscription health check
> 
> @arbulu89 I managed to rebase with the main branch, but i have a doubt. Should we consider an error not finding `SUSEConnect` in path making the check fail on other distribution or we could, for example, write a message that no subscription was discovered?

We have these scenarios:
- SUSEConnect exists and works fine 
- SUSEConnect exists but fail -> `No subscriptions were found. SUSEConnect command execution failed`
- SUSEConnect not available (other distros) -> `No subscriptions were found. SUSEConnect tool is not available`

What about something like this? We could maybe add the error message at the end of the error message too

agent/discovery/subscription.go
```diff
@@ -28,7 +30,11 @@ func (d SubscriptionDiscovery) GetId() string {
 func (d SubscriptionDiscovery) Discover() (string, error) {
 	subsData, err := subscription.NewSubscriptions()
 	if err != nil {
-		return "", err
+		if errors.Is(err, exec.ErrNotFound) {
+			return "No subscriptions were found. SUSEConnect tool is not available.", nil
```
arbulu89, line 34: We should raise this with an error. If SUSEConnect is not available, the agent cannot run properly, so this is a critical state

arbulu89 (changes requested):
Hey @fabriziosestito ,
Just a minor change requested

arbulu89 (approved):
Great!!!

By now this is the best we can have. If we have some business decision to raise an error if we don't have SLES_SAP subscription we will do the changes, but maybe by them we won't have consul at all XD

PR #241, opened 2021-09-15
Use distroless container / multistage Dockerfile

fabriziosestito:
This PR changes the current Dockerfile to use distroless containers (https://github.com/GoogleContainerTools/distroless) and multistage build for the go artifact, assets and adds python ansible dependencies.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+web/frontend/node_modules
+web/frontend/assets
+trento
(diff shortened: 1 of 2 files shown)


arbulu89 (approved):
Hi @fabriziosestito ,
The changes look to me, and if the `distroless` discussion already happened with @stefanotorresi , I don't have any objection at least

dottorblaster (approved):
super LGTM, go go go :dash:

PR #242, opened 2021-09-15
Add persistent volume claim to ara pod

fabriziosestito:
Adds a pvc to store the ara sqlite database in a persistent manner


arbulu89 (approved):
It looks good (talking from my ignorance, as I don't understand anything XD)

PR #244, opened 2021-09-15
Add Helm chart README and pin container version

fabriziosestito:
Pins the container version to the next release and adds a helm chart installation guide for the server component


arbulu89 (approved):
Go!!

PR #245, opened 2021-09-15
Use full version in web ui

fabriziosestito:
Uses the full version in the web ui instead of the short one

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -1,9 +1,3 @@
 package version
 
-import "strings"
-
 var Version string
-
-func GetShortVersion() string {
-	return strings.Split(Version, "+")[0]
-}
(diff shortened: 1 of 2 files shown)


dottorblaster (approved):
Sounds reasonable, I like that.

Do you think we should also reference the current commit sha?

stefanotorresi (approved):
you get to do the Clint move, today, pal!

stefanotorresi:
> Do you think we should also reference the current commit sha?

@dottorblaster yep, that's already in! 😁  https://github.com/trento-project/trento/blob/f3b524cf8bff977704ca16c0acb01710a123e785/Makefile#L1-L2

PR #247, opened 2021-09-15
Set ID tasks can be delegated to local host.

brett060102:
The tasks that set the test id can be delegated to localhost.

### DIFF
--- a/runner/ansible/roles/1.1.1/tasks/main.yml
+++ b/runner/ansible/roles/1.1.1/tasks/main.yml
@@ -4,6 +4,8 @@
   set_fact:
     test_id: '1.1.1'
     key_name: 'token'
+  delegate_to: localhost
+
 
 - name: "{{ test_id }}.check"
   lineinfile:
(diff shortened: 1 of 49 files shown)


runner/ansible/roles/1.1.1/tasks/main.yml
```diff
@@ -4,6 +4,8 @@
   set_fact:
     test_id: '1.1.1'
     key_name: 'token'
+  delegate_to: localhost
+
```
stefanotorresi, line 8: nit: one more line break than needed 🙂

stefanotorresi (approved):
LGTM, other than a mini nitpick

runner/ansible/roles/1.1.1/tasks/main.yml
brett060102: removed

arbulu89:
@brett060102 Closing this as these tasks are removed in https://github.com/trento-project/trento/pull/257

PR #248, opened 2021-09-16
Fix obs workflow so that the submit job runs correctly on releases

stefanotorresi:
...also renames few things here and there. 😬


arbulu89 (approved):
Looks good!
I guess we will need to run the operation manually for this time

aleksei-burlakov:
LGTM.

aleksei-burlakov (approved):

PR #249, opened 2021-09-16
Change: use devel:sap:trento:factory instead of devel:sap:trento

aleksei-burlakov:
The default repository has changed.


stefanotorresi:
I think we established that we would keep the "stable" one as default and let users override it to the factory ones if they want to get the latest development version.

PR #250, opened 2021-09-16
Build containers in CI/CD

fabriziosestito:
Uses github actions to build containers on every merge on main.

If a release event is detected a container with the release `tag` is built and pushed, otherwise the container the `rolling` tag container gets updated.

It also supports cache + manual dispatch.


dottorblaster (dismissed):
YES

.github/workflows/ci.yaml
```diff
@@ -194,3 +193,43 @@ jobs:
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
```
fabriziosestito, line 213: This needed to add the capability to use an external cache

.github/workflows/ci.yaml
```diff
@@ -194,3 +193,43 @@ jobs:
...
+
+      - name: Build and push Docker image
+        uses: docker/build-push-action@ad44023a93711e3deb337508980b4b5e9bcdc5dc
+        with:
+          context: .
+          push: true
+          tags: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ env.IMAGE_TAG }}
+          labels: ${{ steps.meta.outputs.labels }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
```
fabriziosestito, line 235: Uses GitHub Action cache

.github/workflows/ci.yaml
dottorblaster: Dude are you _really_ commenting your own PR? :joy:

fabriziosestito: aha yes i thought it would help reviewing since it is a magic `yaml`

stefanotorresi:
Since #248 is causing a conflict on the workflow file name here, I also realized that this workflow doesn't actually depend on the `build` job of the CI workflow, so we might as well have it in a completely separate file, like the one for the OBS delivery.
What do you think?

fabriziosestito:
> Since #248 is causing a conflict on the workflow file name here, I also realized that this workflow doesn't actually depend on the `build` job of the CI workflow, so we might as well have it in a completely separate file, like the one for the OBS delivery.
> What do you think?

@stefanotorresi we choose to depend on the `build` job since it performs the testing as well. it would be nice to have separate jobs for testing etc but I think it still means the job should be in the same workflow to be dependent to the test stage?

stefanotorresi:
Hmm, you make a good point, but the discussion then gets tricky: by the same logic, we shouldn't build broken code in OBS either.
Shall we consolidate all the builds in the same workflow where all the downstream jobs are blocked by the one where we execute the test, then?

fabriziosestito:
@stefanotorresi I think this is an option. if we want to keep separate workflows we will need to move the test steps in composite action, and repeat the step in every workflow that is needed, this way we repeat the step,yes, but the workflow will work in a parallel fashion.

stefanotorresi:
> EDIT: I'm leaning more toward the first option you proposed tho, it's easier to add a test step and let the other jobs depend on it, also they will work in parallel also this way.

all right then; unless the rest of the team has objections, go for it!

stefanotorresi (changes requested):
Let's consolidate all the CI jobs into a single workflow, with all the CD jobs depending on the test one.

rtorrero:
> Let's consolidate all the CI jobs into a single workflow, with all the CD jobs depending on the test one.

I'm on it already

stefanotorresi (approved):
Good. As discussed in Slack, I'll leave a follow-up to fork https://github.com/arbulu89/continuous-delivery in the trento-project org.

PR #251, opened 2021-09-16
Some improvements to the server installation on k3s through Helm

dottorblaster:
In the rush for the new release I think we forgot a couple things about the new k3s/k8s Helm chart setup for the Trento server :-)


fabriziosestito (approved):
good catch! thanks for checking this out

PR #253, opened 2021-09-17
Add server installer

fabriziosestito:
- Fixes some minor stuff in the helm chart
- Adds a server installer that can be used in the CI as well by overriding IMAGE_TAG env var
- Updates README with server installer instruction + other improvements

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,12 @@ of existing clusters, rather than deploying new one.
   - [Build dependencies](#build-dependencies)
   - [Development dependencies](#development-dependencies)
 - [Quick-start installation](#quick-start-installation)
+  - [Trento Server installation](#trento-server-installation)
+  - [Trento Agent installation](#trento-agent-installation)
 - [Manual installation](#manual-installation)
+  - [Pre-built binaries](#pre-built-binaries)
+  - [Compile from source](#compile-from-source)
+  - [Helm chart](#helm-chart)
 - [Running Trento](#running-trento)
   - [Consul](#consul)
   - [Trento Agents](#trento-agents)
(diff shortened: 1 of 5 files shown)


dottorblaster:
Could you resolve conflicts with `main`? :innocent: 

(don't hate me, I love you)

fabriziosestito:
> Could you resolve conflicts with `main`? innocent
> 
> (don't hate me, I love you)

you guys are too quick

README.md
```diff
@@ -88,102 +93,45 @@ Additionally, for the development we use:
 
 > See the [Development](#development) section for details on how to install `mockery`.
 
-# Quick-Start Installation
+# Quick-Start installation
 
-We provide an installation script to automatically install and update the latest version of Trento.
+We provide installations scripts for to automatically install and update the latest version of Trento.
```
arbulu89, line 98: I know that this is inherited, but in my opinion (and this came from our Doc team), we shouldn't use the `we` person here. I would re-write to something like: `The project provides installation scripts...`

install-server.sh
```diff
@@ -0,0 +1,122 @@
...
+readonly PROGNAME="./install-server.sh"
+
+usage() {
+    cat <<- EOF
+    usage: $PROGNAME options
+
+    Install Trento Server
+
+    OPTIONS:
+       -p --private-key         the private key used by the runner to connect to the hosts
```
arbulu89, line 15: Maybe we should clarify that we refer to a ssh key and that this key must be authorized in all the trento agents

README.md
```diff
@@ -88,102 +93,45 @@ Additionally, for the development we use:
...
+to bootstrap a complete Trento server component.
 
-Please refer to the [Trento Runner](#trento-runner) for more information.
-
-Install Trento server chart:
+You can `curl | bash` if you want to live on the edge.
 
 ```
-$ helm install trento . --set-file trento-runner.privateKey=/your/path/id_rsa_runner
+$ curl -sfL https://raw.githubusercontent.com/trento-project/trento/main/install-server.sh | bash
```
arbulu89, line 109: Can this work without providing the ssh key path?

arbulu89 (changes requested):
@fabriziosestito It looks an excellent work.

Anyway, there are some issues hehe:

This happens running the script.
```
Please provide the path of the runner private key: /root/.ssh/id_rsa
Installing trento-server on k3s...
Installing K3s...
Created symlink /etc/systemd/system/multi-user.target.wants/k3s.service → /etc/systemd/system/k3s.service.
cp: cannot create regular file '/root/.kube/config': No such file or directory
```

Finally, is this starting the server? Otherwise, how can I start it? I don't find anything on the docs

arbulu89 (dismissed):
More issues.
It looks like the private key flag is not really working:
```
vmmonitoring:/home/cloudadmin # sudo ./install-server.sh --private-key /root/.ssh/id_rsa
Please provide the path of the runner private key:
```

When the installation is finished I have an empty url:
```
Trento Server installed!

castle image...

URL: http://
```

install-server.sh
```diff
@@ -0,0 +1,122 @@
...
+    if ! which unzip >/dev/null 2>&1; then
+        echo "unzip is required by this script, please install it and try again."
+        exit 1
+    fi
+}
+
+install_k3s() {
+    echo "Installing K3s..."
+    curl -sfL https://get.k3s.io | sh >/dev/null
+    sudo cp /etc/rancher/k3s/k3s.yaml ~/.kube/config
```
aleksei-burlakov, line 78: The ~/.kube/config might not exist

install-server.sh
```diff
@@ -0,0 +1,122 @@
...
+            ;;
+            *)
+                usage
+                exit 0
+            ;;
+        esac
+    done
+
+    if [[ -z "$PRIVATE_KEY" ]]; then
+        read -rp "Please provide the path of the runner private key: " PRIVATE_KEY </dev/tty
```
aleksei-burlakov, line 58: Maybe we should add a default option (~/.ssh/id_rsa)

install-server.sh
```diff
@@ -0,0 +1,122 @@
...
+EOF
+}
+
+cmdline() {
+    local arg=
+    for arg
+    do
+        local delim=""
+        case "$arg" in
+            --private-key)  args="${args}-p ";;
```
aleksei-burlakov, line 30: It didn't work for me. It has prompted "Please provide the path of the runner private key:" even when the key was provided.

install-server.sh
rtorrero: I'd advise against such a default otherwise a user might be annoyed that we are copying around his private key without confirmation

install-server.sh
```diff
@@ -0,0 +1,122 @@
...
+            --help)         args="${args}-h ";;
+            
+            *) [[ "${arg:0:1}" == "-" ]] || delim="\""
+            args="${args}${delim}${arg}${delim} ";;
```
aleksei-burlakov, line 34: That's a good place to put a comment

install-server.sh
aleksei-burlakov: I mean a default option that still requires the users confirmation 
`Please provide the path of the runner private key [~/.ssh/id_rsa]:`

README.md
rtorrero: The user will be prompted when not providing the arguments through the command line. IMHO that's fine, right?

install-server.sh
rtorrero: Putting myself from the user's perspective, I would definitely not like that.

What maybe would make sense for a future improvement would be to automatically create a new SSH key in case the user chose not to give us one, wdyt?

rtorrero:
I addressed most of the points here, the ones I didn't are left open for discussion.

rtorrero:
May the force-push be with me (small typo fixed)

README.md
arbulu89: Yes, it works. I most probably did something wrong the last time as it was not promptying anything... XD

install-server.sh
arbulu89: Actually, I don't dislike Aleksei's suggestion. It is pretty common to have default values in the user intervention actions, to allow a quick click and go. 
@rtorrero I think you miss understood Aleksei's suggestion. He doesn't ask for a default value to not ask to the user, he just says that we might have a default value for the cases where you are asked, but you don't type anything, and simply click enter.

aleksei-burlakov (approved):
LGTM!

aleksei-burlakov:
I've noticed, after executing the script it lefts artifacts in the `/tmp/trento-main`

install-server.sh
stefanotorresi: hmm, this is a tricky discussion but also a gigantic bike shed. Let's follow-up later.

rtorrero:
> I've noticed, after executing the script it lefts artifacts in the `/tmp/trento-main`

Should be fixed now. Thanks!

stefanotorresi (approved):
LGTM

PR #255, opened 2021-09-20
Improve the metadata and check finding and move variables to defaults

arbulu89:
Many improvements in the ansible code:

- Move the actual check roles to the `checks` folders. This makes easier to find the real checks in the code
- Move the metadata information to the `defaults` folder. This allows us to use these variables in the checks execution as well, which is really interesting if we want to add some new options, like `on_failure: warning`, etc
- Automatize the check and metadata files finding. Now we don't need to declare all the checks, we just need to put them in the checks folder.

PD: Most of the changes are just movement of metadata variables


runner/ansible/roles/post-results/tasks/main.yml
```diff
@@ -1,14 +1,14 @@
 ---
 
-- name: set_global_results {{ test_id }}
+- name: set_global_results {{ id }}
   set_fact:
     results: |
       {{ (results | default({})) | combine(
         {
           'results': {
             item: {
```
brett060102, line 9: with the rewrite I am seeing:
TASK [post-results : set_global_results 1.4.2] *******************************************************************
[WARNING]: The loop variable 'item' is already in use. You should set the `loop_var` value in the `loop_control`
option for the task to something else to avoid variable collisions and unexpected behavior.
[WARNING]: The loop variable 'item' is already in use. You should set the `loop_var` value in the `loop_control`
option for the task to something else to avoid variable collisions and unexpected behavior.

since item is also being use as a loop var in check.yml and meta.yml this is causing the warning.

runner/ansible/check.yml
```diff
@@ -35,255 +35,19 @@
...
+        file_type: directory
+      register: checks
+      run_once: yes
+      delegate_to: localhost
+
+    - name: Run check
+      include_role:
+        name: "{{ item.path }}"
+      loop: "{{ checks.files }}"
+      when: "{{ (item.path | basename) in cluster_selected_checks }}"
```
brett060102, line 50: Thus use of item here, is conflicting with the use in post results. Also you don't need the "{{ }}" in the when condition Please consider changing to:
    - name: Run check
      include_role:
        name: "{{ CurTest.path }}"
      loop: "{{ checks.files }}"
      loop_control:
        loop_var: CurTest
      when: (CurTest.path | basename) in cluster_selected_checks

brett060102 (changes requested):

runner/ansible/roles/post-results/tasks/main.yml
arbulu89: Thanks! I will have a look

runner/ansible/check.yml
arbulu89: Yep, already noticed this myself!
Fix is on the way

arbulu89: Fixed on: https://github.com/trento-project/trento/pull/255/commits/e54decdc127415b9e2a9c65542c2091395dbd1f6

runner/ansible/roles/post-results/tasks/main.yml
arbulu89: Fixed on: https://github.com/trento-project/trento/pull/255/commits/e54decdc127415b9e2a9c65542c2091395dbd1f6

arbulu89:
@brett060102 Issues fixed. Could you review back again? Thanks

brett060102 (approved):
These changes look good. A couple of things to note. The test execution order is not longer sequential in my testing. This is not an issue, but its a change and it looks like this code take about 10% longer to run. Again not a real issue, but something to be aware of.

PR #256, opened 2021-09-21
provision.sh: include required trento-runner service file

rtorrero:
This PR installs the trento-runner service file when executing the provision.sh script to avoid a failure on a pipeline run.

### DIFF
--- a/hack/provision.sh
+++ b/hack/provision.sh
@@ -25,6 +25,7 @@ ACTIONS_RUNNER_NAME=$4
 ACTIONS_RUNNER_PATH="$ACTIONS_RUNNER_HOME/actions-runner"
 ACTIONS_RUNNER_REPO_URL="https://github.com/$ACTIONS_RUNNER_REPO_OWNER/$ACTIONS_RUNNER_REPO_NAME"
 ACTIONS_RUNNER_SYSTEMD_UNIT="actions.runner.$ACTIONS_RUNNER_REPO_OWNER-$ACTIONS_RUNNER_REPO_NAME.$ACTIONS_RUNNER_NAME.service"
+TRENTO_RUNNER_SYSTEMD_UNIT="trento-runner.service"
 
 CONSUL_VERSION=1.9.6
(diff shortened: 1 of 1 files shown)


hack/provision.sh
```diff
@@ -101,6 +102,16 @@ install_actions_runner() {
...
   systemctl enable --now "$ACTIONS_RUNNER_SYSTEMD_UNIT"
 
+  if [ -f "/etc/systemd/system/$TRENTO_RUNNER_SYSTEMD_UNIT" ]; then
+    echo "  Warning: trento-runner systemd unit already installed. Removing..."
+    systemctl stop "$TRENTO_RUNNER_SYSTEMD_UNIT"
+    rm "/etc/systemd/system/$TRENTO_RUNNER_SYSTEMD_UNIT"
+  fi
+
+  echo "  Installing systemd unit"
+  ./svc.sh install >/dev/null
```
dottorblaster, line 112: Could we separate the bracket with a space after, like `./svc.sh install > /dev/null`?

dottorblaster (dismissed):
LGTM but have a look at the comment

arbulu89 (dismissed):
@rtorrero We didn't actually work in the `provision.sh` script, as we thought that It would be removed, and we would have some new strategy for the CD.
I did all the installation manually for the 0.4.0.

I don't really know if it makes sense to continue with this approach, as it is not scalable at all and requires fair amount of maintenance. Anyway, I think that this discussion already happened within the team.

rtorrero:
@arbulu89 totally agreed, but as we had some new faces joining us I wanted to have this fixed until we are able to remove it

rtorrero:
> Could we separate the bracket with a space after, like `./svc.sh install > /dev/null`? 

sure thing

arbulu89 (approved):

rtorrero:
Closing this as it doesn't make sense anymore now that the hacks folder is gone. I'm actually happy about this :-)

PR #257, opened 2021-09-21
Create the Trento ansible callback code

arbulu89:
Create the new Trento ansible callback code. This code tries to simplify and improve the actions that now are being done in Ansible, to python code (unfortunately, the callback code must be in this language and we cannto use go).

- It helps to simplify the ansible code itself, as it doesn't need to take care of storing and uploading the results. This way, ansible code only focus on having the checks
- It provides much more options to deal with the results in a comfortable way (store skipped checks for example, something we cannot do now, without super complex ansible code)
- Speeds up a bit the execution, as it removes many "operational" tasks
- It opens the gate to replace ARA by something better in the future (some kind of time series database)

PD: It has zero impact in the app in general. The results have the exact same format. We will work on that in future changes (add skipped checks, and warnings, add execution date, etc)


runner/ansible/callback_plugins/trento.py
```diff
@@ -0,0 +1,223 @@
...
+        hosts = checks[test]["hosts"]
+        if host not in hosts:
+            hosts[host] = {}
+
+        hosts[host]["result"] = result
+
+
+class CallbackModule(CallbackBase):
+    """
+    Tremtp Callback module
```
dottorblaster, line 81: Is `Tremtp` a typo? :joy:

runner/ansible/callback_plugins/trento.py
```diff
@@ -0,0 +1,223 @@
...
+        """
+        changed = False
+        record = self.client.get("/api/v1/records?playbook=%s&key=%s" % (playbook, key))
+        if record["count"] == 0:
+            # Create the record if it doesn't exist
+            record = self.client.post(
+                "/api/v1/records", playbook=playbook, key=key, value=value, type=record_type)
+            changed = True
+        else:
+            # Otherwise update it if the data is different (idempotency)
```
dottorblaster, line 215: What about moving these two comments into the function documentation?

dottorblaster (commented):
Sorry, started looking at this briefly then I noticed it was still a draft

arbulu89:
@dottorblaster Ready to be reviewed now ;)

dottorblaster (approved):
I'm not that strong with Python but LGTM :D

PR #259, opened 2021-09-21
Frontend tooling: introduce Prettier and ESLint

dottorblaster:
This PR introduces Prettier and ESLint inside our development environment. This is propedeutical towards the adoption of a new frontend library to architect things in a more modern way.

You can actually see the steps I took looking at the commits, but I'll summarize that here too:

- I installed the new packages using NPM;
- Added some sensible default configuration just to ensure everything is on track from day one. I also added the eslint React plugin since we're moving towards React;
- Formatted all the frontend source code using Prettier;
- Corrected all the eslint errors either adjusting the configuration or disabling noisy rules for legacy code;
- Added a make step in the CI so we're covered from now on.

Hope you like it!


Makefile
```diff
@@ -70,6 +70,15 @@ web-deps: web/frontend/node_modules
 web/frontend/node_modules:
 	cd web/frontend; npm install
 
+web-format: web/frontend/format
```
nelsonkopliku, line 73: `web-format` does not seem to be used anywhere.

Is it supposed to be used locally to actually apply changes and not only check as it happens on CI/CD?

nelsonkopliku (dismissed):
LGTM.

Didn't go too deep in style setting now. I think that's a good starting point.
I also think that, as a team, we will find then our style and update accordingly when and if needed.

Question: No typescript if I understand correctly, right?

Makefile
dottorblaster: > I also think that, as a team, we will find then our style and update accordingly when and if needed.

That was actually my mindset, I just scaffolded a minimal configuration :-)

> Question: No typescript if I understand correctly, right?

Not until somebody asks for it :joy: let's keep it simple for now :grimacing:

dottorblaster: Also, sorry for not answering your main question: `web-format` is meant to be used only locally, what do you think about it?

nelsonkopliku: Don't be sorry :slightly_smiling_face: 
That's super, it's a nice goodie. I would just keep it.

One thing I am thinking right now is: probably we could also hook in github pre-commit and attach style checking (not fixing), so you wouldn't even be able to commit and push unstyled code.
But that's just an idea, we can introduce this if needed.

arbulu89 (approved):
From a double quote lover, this makes me sad, but it is what is is (just joking hehe).
Good job!

dottorblaster:
@arbulu89 well this can be changed in the future, the cool thing about prettier is that we have one config to rule them all :D

PR #260, opened 2021-09-22
Discover SUSE subscription details

arbulu89:
**Ready to be reviewed. In draft because I didn't implement the UT and becaues it will conflict with https://github.com/trento-project/trento/pull/240**

Implement the SUSE subscription details discovery mechanism.
It simpley runs `SUSEConnect -s` and stores the output in the KV storage under the hosts entry.

I have omitted the storage of the registration code, as it look private (easy to change in any case).

To show the data, I have implemented a basic table in the hosts view. We can definitely improve this or use the data in any other place:
image


arbulu89:
I have added a new `subscriptions` service in this last commit; https://github.com/trento-project/trento/pull/260/commits/ceaafefae4de46a741a31fd5dd0b7f4f756e64e8

I think it makes sense to have a more service oriented code, which we can use in the web handlers. This way, if we decide to change consul at some point, it should be easier.

What do you think guys?

fabriziosestito:
> I have added a new `subscriptions` service in this last commit; [ceaafef](https://github.com/trento-project/trento/commit/ceaafefae4de46a741a31fd5dd0b7f4f756e64e8)
> 
> I think it makes sense to have a more service oriented code, which we can use in the web handlers. This way, if we decide to change consul at some point, it should be easier.
> 
> What do you think guys?

LGTM! not sure we will need a dedicated service for host subscriptions when we will drop the KV store in favor of the relational model since it could be part of the host repo, but looks cleaner as for now.

fabriziosestito (dismissed):

arbulu89:
@fabriziosestito This PR will conflict with https://github.com/trento-project/trento/pull/240 , so let's see if we can close that one first to adapt the current code to the unique consul health check system

stefanotorresi:
@arbulu89 can you undraft this and add tests?
I think it will be easier to rebase #240 against this one, instead of the other way around.
Also this is kinda more important and urgent, since it introduces net-new value.

arbulu89:
@stefanotorresi @fabriziosestito 
UT added and PR moved to `Ready to review`. I have not done any change with this last commit (besides the UT), so you can just have a superficial check.

fabriziosestito (approved):

PR #261, opened 2021-09-22
Add a default ansible configuration file usage

arbulu89:
In order to optimize the ansible execution, we can tune some parameters in the config file. Some info:
https://docs.ansible.com/ansible/2.3/intro_configuration.html

And a guide I followed (there are many, but all of them reference to these 3 elements):
https://gryzli.info/2019/02/21/tuning-ansible-for-maximum-performance/

In order to use a configuration file, the env variable `ANSIBLE_CONFIG` is used.

The applied changes are:
- `forks = 100` (default 5). How many ssh connections in parallel are allowed. I put a quite big number. We might have bigger deployments, but higher values might affect the ansible performance for the bad
- `ControlPersist=300s` (default 60s). The ssh connection "keep-alive" kind of option. 300s should be always higher than the execution time, so we should be fine.
- `pipelinening=true` (default false). Reduces the number of SSH operations required to execute a module on the remote server

@brett060102 Could you have a look too?

### DIFF
--- a/runner/ansible/ansible.cfg
+++ b/runner/ansible/ansible.cfg
@@ -1,0 +1,7 @@
+[defaults]
+forks = 100
+
+[ssh_connection]
+ssh_args = -o ControlMaster=auto -o ControlPersist=300s -o PreferredAuthentications=publickey
+control_path = %(directory)s/ansible-ssh-%%h-%%p-%%r
+pipelining = True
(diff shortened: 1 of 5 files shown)


brett060102 (commented):
The config file looks good and I approve that part of this review. I think the go code looks OK, but I am not a qualified reviewer for that.

PR #264, opened 2021-09-23
Minor Helm chart updates

stefanotorresi:
- remove the trento-runner service, because it has nothing to expose, it's actually just a background worker for now.
- rename ara port to `ara-http` because just `http` was ambiguous with the default port 80 (ara runs on 8000 by default instead).


stefanotorresi:
self-merged because I need this NAO, feel free to provide feedback post-facto!

Issue #262, opened 2021-09-22
Trento RPM built in devel:sap:trento is not being injected the version number correctly


stefanotorresi:
The binary built in this package: https://build.opensuse.org/package/show/devel:sap:trento/trento

Has the following behaviour:
```
> trento version
Trento version 0
built with go1.16.5 linux/amd64
```

The expected version number should instead match the RPM version, which currently is `0.4.0+1630926451.42dff2b`

aleksei-burlakov:
The fix is proposed in #265

PR #265, opened 2021-09-23
Fix: set the version in the Makefile explicitly

aleksei-burlakov:
The Makefile takes the version from the get_version_from_git.sh
but it might not work on the obs, so let's set it explicitly.

(It's not tested, the obs takes forever to publish).

#262


stefanotorresi (approved):
perfect! 🚀

PR #266, opened 2021-09-23
Add warning and skipped states to the checks

arbulu89:
Implement the warning and skipped check results.

- The skipped check result shows up when the check is not selected
- The warning comes from the metadata entry `on_failure: warning`, and it must be defined in the test itself. This follows the same implementation that we had in benchcommon. I have cherry picked the tests with the warning output from there.

Here a pic (the warning icon is obvious, the gray circle is the skipped option):
image


dottorblaster (approved):
LGTM but keep in mind that maybe this part could be reworked very very soon :-D

PR #267, opened 2021-09-23
Fix support for -e cluster_selected_checks= option

brett060102:
When -e is used to set cluster_selected_check on the command line
it effectively prohibits resetting/updating cluster_selected_check
in the yml code. So the conversion of cluster_selected_check from
a text string to a list in load_facts/tasks/main.yml was only
effective for the task scope. After that cluster_selected_check was back
to a string again.

The fix is to not do the list conversion in place, but to place the result
in a new variable and use that in the when condition in main.yml.


arbulu89 (approved):
Thanks @brett060102

PR #268, opened 2021-09-23
Add a zypper setup step to the install-server script

dottorblaster:
As above :-)


dottorblaster:
I'm an idiot, this script is not supposed to know about zypper. I'll rework this.

arbulu89 (commented):
I wouldn't go this way, as this attaches the installer only to zypper based OS, which is only SUSE.
I would add the dependency in the `check_deps` entry if possible (like curl and unzip).
What do you think?
PD: Now I see, the `check_deps` is not even used...
But We had discussions that the server side should work on any distribution, and I guess the installer should follow the same approach

PD2: It looks like you realized before I posted the message hehe

stefanotorresi:
wtf is `apparmor_parser` btw? 😁

dottorblaster:
@stefanotorresi it's a tiny thing that loads apparmor profiles inside the kernel, actually k3s' installer should take care of setting it up but it doesn't, so I'd go a bit defensive on our setup steps :-)

dottorblaster:
@arbulu89 you're right, I changed the script but still I'd like to detect if the system is apparmor-enabled too

install-server.sh
```diff
@@ -71,6 +71,10 @@ check_deps() {
         echo "unzip is required by this script, please install it and try again."
         exit 1
     fi
+    if (( ! which aa-status >/dev/null 2>&1 && ! which apparmor_parser >/dev/null 2>&1 )); then
```
arbulu89, line 74: I have tried this, and (apart of having a syntax error that is fixed removing the parenthesis from the comparison) in my case even though having `apparmor-parser` installed (by zypper in my case), these 2 binaries are not found by `which`.

arbulu89: Update: apparmor-parser is there but in `/sbin/apparmor_parser` which was not set in my PATH, and I don't know if this is normal or not... but `aa-status` is not anywhere. But k3s works without this last utility

arbulu89: `aa-status` comes in the `apparmor-utils` package

dottorblaster: Oh hell ok, then I have to find a cooler way to see if the system is apparmor-enabled :(

dottorblaster:
@arbulu89 are you on Leap or Tumbleweed?

arbulu89:
> @arbulu89 are you on Leap or Tumbleweed?

`Leap 15.2`

dottorblaster:
@arbulu89 you're right, I run that with root so I didn't even notice that /sbin and /usr/sbin are in the PATH only for `root` (and it's on purpose!)

Do you think a `sudo which` would work on this? Do you feel like any other solution would serve the purpose?

fabriziosestito:
> @arbulu89 you're right, I run that with root so I didn't even notice that /sbin and /usr/sbin are in the PATH only for `root` (and it's on purpose!)
> 
> Do you think a `sudo which` would work on this? Do you feel like any other solution would serve the purpose?

you could `stat` the file

rtorrero:
> @arbulu89 you're right, I run that with root so I didn't even notice that /sbin and /usr/sbin are in the PATH only for `root` (and it's on purpose!)
> 
> Do you think a `sudo which` would work on this? Do you feel like any other solution would serve the purpose?

Or also just add `/sbin` to $PATH before using which?

install-server.sh
```diff
@@ -71,6 +71,12 @@ check_deps() {
         echo "unzip is required by this script, please install it and try again."
         exit 1
     fi
+    if grep -q "Y" /sys/module/apparmor/parameters/enabled; then
+        if ! which apparmor_parser >/dev/null 1>&1; then
```
stefanotorresi, line 75: nit: consider using the `command` built-in instead of `which` (ref: https://stackoverflow.com/questions/37056192/which-vs-command-v-in-bash).

stefanotorresi (changes requested):

PR #269, opened 2021-09-24
CD: Use only one GitHub runner instead of 1 per node using new install scripts

rtorrero:
This PR done together with @aleksei-burlakov @nelsonkopliku makes use of the new installation scripts to deploy `trento` to our azure demo environment. It simplifies ci-cd.yml by requiring only one GitHub Action runner registered instead of requiring one GH Action runner per node.

It requires some addition variables to be set in a GH environment named `AZURE_DEMO`

`SSH_CONFIG` -> Configuration file for the SSH client used to reach the agents

`SSH_KEY` -> pre-authorized SSH key in the target hosts

`TARGET_USER` -> the user under which the cluster was deployed

`TRENTO_SERVER_IP` -> the IP of the machine where we should run install-server.sh on

`TRENTO_TARGET_AGENTS` -> a comma-separated list of hosts where to run install-agent.sh on (cluster nodes)

Note: The IPs provided above should be the internal IPs
Feedback is appreciated :-)


aleksei-burlakov (dismissed):
That's a masterpiece!

stefanotorresi (dismissed):
This is great, guys!

I would only like more consistent naming of the environment variables; consider the following suggestion:

`TARGET_USER` -> `TRENTO_USER`
`TRENTO_TARGET_AGENTS` -> `TRENTO_AGENT_HOSTS`
`TRENTO_SERVER_IP` -> `TRENTO_SERVER_HOST`

We also need to document these variables somewhere, either via some docblocks in the worfklow file, or maybe even a dedicated markdown file in the `/docs` dir.

Also, did you test what happens with subsequent deloyments? Xabi experienced some issues with the k3s installer when k3s already exists in the system, so maybe we need to tweak this in our server installation script.

stefanotorresi:
One more thing, and this is going to be annoying, I know, but we really ought to do this already: we need to use a bastion host by default. We've been a little too reckless in the security regard, so far.
I guess that would require a new environment variable, e.g. `BASTION_HOST`, and then adding the `-j` flag to all the ssh commands.

stefanotorresi:
Oh, and don't forget to remove outdated scripties from the `hacks` dir! 😁

install-server.sh
```diff
@@ -84,6 +84,8 @@ install_k3s() {
 
 install_helm() {
     echo "Installing Helm..."
+    # FIXME: why does /usr/local/bin vanish from PATH?
```
dottorblaster, line 87: This actually happened to me too, I really don't know and I think it's worth investigating in a second moment

dottorblaster (dismissed):
Same as @stefanotorresi, sorry to interfere but I think we need a setup with the bastion :/

Anyway, astonishing job, I really like it from a purely CD-related standpoint!

nelsonkopliku (commented):
While running this on my cluster I get an error during consul startup.
This happens in an agent node.

```
Sep 24 16:09:21 vmhana01 consul[563]: ==> failed to setup node ID: mkdir /srv/consul/data: permission denied
```
Guess: the user that's running the github action (cloudadmin) is not able to write into `/srv/consul`

**Edi**t: after having provisioned my cluster nodes with the old `provision.sh`, a conflict occurred in `trento-agent` and `consul` system units.

By cleaning up the node:
```
zypper rm trento
rm -fr /srv/consul/
rm -fr /srv/trento/
# cleanup systemd unit files
```
the problem does not occur anymore

fabriziosestito (changes requested):
Great work guys!

However, looking at the jobs it doesn't seem they install the just built version of trento.

The install server and install agent scripts are pinned to the latest release (currently 0.4.0), not to the `rolling` binary/image, we need to modify those to add a rolling option or figure out something for this ones.

fabriziosestito:
> One more thing, and this is going to be annoying, I know, but we really ought to do this already: we need to use a bastion host by default. We've been a little too reckless in the security regard, so far.
> I guess that would require a new environment variable, e.g. `BASTION_HOST`, and then adding the `-j` flag to all the ssh commands.

@stefanotorresi  @dottorblaster The jobs are linked to the `self hosted runner`, which works in a pull fashion inside the private network already. So no connection from the outside it is required.

This was to keep the CI agnostic from Azure, and to leave the freedom to whoever forks the project to run the same CI in their private env.
However if we agree that we are coupling with Azure for the CI/CD, we could use bastion to remove the self-hosted runner.

Keep in mind we might want to use the self hosted runner in a future anyway since GitHub credit might be not enough at certain point.

Thoughts?

arbulu89:
> > One more thing, and this is going to be annoying, I know, but we really ought to do this already: we need to use a bastion host by default. We've been a little too reckless in the security regard, so far.
> > I guess that would require a new environment variable, e.g. `BASTION_HOST`, and then adding the `-j` flag to all the ssh commands.
> 
> @stefanotorresi @dottorblaster The jobs are linked to the `self hosted runner`, which works in a pull fashion inside the private network already. So no connection from the outside it is required.
> 
> This was to keep the CI agnostic from Azure, and to leave the freedom to whoever forks the project to run the same CI in their private env.
> However if we agree that we are coupling with Azure for the CI/CD, we could use bastion to remove the self-hosted runner.
> 
> Keep in mind we might want to use the self hosted runner in a future anyway since GitHub credit might be not enough at certain point.
> 
> Thoughts?

I think we are mixing things here. I don't see that @stefanotorresi recommends to use the bastion as the github runner. We can still continue using (and should) and totally external runner, that it is out of the deployment VMs group.

What I understand from Stefano is that our deployment should be using the bastion, which eventually changes how we tell the github CI process to find the machines, so we should only provide the Bastion public address and after the private addresses of all the other machines, so we don't expose so openly the public IP addresses.

If for some reason we want to provide the 2 options (connect through bastion or directly to the nodes) it would be even better.

stefanotorresi:
> What I understand from Stefano is that our deployment should be using the bastion, which eventually changes how we tell the github CI process to find the machines, so we should only provide the Bastion public address and after the private addresses of all the other machines, so we don't expose so openly the public IP addresses.

Yes, exactly, thanks for clarifying that Xabi.

We can and should use a dedicated github action runner, but that should live in a separate host than the bastion host, or any other host of the target infrastructure. That is: we should completely separate the demo deployment from the CI. The CI just connects to it.
If you want to simplify things, feel free to use a GitHub provided runner one for the time being.

rtorrero:
> > What I understand from Stefano is that our deployment should be using the bastion, which eventually changes how we tell the github CI process to find the machines, so we should only provide the Bastion public address and after the private addresses of all the other machines, so we don't expose so openly the public IP addresses.
> 
> Yes, exactly, thanks for clarifying that Xabi.
> 
> We can and should use a dedicated github action runner, but that should live in a separate host than the bastion host, or any other host of the target infrastructure. That is: we should completely separate the demo deployment from the CI. The CI just connects to it.
> If you want to simplify things, feel free to use a GitHub provided runner one for the time being.

so to clarify, we don't need to address anything in that regarding right?

The selection of which host gets used is done through registering the GH actions runner from the desired machine (as the CI code doesn't have any requirements besides from being in the same private network as the `TRENTO_AGENT_HOSTS`). 

And the `ssh -J` for the bastion is not really needed here, as registering a runner against GH makes the bastion unnecessary (CI code gets executed on a host which can already reach the targets without the need of going through the bastion)

rtorrero:
Requested changes should be there, except for the bastion. In other words, the requirement of registering the action runner on a machine that is within the same network of the nodes is still there. I'd address this in a separate PR.

Kudos to @aleksei-burlakov and @nelsonkopliku !

install-agent.sh
```diff
@@ -181,6 +201,19 @@ function install_trento() {
...
+    ARCH=$(uname -m | sed "s~x86_64~amd64~" | sed "s~aarch64~arm64~" )
+    if [[ -n "$USE_ROLLING" ]] ; then        
+        TRENTO_TGZ_URL=https://github.com/trento-project/trento/releases/download/rolling/trento-${ARCH}.gz        
+    else
+        TRENTO_TGZ_URL=https://github.com/trento-project/trento/releases/download/${TRENTO_VERSION}/trento-${ARCH}.gz
+    fi
+  
+    curl -f -sS -O -L "${TRENTO_TGZ_URL}" >/dev/null
+    tar -zxf trento-${ARCH}.gz /usr/bin/
+    rm trento-${ARCH}.gz
```
aleksei-burlakov, line 214: We need to also copy the service file.

rtorrero:
We still found a few minor fixes required, doing them now

fabriziosestito (approved):

aleksei-burlakov (approved):
Let's merge we will fix it up later.

PR #270, opened 2021-09-27
Add a --rolling option to the install-agent script to use factory repos

dottorblaster:
In the past week the team spotted the need for the installers to set up the latest RPM package coming from the Factory repository, so we're adding this to the `install-agent.sh` script. We'll do that in a further PR for the `install-server.sh` script too.

### DIFF
--- a/install-agent.sh
+++ b/install-agent.sh
@@ -37,6 +37,9 @@ ARGUMENT_LIST=(
     "server-ip"
 )
 
+TRENTO_REPO=${TRENTO_REPO:-"https://download.opensuse.org/repositories/devel:/sap:/trento/15.3/devel:sap:trento.repo"}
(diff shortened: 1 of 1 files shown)


rtorrero (dismissed):
Looks perfect to me :-)

arbulu89 (approved):
Looks good!

PD: For future (or current improvements)
As @stefanotorresi is trying to understand, could we even confirm if this KEY thing is needed, or we could get rid off it?

arbulu89 (changes requested):
@dottorblaster We would need to add the new flag usage in the `-h` output

dottorblaster:
@arbulu89 I updated the help text, thank you for reminding me :sweat_smile:

arbulu89 (approved):
Thanks!

PR #272, opened 2021-09-27
Run ansilble-lint in the CI process

arbulu89:
Run `ansible-lint` during the CI `test` stage.
This will help to have a better code quality in the ansible code.
The PR includes the warmed issues (I have disabled some checks as they are quite picky and not useful in our case)

PD:
We cannot use the github CI plugin because it doesn't have the option to set ENV variables, so I had to run a plain run stage
https://github.com/marketplace/actions/ansible-lint

### DIFF
--- a/.github/workflows/ci-cd.yaml
+++ b/.github/workflows/ci-cd.yaml
@@ -52,6 +52,21 @@ jobs:
         run: make fmt-check
       - name: frontend lint and coding style check
         run: make web-check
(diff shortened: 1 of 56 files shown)


dottorblaster (approved):
I am and will always be in favor of any linter that helps us concentrate on other aspects :-D

fabriziosestito (approved):

PR #273, opened 2021-09-28
Add About page with subscription details

arbulu89:
New `About` page. Some context:

The initial idea of all of this code was to have a place to show the user if they have a Premium* or Free Trento deployments.
In order to show this, we agreed that the initial best place would be an About page.
This information usually goes in the User profile page, but this requires much more effort, so having it at the beginning in the About page looks reasonable.

With this I have taken the advantage to add some other small items to have a more populated page.

Special input needed from @lee-martin , to discuss about wording and any other aspect

*Premium means that at least one of the connected agents is subscribed to the SLES_SAP SCC channel.

Here some pictures:
image

image


stefanotorresi (approved):
Awesome! 🥇

dottorblaster (approved):
LGTM

web/services/subscriptions.go
```diff
@@ -1,14 +1,28 @@
...
+
+	consulApi "github.com/hashicorp/consul/api"
 	"github.com/trento-project/trento/internal/consul"
 	"github.com/trento-project/trento/internal/subscription"
 )
 
+const (
+	SlesIdentifier string = "SLES_SAP"
+	Premium        string = "Premium"
+	Free           string = "Free"
```
rtorrero, line 14: nitpicking here but... maybe use a different word; maybe community for the non-premium version?

rtorrero (approved):
LGTM! Made a small comment but since it's more of a flavor thing, approving anyways

web/services/subscriptions.go
dottorblaster: Woh I think you're right, good catch

stefanotorresi: We don't have definitive instructions WRT to the branding here, but there is no particular indicator for the "free" version to be implemented: it's either "premium" or it isn't.

stefanotorresi: We can throw some ideas around and then validate with PO, but you can keep it as it is for now.

arbulu89: I had my internal backs and forths already with this hehe
Initially I put `Standard`, but what the hell does this mean? hehe
Later on I tried with a `-` character, but It looked so empty

Maybe I could put `Basic`...

But as this is just a wording thing, we can discuss with Lee, Chrisitna, Diego, etc, and It would be an easy thing to change.
This will go to our internal Review state, so I hope it won't go away without this new discussion

PR #276, opened 2021-09-29
Update subscription Load test to use ElementsMatch

arbulu89:
The `TestLoad` in `subscription_kvstore_test.go` if failing in other CI executions.
I have not been able to reproduce it locally, but I hope this change fixes it in any case, as the error might be caused by the order of the elements

PD: Confirmed that it was an issue in the order to the list. So this fix will have good effect 100%


rtorrero (approved):
LGTM!

PR #279, opened 2021-09-29
CI: fix wrong output path for the binaries

rtorrero:
This PR fixes a bug in PR#269 as we forgot to update the paths for the new tgz output files. This should fix it.


aleksei-burlakov (approved):
It's almost there.

PR #280, opened 2021-09-29
install-agent: fix typo in the script

rtorrero:
The diff is pretty self-explanatory. Now I fade into darkness out of pure shame :-)


aleksei-burlakov (approved):
I was sure we fixed {$} before. It should work now.

PR #282, opened 2021-09-30
Runner - Create new check ids system

arbulu89:
**Don't merge, UT are missing**

Create a new Trento checks identifier system to have always unique-lifetime check identifier.

In order to facilitate the check identification, without any direct impact in the technical details, this PR introduces a new `external_id` fields to checks metadata. This ID is always unique within the checks catalog, and must not be changed during the lifetime of the check.

More details:
- The old identifier is kept to provide an easier way to continue working with the checks. If the external id would've replaced the id in the code structure, it would be super complex to find and change the checks. It would be source of errors. So, to make the development easier (we can mentally identify better the code with numerical order), we will still continue having the current numbering in the code. In any, case this numbers don't affect at all in the user interface. They won't even be displayed.
- The old identifier is still used to provide the ordering of the checks in the user-interface
- We can freely change the internal id at any point, if we find any better solution
- The checks metadata type is moved from dictionary to list. Dictionary are pretty bad option to manage order in golang. Using lists makes life easier, and I don't see any disadvantage.

Besides this, there is not mayor impact in the user side, except we show the `external_id`.

### DIFF
--- a/runner/ansible/README.md
+++ b/runner/ansible/README.md
@@ -8,17 +8,35 @@ To be defined...
 
 ## Metadata files
 
-The metadata files provide information about the test themselves. They are used to get information
+The metadata files provide information about the check themselves. They are used to get information
 from the Trento Web GUI and render properly everything related with the Ansible tasks.
 
 In order to use them properly, some fields are required. An example is available at [defaults/main.yml](roles/checks/1.1.1/defaults/main.yml).
 
 These are the fields needed by Trento:
 
-- `id`: The test unique identifier.
-- `name`: A short name for the check. It is used as more user friendly identifier for the test
+- `external_id`: The check unique identifier. This value must not be changed during the lifetime of the check
+- `id`: The internal identifier. It is used to identify the checks internally (a more user-friendly nomenclature), to specify the expected values and to apply the ordering to the lists in the web side. This value can be changed to adapt to any new internal need.
+- `name`: A short name for the check. It is used as more user friendly identifier for the check
 - `group`: The group which the check belongs to. It is used to group the checks under different visual elements in Trento
 - `labels`: A list of labels (separated by command) which helps to group the checks by execution groups. The difference between this and the `group` field
 is that the labels are used for control purpose (select all the checks with this label e.g.), and the groups are used just for visual purposes
-- `description`: A longer description about the test purpose. It can be written using markdown.
+- `description`: A longer description about the check purpose. It can be written using markdown
 - `implementation`: Usually the task `main.yml` content
+
+## Creating a new external_id
+
+The `external_id` must be unique in the check collection. It must be 6 hexadecimal digits string.
+In order to create a new unique identifier, and to check if there are any duplicated entries, the
+`id_checker.py` script can be used.
+
+To use it:
+```
+# Check if the checks include all the required metadata values and if there is any duplicated external_id
+python3 support/id_checker.py
+```
+
+To add a new unique `external_id`:
+```
+python3 support/id_checker.py --generate
+```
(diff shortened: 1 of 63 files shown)


runner/ansible/README.md
```diff
@@ -8,17 +8,35 @@ To be defined...
...
+To use it:
+```
+# Check if the checks include all the required metadata values and if there is any duplicated external_id
+python3 support/id_checker.py
+```
+
+To add a new unique `external_id`:
+```
+python3 support/id_checker.py --generate
+```
```
stefanotorresi, line 42: I think this entire file should go into `/docs/runner.md`, so that we have all the docs in one place. :)

runner/ansible/README.md
```diff
@@ -8,17 +8,35 @@ To be defined...
...
 from the Trento Web GUI and render properly everything related with the Ansible tasks.
 
 In order to use them properly, some fields are required. An example is available at [defaults/main.yml](roles/checks/1.1.1/defaults/main.yml).
 
 These are the fields needed by Trento:
 
-- `id`: The test unique identifier.
-- `name`: A short name for the check. It is used as more user friendly identifier for the test
+- `external_id`: The check unique identifier. This value must not be changed during the lifetime of the check
+- `id`: The internal identifier. It is used to identify the checks internally (a more user-friendly nomenclature), to specify the expected values and to apply the ordering to the lists in the web side. This value can be changed to adapt to any new internal need.
```
stefanotorresi, line 19: I am very confused by this. Why are we keeping the old id? Having two ids seems very complicated to me. My expectation is that we would just use the new alfanumeric convention in the existing `id` field.

stefanotorresi (changes requested):

runner/ansible/README.md
stefanotorresi: Sorry, rushed to review without reading the opening post. 😅 

So, I really don't like having two id fields: it makes things very confusing.

If the purpose of what was previously called `id` is now just to order things and being a cosmetic name, then I would rename to something else. The identifier, as a concept, should be just one; here we instead are telling advanced users that they need to come up with two ways to identify the checks, which will be very confusing and cumbersome to them.

stefanotorresi: We already have a `name` field to provide a user-friendly identifier, so I guess the only thing you want here is an optional `order` field.

arbulu89: I knew that we would have this discussion hehe
Agree that having 2 ids is confusing (unless you read the documentation).

Let's do the next:
- `external_id` is renamed as `id`, so we have one id field
- The current `id` is renamed to `name`. We remove the current `name` content (we don't use it anywhere). The `name` is a developer friendly string, which is actually used to specify the expected values map, etc
- New `order` field to specify the check order **within** the test group (so we don't need to change the order for all the checks if we only want to change the check order withing the group)

arbulu89:
People, don't be afraid of the number of changed files, 95% of the files are about the ansible checks, which are reiterative change of some fields

dottorblaster (approved):
Absolutely LGTM

hack/id_checker.py
```diff
@@ -0,0 +1,151 @@
...
+    """
+    Get the checks folder absolute path
+    """
+    return os.path.join(os.path.dirname(os.path.abspath(__file__)), CHECKS_FOLDER)
+
+def create_id():
+    """
+    Create ID with specified lenght based on HEX digits
+    """
+    generated_id = "".join([random.choice(HEXDIGITS) for _ in range(ID_LENGTH)])
```
stefanotorresi, line 53: Maybe we could use `uuid.uuid4().hex[ID_LENGTH]` instead, and do without the `HEXDIGITS` constant :)

stefanotorresi (changes requested):

hack/id_checker.py
arbulu89: Fixed in: https://github.com/trento-project/trento/pull/282/commits/c796f58561bf795d5050ef3c80944da8d284a9e1

stefanotorresi (approved):

PR #284, opened 2021-09-30
CD: halt deploy until images / packages are ready

rtorrero:
This PR does a small fix in the ordering of the jobs so that the deployment to the demo environment is done only when the required packages have been successfully built.


aleksei-burlakov (approved):
Perfect!

PR #286, opened 2021-09-30
Minor improvements

aleksei-burlakov:
Polishing of the build.compress job

### DIFF
--- a/.github/workflows/ci-cd.yaml
+++ b/.github/workflows/ci-cd.yaml
@@ -149,7 +149,7 @@ jobs:
           if_key_exists: replace
           config: ${{ secrets.SSH_CONFIG }}
       - name: deploy trento services on K3S cluster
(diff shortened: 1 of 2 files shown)


stefanotorresi (approved):
LGTM!! 🚀

arbulu89:
Coming late and without any specific value, but I would encourage everyone to not use PR names like `Minor improvements` (the same goes for commit messages). This is terrible to track back what we changed. For instance, in the future release changelog. What will be put? hehe

All the PRs aim to solve a specific thing. This minor improvement might be renamed too `Fix CI build for xxx`

aleksei-burlakov:
> Coming late and without any specific value, but I would encourage everyone to not use PR names like `Minor improvements` (the same goes for commit messages). This is terrible to track back what we changed. For instance, in the future release changelog. What will be put? hehe
> 
> All the PRs aim to solve a specific thing. This minor improvement might be renamed too `Fix CI build for xxx`

I agree. It was one improvement at first, then there were more and it became that generic.

PR #288, opened 2021-09-30
Fix the liveness probe in the ARA chart due to sporadic sigterms

rtorrero:
This PR ~removes~ fixes the liveness probe for ARA that is causing sporadic failures in our azure demo environment.


fabriziosestito (approved):

rtorrero:
Turns out we don't need to disabled it, switching to `/healthcheck/` on the liveness probe made it consistently detect the correct health status.

stefanotorresi (approved):
perfect!

PR #290, opened 2021-09-30
Added how-to about adding checks

MMuschner:


runner/ansible/README.md
```diff
@@ -4,7 +4,12 @@ Place to store Ansible playbooks that are executed by the Trento runner
 
 ## Implementing a check
 
-To be defined...
+# Table of contents
+
+- [Structure]
+   - [Metadata files](#metadata-files)
```
arbulu89, line 10: Let's put other entries here:
- `Check structure`. In this sub chapter explain where the check folder should go `roles/checks/etc`. How we should select a name (explaining that this will determine the order which the test is executed), and that we need the `defaults` and `tasks` folder, with the `main.yml` in each of them

- `Check file` Here I want you to explain how to create the check code, with an actual example (put the code here in the docu). I want a description saying that check is task or a group of tasks that eventually generates a `true` or `false` output (`true -> check passed`, `false -> check failed`). And that this result must be passed to the `post-results` block, with the result in the `status` field. I want to be meticulous on this. Explain every bit, as you were writting a test for the 1st time
- `Metadata files`. Keep as it is. Maybe we could include a dummy metadata file content below the fields list, like in the check file

runner/ansible/README.md
```diff
@@ -22,3 +27,38 @@ These are the fields needed by Trento:
...
 - `description`: A longer description about the test purpose. It can be written using markdown.
 - `implementation`: Usually the task `main.yml` content
+
+## Check files
+
+The check files contain the actual task which the runner will execute. An
+example is available at [tasks/main.yml](roles/checks/1.1.1/tasks/main.yml).
+These files are written like a normal ansible task.
+
+## Workflow
```
arbulu89, line 37: Let's remove completely this `Workflow` page. I don't want to explain here how to setup the runner.
This document only aims to the Check implementation itself

arbulu89 (changes requested):
Hi @MMuschner,
Thanks for the 1st contribution!
We will need a bit more work anyway XD

runner/ansible/README.md
MMuschner: It was not meant to explain the setup but to tell that a restart of the services are needed. Should I shorten it or remove it?

arbulu89: I would totally remove it. If you want, you can mention that to get the new tests in the application, obviously the application must be re-built

stefanotorresi (changes requested):
This is great, but see https://github.com/trento-project/trento/pull/282#discussion_r720074627

We should probably distinguish documentation about the runner component and documentation about how to write a check. These are related, but separate.

I guess what is being worked on here should be kind of a user-friendly how-to, while the "workflow" described is more like developer notes about how to work with the runner.

runner/ansible/README.md
```diff
@@ -4,7 +4,13 @@ Place to store Ansible playbooks that are executed by the Trento runner
 
 ## Implementing a check
 
-To be defined...
+# Table of contents
+
+- [Structure]
+   - [Metadata files](#metadata-files)
+   - [Check files](#check-files)
```
arbulu89, line 11: This line has an extra space before `[Check files]`

runner/ansible/README.md
```diff
@@ -4,7 +4,13 @@ Place to store Ansible playbooks that are executed by the Trento runner
...
+   -[Check structure](#check-structure)
```
arbulu89, line 12: Let's put the `Check structure` 1st in the list

runner/ansible/README.md
```diff
@@ -22,3 +28,102 @@ These are the fields needed by Trento:
...
+The checks folder is in `runner/ansible/roles/checks`. Each check has its own
+number (e.g. 1.1.1) which refers to its place in the queue (1.1.2 will be 
+executed after 1.1.1 etc.). Inside these numbered folders are two subfolders, both of these each has one file in them named `main.yml`:
+- `defaults`
+   The defaults directory (or better the main.yml file inside of it)
+   contains all the required [metadata](#metadata-files) for the check. 
+
+- `tasks`
+
+The checks themselfs are written like any ansible task, however their
```
arbulu89, line 49: I think all this description about the `task` should go in the `Check file` section

arbulu89: Typo `themselfs -> themselves`

runner/ansible/README.md
```diff
@@ -22,3 +28,102 @@ These are the fields needed by Trento:
...
+    - import_role:
+        name: post-results
+  when:
+    - ansible_check_mode
+  vars:
+    status: "{{ config_updated is not changed }}"
+``` 
+
+
+After implementing your checks, trento needs to be rebuild, moved to the
```
arbulu89, line 76: I would remove this paragraph.

runner/ansible/README.md
```diff
@@ -22,3 +28,102 @@ These are the fields needed by Trento:
...
+``` 
+
+
+After implementing your checks, trento needs to be rebuild, moved to the 
+target machine and before beeing started, all running services related to
+trento need to be stopped.
+
+## Examples
+
+As an example (we will use 3.0.0), lets add a test which simply creates a simple file in /tmp.
```
arbulu89, line 82: I wouldn't reference any existing test (it might go away at some point). Make an example with the `dummy` name or similar
Besides this, I would actually implement a real check (which checks something).
Like check the file `/tmp/my_file` exists or similar.
We need to emphasize that the check must not apply any change in the machine, itself. It is a `read-only` operation

arbulu89 (changes requested):
Hi @MMuschner ,
More suggestions :)

runner/ansible/README.md
```diff
@@ -22,3 +41,84 @@ These are the fields needed by Trento:
...
 - `description`: A longer description about the test purpose. It can be written using markdown.
 - `implementation`: Usually the task `main.yml` content
+
+## Check files
+
+The check files contain the actual task which the runner will execute. An
+example is available at [tasks/main.yml](roles/checks/1.1.1/tasks/main.yml).
+These files are written like a normal ansible task, but:
+
+- The tasks are of `read only` nature, meaning that they are meant to check for things
```
arbulu89, line 51: This 1st bullet list entry looks confusing. I would rewrite it as:
`The checks are of `read-only` nature, so they must not apply any change on the system.

runner/ansible/README.md
```diff
@@ -22,3 +41,84 @@ These are the fields needed by Trento:
...
+        name: post-results
+  when:
+    - ansible_check_mode
+  vars:
+    status: "{{ config_updated is not changed }}"
+``` 
+
+## Examples
+
+As an example (we will use `dummy`), lets add a test which simply creates a simple file in /tmp.
```
arbulu89, line 78: This paragraph must be updated to match with the dummy check

runner/ansible/README.md
```diff
@@ -22,3 +41,84 @@ These are the fields needed by Trento:
...
+
+- block:
+    - import_role:
+        name: post-results
+  when:
+    - ansible_check_mode
+  vars:
+    status: "{{ config_updated is not changed }}"
+``` 
+
```
arbulu89, line 122: Remove the leftover final empty lines

runner/ansible/README.md
```diff
@@ -22,3 +41,84 @@ These are the fields needed by Trento:
...
+```
+---
+
+- name: "{{ id }}.check"
+  stat:
+      path: "/etc/os-release"
+  check_mode: no
+  register: config_updated
+  changed_when: config_updated.stdout != expected[id]
```
arbulu89, line 112: This `expected[id]` might lead to a confusion, as it is not mentioned anywhere.
As a summary, the `expected` values are in the `vars` folder, and they might be updated by the user (from the web interface in the future)

arbulu89 (changes requested):
Hey @MMuschner ,
Really looking good.
More comments from my side.
PD: This file has been moved to `docs/runner.md` and actually has some updates, so you will have some conflicts for sure

PR #291, opened 2021-09-30
Cluster health details view makeover

dottorblaster:
This PR introduces React into the project, with a minimal Webpack configuration alongside it. To put the power of React at work from the first moment, we decided to directly implement a new feature, that is the checks' hierarchical view so users can spare some time and have checks divided by category.

This first implementation uses the same accordions from the Checks Catalog, and it's a first step towards the mockups we were provided.

### Notes for the reviewers
You'll see some things that maybe you won't grasp at a first glance. It's totally normal, as JS is a totally different language than (e.g.) Golang, used in the backend side. I'd like to give a presentation of this to the rest of the team, so feel free to ask questions!

Here you can see a screenshot of how the result looks like:

2021-09-30-165019_1764x1381_scrot

### DIFF
--- a/web/api.go
+++ b/web/api.go
@@ -331,16 +331,19 @@ func ApiSAPSystemDeleteTagHandler(client consul.Client) gin.HandlerFunc {
 // @Success 200 {object} map[string]interface{}
 // @Error 500
 // @Router /api/clusters/{id}/results
-func ApiCheckResultsHandler(client consul.Client, s services.ChecksService) gin.HandlerFunc {
+func ApiClusterCheckResultsHandler(client consul.Client, s services.ChecksService) gin.HandlerFunc {
 	return func(c *gin.Context) {
-		clusterId := c.Param("id")
-		checkResults, errResult := s.GetChecksResultByCluster(clusterId)
-		checksCatalog, errCatalog := s.GetChecksCatalog()
-		if errResult != nil {
-			c.Error(errResult)
+		clusterId := c.Param("cluster_id")
+
+		checkResults, err := s.GetChecksResultByCluster(clusterId)
+		if err != nil {
+			c.Error(err)
 			return
-		} else if errCatalog != nil {
-			c.Error(errCatalog)
+		}
+
+		checksCatalog, err := s.GetChecksCatalog()
+		if err != nil {
+			c.Error(err)
 			return
 		}
 
--- a/web/api.go
+++ b/web/api.go
@@ -4,11 +4,13 @@ import (
 	"net/http"
 
 	"github.com/gin-gonic/gin"
+
 	"github.com/trento-project/trento/internal/cluster"
 	"github.com/trento-project/trento/internal/consul"
 	"github.com/trento-project/trento/internal/hosts"
 	"github.com/trento-project/trento/internal/sapsystem"
 	"github.com/trento-project/trento/internal/tags"
+	"github.com/trento-project/trento/web/models"
 	"github.com/trento-project/trento/web/services"
 )
 
@@ -347,12 +349,17 @@ func ApiClusterCheckResultsHandler(client consul.Client, s services.ChecksServic
 			return
 		}
 
+		resultSet := []models.ClusterCheckResults{}
 		for checkId, check := range checkResults.Checks {
-			check.Group = checksCatalog[checkId].Group
-			check.Description = checksCatalog[checkId].Description
-			check.ID = checksCatalog[checkId].ID
+			current := models.ClusterCheckResults{
+				Group:       checksCatalog[checkId].Group,
+				Description: checksCatalog[checkId].Description,
+				Hosts:       check.Hosts,
+				ID:          checksCatalog[checkId].ID,
+			}
+			resultSet = append(resultSet, current)
 		}
 
-		c.JSON(http.StatusOK, checkResults)
+		c.JSON(http.StatusOK, resultSet)
 	}
 }
--- a/web/api.go
+++ b/web/api.go
@@ -9,6 +9,7 @@ import (
 	"github.com/trento-project/trento/internal/hosts"
 	"github.com/trento-project/trento/internal/sapsystem"
 	"github.com/trento-project/trento/internal/tags"
+	"github.com/trento-project/trento/web/services"
 )
 
 func ApiPingHandler(c *gin.Context) {
@@ -321,5 +322,34 @@ func ApiSAPSystemDeleteTagHandler(client consul.Client) gin.HandlerFunc {
 		}
 
 		c.JSON(http.StatusNoContent, nil)
+	}
+}
+
+// ApiCheckResultsHandler godoc
+// @Summary Get a specific cluster's check results
+// @Produce json
+// @Success 200 {object} map[string]interface{}
+// @Error 500
+// @Router /api/clusters/{id}/results
+func ApiCheckResultsHandler(client consul.Client, s services.ChecksService) gin.HandlerFunc {
+	return func(c *gin.Context) {
+		clusterId := c.Param("id")
+		checkResults, errResult := s.GetChecksResultByCluster(clusterId)
+		checksCatalog, errCatalog := s.GetChecksCatalog()
+		if errResult != nil {
+			c.Error(errResult)
+			return
+		} else if errCatalog != nil {
+			c.Error(errCatalog)
+			return
+		}
+
+		for checkId, check := range checkResults.Checks {
+			check.Group = checksCatalog[checkId].Group
+			check.Description = checksCatalog[checkId].Description
+			check.ID = checksCatalog[checkId].ID
+		}
+
+		c.JSON(http.StatusOK, checkResults)
 	}
 }
--- a/web/frontend/components/Accordion/Accordion.js
+++ b/web/frontend/components/Accordion/Accordion.js
@@ -1,0 +1,28 @@
+import React, { useState } from 'react';
+import classNames from 'classnames';
+
+const Accordion = ({ className, title, children }) => {
+  const [isOpen, setOpen] = useState(true);
+  const collapseClassNames = classNames('collapse', { show: isOpen });
+  const iconClassNames = classNames(
+    'eos-icons',
+    'eos-18',
+    'collapse-toggle',
+    'clickable',
+    { collapsed: !isOpen }
+  );
+
+  return (
+    <div className={`${className} card`}>
+      <div className="card-header">
+        <h4 className="float-left">{title}</h4>
+        <i className={iconClassNames} onClick={() => setOpen(!isOpen)}></i>
+      </div>
+      <div className={collapseClassNames}>
+        <div className="card-body">{children}</div>
+      </div>
+    </div>
+  );
+};
+
+export default Accordion;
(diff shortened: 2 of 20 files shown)


web/api.go
```diff
@@ -323,3 +324,32 @@ func ApiSAPSystemDeleteTagHandler(client consul.Client) gin.HandlerFunc {
...
+		clusterId := c.Param("id")
+		checkResults, errResult := s.GetChecksResultByCluster(clusterId)
+		checksCatalog, errCatalog := s.GetChecksCatalog()
+		if errResult != nil {
+			c.Error(errResult)
+			return
+		} else if errCatalog != nil {
+			c.Error(errCatalog)
+			return
+		}
```
fabriziosestito, line 345: Usually it is idiomatic to check for the error near where the error could happen 
```
	return func(c *gin.Context) {
		clusterId := c.Param("id")
		checkResults, err  := s.GetChecksResultByCluster(clusterId)
		if err != nil {
			c.Error(errResult)
			return
		} 
		
		checksCatalog, err  := s.GetChecksCatalog()
		if err != nil {
			c.Error(errCatalog)
			return
		} 
		```

File: web/frontend/javascripts/check_results.js [hunk 2]
```diff
@@ -0,0 +1,59 @@
...
+  });
+};
+
+const ClustersChecks = ({ clusterId }) => {
+  const [results, setResults] = useState([]);
+
+  useEffect(() => {
+    get(`/api/clusters/${clusterId}/results`).then(({ data: { checks } }) => {
+      const groupedChecks = groupChecks(checks);
+      console.log(groupedChecks);
```
fabriziosestito, line 31: `console.log` leftover

Shouldn't we catch this with the linter?

web/frontend/components/Accordion/Accordion.js
```diff
@@ -0,0 +1,28 @@
...
+        <i className={iconClassNames} onClick={() => setOpen(!isOpen)}></i>
+      </div>
+      <div className={collapseClassNames}>
+        <div className="card-body">{children}</div>
+      </div>
+    </div>
+  );
+};
+
+export default Accordion;
```
fabriziosestito, line 28: Lovely!

web/api.go
```diff
@@ -323,3 +324,32 @@ func ApiSAPSystemDeleteTagHandler(client consul.Client) gin.HandlerFunc {
...
+
+		for checkId, check := range checkResults.Checks {
+			check.Group = checksCatalog[checkId].Group
+			check.Description = checksCatalog[checkId].Description
+			check.ID = checksCatalog[checkId].ID
+		}
+
+		c.JSON(http.StatusOK, checkResults)
+	}
+}
```
fabriziosestito, line 355: Let's add a tests for this API

web/api.go
```diff
@@ -323,3 +324,32 @@ func ApiSAPSystemDeleteTagHandler(client consul.Client) gin.HandlerFunc {
...
+
+// ApiCheckResultsHandler godoc
+// @Summary Get a specific cluster's check results
+// @Produce json
+// @Success 200 {object} map[string]interface{}
+// @Error 500
+// @Router /api/clusters/{id}/results
+func ApiCheckResultsHandler(client consul.Client, s services.ChecksService) gin.HandlerFunc {
+	return func(c *gin.Context) {
+		clusterId := c.Param("id")
```
arbulu89, line 336: This `id` might be a bit confusing/ What about `cluster_id`?

Or otherwise we could rename the func name to be specific for the cluster. I forsee other checks out of the cluster topic in the future

web/frontend/components/ChecksTable/CheckResultIcon.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+
+const PASSING = 'passing';
```
arbulu89, line 3: We have the SKIPPED state as well (which uses the `fiber_manual_record` type of icon).
Maybe useful to know, if we want to add tooltips or similar at some point

web/models/check.go
```diff
@@ -28,7 +28,10 @@ type Results struct {
 
 // The ChecksByHost struct stores the checks list, but the results are grouped by hosts
 type ChecksByHost struct {
-	Hosts map[string]*Check `json:"hosts,omitempty" mapstructure:"hosts,omitempty"`
+	ID          string            `json:"id,omitempty" mapstructure:"id,omitempty"`
```
arbulu89, line 31: I don't really get the purpose of this change. Is this to send the information in the new API handler?
Maybe we should create new struct.
This was originally designed to parse the data coming from ARA.
Anyway, we can discuss it

web/layout.go
```diff
@@ -152,15 +152,21 @@ func (r *LayoutRender) addFileFromFS(templatesFS fs.FS, file string) {
...
+		"split":         strings.Split,
+		"includeBundle": includeBundle,
 	})
 	patterns := append([]string{r.root, file}, r.blocks...)
 	tmpl = template.Must(tmpl.ParseFS(templatesFS, patterns...))
 
 	r.addTemplate(name, tmpl)
 }
 
+func includeBundle(filename string) template.HTML {
```
arbulu89, line 165: We don't really need this method.
We can add any specific js file in the template itself.
Check this as example:
https://github.com/trento-project/trento/blob/main/web/templates/clusters.html.tmpl#L2

We could this way simply change the template file to include the needed js script

arbulu89 (commented):
Hey @dottorblaster ,
Love the changes! Really well done.

There are anyway things to be discussed.
As you say, we need to have some introductory life discussion about it.

But I will throw here the 1st question XD:
- I don't see how we manage the `Unreachable` case. This means that we don't have any result for the some of the nodes in the cluster (or both). Maybe we need to check more this scenario

PD: I already did some changes in this page in other PR. I think it makes sense to merge this 1st, so I can practice afterwards to adapt to this new tecnique

web/api.go
dottorblaster: I'll do that in a further PR if I can :innocent:

web/frontend/javascripts/check_results.js
dottorblaster: I'm gonna add the rule to the linter, you're right

web/api.go
dottorblaster: Addressed, you're right

web/frontend/components/ChecksTable/CheckResultIcon.js
dottorblaster: Yeah, I defaulted to it, thanks for discussing! Do you prefer another default?

web/layout.go
dottorblaster: You're right, we can actually just include raw `script` tags but this allows us

- To do that in a much cleaner way
- To directly reference a webpack bundle

Other frameworks adopt similar techniques to reference static paths, also: https://docs.djangoproject.com/en/3.2/howto/static-files/

web/models/check.go
dottorblaster: I think you're actually right, let's see if I can come up with something better

dottorblaster:
Hey @arbulu89,

> I don't see how we manage the Unreachable case. This means that we don't have any result for the some of the nodes in the cluster (or both). Maybe we need to check more this scenario

Yes I think so :sweat_smile: 

> I think it makes sense to merge this 1st, so I can practice afterwards to adapt to this new tecnique

Yep, this PR is becoming huge and I think it's reasonable to merge this and iterate upon the code in the coming days :)))

web/frontend/components/ChecksTable/CheckResultIcon.js
arbulu89: Yes, add other variable as `SKIPPED = 'skipped'`. Even that it will use the same icon, let's have it in the code to have a better picture

web/layout.go
arbulu89: I don't really get your point. I don't mean to include raw js code.
As far as I understand, here you only want to add the `<script src=\"/static/...>` element to the html directly (the link to the script).
We already do this in other parts of the code (include statis js files), like in the example link I put before.
As far as I see, you just implemented something we are already doing (and the current way looks better in my opinion).

Or am I missing something?

arbulu89:
> Hey @arbulu89,
> 
> > I don't see how we manage the Unreachable case. This means that we don't have any result for the some of the nodes in the cluster (or both). Maybe we need to check more this scenario
> 
> Yes I think so sweat_smile
> 
> > I think it makes sense to merge this 1st, so I can practice afterwards to adapt to this new tecnique
> 
> Yep, this PR is becoming huge and I think it's reasonable to merge this and iterate upon the code in the coming days :)))

No issues on doing that in coming PRs, as long as we create follow up issues, and that we are sure that they will be ready for the upcoming new release (we don't want to remove currently existing features hehe)

web/layout.go
dottorblaster: Sorry, I meant raw `script` tags, not raw JS code :D

> We already do this in other parts of the code (include statis js files)

Yeah my point is that we shouldn't do that because if we modify the static assets directory in the future we will have to go manually through the whole project changing references to `/static/frontend/assets/js` :) that's why frameworks like Django or Phoenix have a `static_path` or `static` annotation for templates, that allows you to reference that path in a cleaner way. Also, I specifically created this for webpack bundles so we only have to worry about the actual filename, and not the whole path.

This allows us to streamline adding a new webpack bundle, just adding a new file, editing the webpack configuration and using this function in the template referencing the filename.

> As far as I see, you just implemented something we are already doing (and the current way looks better in my opinion).

IMHO it does not, but it's just my opinion :D

arbulu89: Ok, now I understand better the motivation.
If you feel this is the way to go, just open other follow up ticket, as we should use the same pattern in all the places where we import static js files

dottorblaster: You're right, I have a plan for the other JS files :D

rtorrero: Hadn't thought of it before, but I like this approach :-)

rtorrero (approved):
I still need to fill a lot of react knowledge into my brain, but it looks good. Thanks!

web/layout.go
stefanotorresi: To be honest, I agree with Xabi: I don't find this function particularly useful, since you are just moving a string interpolation from the actual template to a template function, but at the end of the day it's still just a path interpolated with a file name.

That said, if you want to keep it it's fine, but `includeBundle` doesn't seem exactly fitting, because you could include any kind of javascript file.
As a suggestion, consider calling it just `script()`, so that the end result in the template is `{{ script "check_results.js" }}`, which resembles what the original HTML was.

stefanotorresi (changes requested):

web/layout.go
stefanotorresi: As a side note, full-fledged web application frameworks have these kind of functions because they usually also have a super magical black boxes of complex asset management systems, which we fortunately don't. 😛

dottorblaster: Well I agree with you to some extent but what really Phoenix, Rails and Symfony expose under different names (like Symfony Encore) is just webpack with a sensible-defaults-oriented configuration, a thing that we fortunately do as well, even if we didn't rebrand it as "Trento Fireworks" :joy:

Anyway, ok I'll rename that to `script` if that looks reasonable to you :sweat_smile:

dottorblaster:
@arbulu89 @stefanotorresi I think this is ok to merge now, I created a new struct type for the results and changed `includeBundle` to `script` :smile:

arbulu89 (approved):
Thanks @dottorblaster

stefanotorresi (approved):

PR #292, opened 2021-10-01
Adjust scrips location

nelsonkopliku:
Moved back some scripts into `hack` folder.


fabriziosestito (approved):

aleksei-burlakov:
Please wait, Im not sure it would work.

.github/workflows/ci-cd.yaml
```diff
@@ -97,7 +97,7 @@ jobs:
       - name: compress
         run: |
           set -x
-          find ./build -maxdepth 1 -mindepth 1 -type d -exec sh -c 'tar -zcf build/trento-$(basename {}).tgz -C {} trento -C $(pwd) trento-agent.service' \;
+          find ./build -maxdepth 1 -mindepth 1 -type d -exec sh -c 'tar -zcf build/trento-$(basename {}).tgz -C {} trento -C $(pwd) packaging/systemd/trento-agent.service' \;
```
aleksei-burlakov, line 100: It should be rather 
$ find ./build -maxdepth 1 -mindepth 1 -type d -exec sh -c 'tar -zcf build/trento-$(basename {}).tgz -C {} trento -C $(pwd)/packaging/systemd trento-agent.service' \;

aleksei-burlakov: I see you extract it also from the folder in install-agent.sh:219. It should work then. I don't think however putting the service file into the packaging/suse inside the tgz is a good idea

PR #296, opened 2021-10-01
Remove docker-compose.yml

stefanotorresi:
It was never actually used anyway, and it is now superseded by the Helm charts.


dottorblaster (approved):
Actually a lil misleading for me too. APPROVED

PR #297, opened 2021-10-01
Fix trento-server chart name

fabriziosestito:
The chart name could be referenced by other charts and the name includes spaces, capital letters, etc..
Changed to an idiomatic name.


stefanotorresi:
hah, I wasn't aware of this! Nice catch.

stefanotorresi (approved):

PR #298, opened 2021-10-01
Attempt clean-up _before_ installing

stefanotorresi:
If we don't clean up before installing, a failed installation will leave the wrong sources hanging around in `/tmp`, causing all sorts of troubles at the next CD run!


rtorrero (approved):
LGTM

PR #299, opened 2021-10-01
Disable gin debug logging on demand

dottorblaster:
Since Gin's debug logs in tests are super noisy to us,I thought that it would be a nice idea to change the behavior using an env var :-)


arbulu89 (commented):
@dottorblaster I'm not sure if we need this.
`gin-gonic` already has an internal Env variable for this `GIN_MODE`.
Maybe, we should just document how to use it.
And we could set this variable to `test` in our tests

Ref:
https://github.com/gin-gonic/gin/blob/master/mode.go#L15

dottorblaster:
@arbulu89 wow, I think you're right :open_mouth:

arbulu89:
> @arbulu89 wow, I think you're right open_mouth

I actually have been testing, and I didn't make it work... I don't even know if this `init` method to set the mode is actually called anywhere.

I just did the next:

Add the next snippet in the `NewAppWithDeps`"
```
mode := os.Getenv(gin.EnvGinMode)
gin.SetMode(mode)
```

In the `Makefile`, change the test call to `GIN_MODE=test go test -v ./...`

This makes the test output cleaner

dottorblaster:
Ah yeah absolutely, it worked for me because I was doing this for testing purposes and with make test it worked seamlessly :)

stefanotorresi:
So, what are we going to do? I would be totally fine with what Xabi just suggested, and defer the integration with the `--log-level` flag to another time.

dottorblaster:
@stefanotorresi @arbulu89 I like that, I'll update the PR accordingly, thank you so much!

stefanotorresi (dismissed):
approved with nitpick 😜

stefanotorresi (approved):
you rule! 🍻

PR #300, opened 2021-10-01
add more guidelines to the release how-to

stefanotorresi:
The document was hastly introduced without review in 0d6086da455c8f6a335a45195b5cac0c3d1b0e0e.

The intent is that most of the things explained there will be automated in the close future, but the indications on how to use GH labels will likely remain.


nelsonkopliku (approved):
LGTM

Issue #301, opened 2021-10-01
container images are not built with the correct version constant


stefanotorresi:
as in
```
vmmonitoring:/home/cloudadmin # kubectl exec trento-server-trento-runner-69c7c8df8-fxhv4 -- /app/trento version
Trento version 0
built with go1.16.8 linux/amd64
```

We probably need to introduce a build argument.

PR #302, opened 2021-10-04
Fetch tags before building container images

fabriziosestito:
We need to fetch the tags before building images so that the script can get the right version.

Fixes: #301


stefanotorresi (changes requested):
Pro tip: no need for a new step, you can just add `fetch-depth: 0` to the checkout one.

stefanotorresi (approved):

PR #306, opened 2021-10-04
Add PostgreSQL

fabriziosestito:
This PR adds PostgreSQL persistence layer and moves the tags implementation on top of it.

GORM is used as an ORM adapter, we also introduce a service layer to access the persistence.
Posgres was added to the CI to be able to integration test the services.

It also adds an Helm chart to be used as a development environment, as for now it just override the postgres chart values of the official chart, but we have plans to extend it (for example with the agent mocks) in a near future.
By starting the helm chart, two databases are available: one for the development and the other one that is used when integration testing the services.


hack/helm/trento-dev/Chart.lock
```diff
@@ -0,0 +1,6 @@
+dependencies:
+- name: trento-server
+  repository: file://../../../packaging/helm/trento-server
+  version: 0.1.0
+digest: sha256:ca1dd3126e0344280cff27f3cb8c67545c8679aaddc3644d9690154d2b7d59c2
```
rtorrero, line 5: Does this mean that we will have to manually update this each time the trento-server helm chart gets updated? or does it happen automatically?

web/web_test.go
```diff
@@ -0,0 +1,13 @@
+package web
+
+import (
+	"github.com/gin-contrib/sessions/cookie"
+	"github.com/gin-gonic/gin"
+)
+
+func testDependencies() Dependencies {
```
rtorrero, line 8: As you suggested, maybe `setupTestDependencies` is a better name?

cmd/web/web.go
```diff
@@ -26,6 +36,22 @@ func NewWebCmd() *cobra.Command {
 	serveCmd.Flags().IntVarP(&port, "port", "p", 8080, "The port for the HTTP service to listen at")
 	serveCmd.Flags().StringVar(&araAddr, "ara-addr", "127.0.0.1:8000", "Address where ARA is running (ex: localhost:80)")
 
+	serveCmd.Flags().StringVar(&dbhost, "dbhost", "localhost", "The database host")
+	serveCmd.Flags().StringVar(&dbport, "dbport", "5432", "The database port to connect to")
```
fabriziosestito, line 40: @nelsonkopliku maybe we should use db-port, db-host etc for consistency with ara-addr and the other flags

rtorrero (commented):
I really like the overall idea how you did this :-) GJ!

cmd/web/web.go
```diff
@@ -26,6 +36,22 @@ func NewWebCmd() *cobra.Command {
...
+	serveCmd.Flags().StringVar(&dbuser, "dbuser", "postgres", "The database user")
+	serveCmd.Flags().StringVar(&dbpassword, "dbpassword", "postgres", "The database password")
+	serveCmd.Flags().StringVar(&dbname, "dbname", "trento", "The database name that the application will usee")
```
arbulu89, line 43: Typo in `usee`

cmd/web/web.go
```diff
@@ -26,6 +36,22 @@ func NewWebCmd() *cobra.Command {
...
+
+	// Bind the flags to viper and make them available to the application
+	serveCmd.Flags().VisitAll(func(f *pflag.Flag) {
+		// Needed if the flags contain dashes
+		if strings.Contains(f.Name, "-") {
+			viper.BindEnv(f.Name, strings.ToUpper(strings.ReplaceAll(f.Name, "-", "_")))
+		}
+
+		viper.BindPFlag(f.Name, f)
```
arbulu89, line 52: As a question. In the flags where we have the `-`, the name is uppercased, but here it is not.
Is this done by viper? Or it is just forgotten?

web/api.go
```diff
@@ -30,14 +29,12 @@ type JSONTag struct {
 // @Success 200 {object} []string
 // @Failure 500 {object} map[string]string
 // @Router /api/tags [get]
-func ApiListTag(client consul.Client) gin.HandlerFunc {
+func ApiListTag(client consul.Client, tagsService services.TagsService) gin.HandlerFunc {
```
arbulu89, line 32: Is there any reason why the consul client is still passed as parameter?
It is not used within the function, and actually, the tags don't use consul at all now.

web/app.go
```diff
@@ -42,27 +45,61 @@ type Dependencies struct {
...
 
 func DefaultDependencies() Dependencies {
 	consulClient, _ := consul.DefaultClient()
 	engine := gin.Default()
 	store := cookie.NewStore([]byte("secret"))
 
-	araService := ara.NewAraService(araAddrDefault)
+	db, err := InitDB()
+	if err != nil {
+		log.Fatalf("failed to connect database: %s", err)
```
arbulu89, line 58: I see a logged message only here. shouldn't we panic the application? Or enter in a connection tries loop?

hack/helm/trento-dev/Chart.lock
fabriziosestito: correct

web/clusters.go
```diff
@@ -296,7 +295,7 @@ type ClustersRow struct {
 
 type ClustersTable []*ClustersRow
 
-func NewClustersTable(s services.ChecksService, client consul.Client, clusters map[string]*cluster.Cluster) (ClustersTable, error) {
+func NewClustersTable(s services.ChecksService, t services.TagsService, client consul.Client, clusters map[string]*cluster.Cluster) (ClustersTable, error) {
```
arbulu89, line 298: Is the consul client parameter used here for anything?

arbulu89 (changes requested):
Great work @fabriziosestito and @nelsonkopliku 

I have commented some potential improvements

cmd/web/web.go
fabriziosestito: Viper skips all the names that are considerd invalid in the `AutoEnv()` method called by `root.cmd` at the end of the command setup.
 
`ara-addr` is one of them so you need to call manually `BindEnv` , otherwise it will not work.

web/clusters.go
fabriziosestito: leftover, thx

web/hosts.go
```diff
@@ -4,7 +4,7 @@ import (
 	"net/http"
 	"strings"
 
-	"github.com/trento-project/trento/internal/tags"
+	"github.com/trento-project/trento/internal"
```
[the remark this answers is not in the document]
fabriziosestito: we use `internal.Contains` so it is needed

cmd/web/web.go
nelsonkopliku: Sure, makes sense to me

web/app.go
fabriziosestito: P.S.: we wrote down a tech-debt about retry/backoff mechanisms, as for now exiting is good enough

web/api.go
fabriziosestito: leftover, thx

cmd/web/web.go
fabriziosestito: EDIT: we found a way to improve it and make this bnehaviour automatic for all the "dashed" commands, just fixup/squashed it

PR #313, opened 2021-10-05
Get correct version to OBS using git release tags

rtorrero:
This PR should fix the issues with the automated RPM generation on OBS by making sure the `rolling` tag doesn't get picked up.

We could probably use a similar approach to get rid of the `get_version_from_git.sh` script

### DIFF
--- a/.github/workflows/ci-cd.yaml
+++ b/.github/workflows/ci-cd.yaml
@@ -251,7 +251,7 @@ jobs:
         /scripts/init_osc_creds.sh
         mkdir -p $HOME/.config/osc
         cp /root/.config/osc/oscrc $HOME/.config/osc
-    - name: prepare tranto.changes file
+    - name: Prepare trento.changes file
       run: |
         VERSION=$(./hack/get_version_from_git.sh)
         TAG=$(echo $VERSION | cut -f1 -d+)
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-cd.yaml
```diff
@@ -251,14 +257,14 @@ jobs:
         /scripts/init_osc_creds.sh
         mkdir -p $HOME/.config/osc
         cp /root/.config/osc/oscrc $HOME/.config/osc
-    - name: prepare tranto.changes file
+    - name: Prepare trento.changes file
       run: |
-        VERSION=$(./hack/get_version_from_git.sh)
+        VERSION=${{steps.getvers.outputs.tag}}
```
aleksei-burlakov, line 262: It would loose the integer date and the commit-hex in the suffix.
for example it would be `trento-0.4.1.tar.gz` instead of `trento-0.4.1+dev51.1633446165.75e2eef.tar.gz`

rtorrero: yep, nice catch. I thought in the beggining that the relevant thing would have been to have the correct version in the trento binary but I guess this would prevent it from properly updating the RPM after a new rolling RPM is out.

.github/workflows/ci-cd.yaml
```diff
@@ -251,14 +257,14 @@ jobs:
...
         TAG=$(echo $VERSION | cut -f1 -d+)
```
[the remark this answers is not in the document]
rtorrero: ok, done

aleksei-burlakov (approved):
LGTM!

Issue #308, opened 2021-10-05
HANA Databases page misses tags API, so tags are broken


fabriziosestito:
We are missing tags api for the database resource, we need to implement that in order to fix the tags in the HANA Databases page.

Good PR for Hacktoberfest here ;)

**NOTE:**
Please defer this issue after the merging of #306, since it re-implements the tag system

PR #318, opened 2021-10-07
Fix tags usage in the databases view

arbulu89:
Fix for #308 

The PR adds a new `api/databases/...` endpoint to manage the tags. It mostly duplicates the code for the sap systems.

I have changed a bit the sapsystems code to use the sapsystems service rather than consul. Which makes things a bit more cleaner.

It is true that the test code is almost a copy/paste, and I don't know if it makes sense to have something better.

PD: I don't know if it would make sense to add the `make swag` to the generic `make` command.
I have noticed that some entries were not added

### DIFF
--- a/docs/docs.go
+++ b/docs/docs.go
@@ -24,12 +24,21 @@ var doc = `{
     "host": "{{.Host}}",
     "basePath": "{{.BasePath}}",
     "paths": {
-        "/api/clusters/{id}/results": {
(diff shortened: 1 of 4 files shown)


stefanotorresi:
> PD: I don't know if it would make sense to add the `make swag` to the generic `make` command.
> I have noticed that some entries were not added

Yes, definitely; we'll also need to publish the API docs at some point. I will file a follow-up for this.

PR #336, opened 2021-10-14
Add minimum ansible version in the documentation and dockerfile

arbulu89:
In order to add more information to the manual installation (and even in the dockerfile), I have included the minimum tested ansible and ara versions to the README file. We already had some issue that the runner was not working if the installed ansible version was older than this.

**I would like to discuss whether we should use exact versions or "at least" versions. At least, in the dockerfile**


arbulu89:
@nelsonkopliku I have force pushed to use the `~=` option, which will make the installation safer. This way, if we find some real security or feature update in Ansible we want to do, we will do it on purpose.

PR #341, opened 2021-10-15
Add secure data collector endpoint

fabriziosestito:
This PR adds the projector infrastructure and the secure mTLS endpoint that collects discovery data from the agents.

- The secure endpoint is started on web command `serve` as a new server
- A `datapipeline` package is added to web and adds the projector logic
- The data received from the collector endpoint is stored in a event table in the pgsql db, and then sent for processing to a pool of workers go routines that use the a projector registry to run the projectors
- Handlers abstract the transformation of data in/data out aka collected data -> read model
- A cluster list "next " view was added to start strangling the old kv store views, this is hidden from the users but we will switch as soon as we have the agent part of the new architecture, and it is a good starting point to demonstrate how the new architecture works.
- Some much needed refactors in web tests and additional chores
---
**Note on mocks:**
we had mocks in the dedicated directory `web/services/mocks` and we have them now squashed in `web/services` as `*_mock.go`.

We agree that having those in a dedicated directory is the best, 
yet we couldn't make it work properly with the additions we made :sweat: It seems to be some quirky [go stuff](https://github.com/golang/mock/issues/352). 
We thought valuable to just keep it this way for the time being until we find a neater solution.
You will understand.

---
**Quick todo list**
a more extensive one will emerge as tasks in the board
- [ ] refactor `SetupTestDatabase` in a common infrastructure init package
- [ ] define separated component dependencies and initialization layers
- [ ] properly handle multiple HANA SIDs
- [ ] refactoir Handler abstractions as we add more implementations
- [ ] tuning of projectors worker pool

Update: Architetural diagram trento-architecture

### DIFF
--- a/web/app.go
+++ b/web/app.go
@@ -186,7 +186,7 @@ func NewAppWithDeps(host string, port int, deps Dependencies) (*App, error) {
 	}
 
 	collectorEngine := deps.collectorEngine
-	collectorEngine.POST("/api/collect_data", ApiCollectDataHandler(deps.collectorService))
+	collectorEngine.POST("/api/collect", ApiCollectDataHandler(deps.collectorService))
 
 	return app, nil
 }
--- a/web/collector_test.go
+++ b/web/collector_test.go
@@ -31,7 +31,7 @@ func TestApiCollectDataHandler(t *testing.T) {
 		DiscoveryType: "discovery",
 		Payload:       []byte("{}"),
 	})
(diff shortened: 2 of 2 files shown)


nelsonkopliku:
Thanks @arbulu89 for the feedbacks!
I answered to your questions, hope I managed to confuse you, lol :joy: 

I chaged the description by adding what you mentioned: todo list and mocks explaination.

@fabriziosestito obviously feel free to integrate.

PR #342, opened 2021-10-15
Allow install-server script to fetch from different repo owners

rtorrero:
This PR improves the CD on a forked `trento` repo as it causes to fetch the images from the same fork instead of fetching them from the main `trento-project`. 

Previously, one would have to manually edit the deployment to specify the images uris if he wanted to test a change in e.g. the trento-runner or the trento-web images.

### DIFF
--- a/.github/workflows/ci-cd.yaml
+++ b/.github/workflows/ci-cd.yaml
@@ -25,7 +25,7 @@ jobs:
       postgres:
         image: postgres
         env:
-          POSTGRES_PASSWORD: postgres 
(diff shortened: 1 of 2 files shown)


aleksei-burlakov:
Good job looks good to me.

rtorrero:
I've changed it a bit and now instead allow to input the whole image repository both for trento-web and trento-runner. The repository owner is now instead a env variable that defaults to `trento-project` (or whatever the owner of the fork is, as it takes it from the predefined GH actions var.

Thanks for the feedback!

PR #344, opened 2021-10-15
Compare versions across nodes

aleksei-burlakov:
DRAFT!

1. The `2.2.X.identical/tasks/mail.yml` files are too verbose, I'm sure they could be simplified.
2. The PR is missing comparing `2.2.8` across all nodes.

@brett060102 could you please have a look? I especially wonder, how could I simplify the `2.2.X.identical/tasks/mail.yml` checks
and the 2.2.X/tasks/mail.yml too. I have replaced there the shell with `ansible_facts.packages['foo']`, and they have also become too verbose.

### DIFF
--- a/runner/ansible/roles/checks/2.2.1.identical/defaults/main.yml
+++ b/runner/ansible/roles/checks/2.2.1.identical/defaults/main.yml
@@ -1,0 +1,20 @@
+---
+
+name: 2.2.1
+group: OS and package versions
+labels: hana
+description: |
+  Test if the used OS is identical on all nodes
+remediation: |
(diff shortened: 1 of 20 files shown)


brett060102:
Just rechecked, and the current code will work so long as  the host is only in one group:
group_names
    List of groups the current host is part of

So assuming that the groups are the cluster names and a host is only in one cluster. This will work, but that
means that in all cases the loop is a single iteration. But since in ansible a node could be in more than one group, I don't see away around the loop.

Sorry for my confusion.

stefanotorresi:
superseded by #396

PR #347, opened 2021-10-18
Refactor ansible inventory creation removing consultemplate

arbulu89:
**Edit: Let's merge this once the 0.5.0 release is created, as it has direct relationship with the consul removal**

As we are in the process of removing `consul` from the infrastructure layer, this PR aims to remove the `consultemplate` dependency on the runner side, which will help to finally remove `consul` at all.

Until, now, we use `consultemplate` to create the ansible inventory file, as all the data is stored in the consul KV storage.
Now on, we need to create the inventory file with custom code. This PR implements this option.

As we still have most of the data in consul, the data is coming from there, but I have structured the code to make it easy to transition to new rest api calls.

For example, this will allow us to start removing the `check` selection and `user connection` data sets from the consul KV storage, which will be the next step.

As a bonus point, removing consultemplate makes possible the testing of the inventory creation in a pretty easy way.


fabriziosestito:
This will allow us to transition easily to the API endpoint and ditch consul completely in the future
LGTM

PR #363, opened 2021-10-21
Migrate cluster settings modal to React

dottorblaster:
Cluster settings modal uses raw bootstrap and jquery to get user input, we're migrating it to the new React setup.

This PR introduces some new things, I'll explain them briefly:

### A new webpack bundle
The most trivial of the news: a new webpack JS bundle to load our new settings modal into the HANA cluster page.

### `react-bootstrap`
As we go forward with the implementation of React components across the application, we'll need an helper to translate our design system's jquery lingo to React. I identified `react-bootstrap` for this: a widely-used component library that wraps and reimplements Boostrap components in plain JS using also React internals. It came in handy and we'll have all the components in EOS automatically broght to the React ecosystem this way.

### New API endpoints
I needed a couple additional API endpoints. See the code for the details.

### A new `Checkbox` component
The checkbox component that `react-bootstrap` offers wasn't enough. We need something that we can drop inside any context, and after a couple evenings fighting with the toolkit I found a way that satisfies me.

### The settings modal needs refactoring
I know, a 200-LOC-long React component is not a big deal for us. Thing is, the refactoring is pretty easy and I'd like to pair with someone to take them on board.


web/frontend/javascripts/cluster_check_settings.js
```diff
@@ -0,0 +1,215 @@
...
+                          <td>
+                            <Form.Control
+                              size="sm"
+                              value={settings[host]}
+                              onChange={({ target: { value } }) =>
+                                setSettings({ ...settings, [host]: value })
+                              }
+                            />
+                          </td>
+                          <td>root</td>
```
arbulu89, line 127: This `root` shouldn't come by default. This now comes from the backend

web/checks_api.go
```diff
@@ -139,7 +182,7 @@ func ApiCheckGetSettingsByIdHandler(s services.ChecksService) gin.HandlerFunc {
 // @Router /api/checks/{id}/settings [post]
 func ApiCheckCreateSettingsByIdHandler(s services.ChecksService) gin.HandlerFunc {
 	return func(c *gin.Context) {
-		id := c.Param("id")
+		clusterId := c.Param("id")
```
arbulu89, line 185: I would keep this to `id`. Actually, I renamed it before in other changes to make the check selection agnostic. We might have other pages with checks, but as all of them use the same backed, we could reuse this API call

web/checks_api.go
```diff
@@ -100,17 +122,38 @@ func ApiCreateChecksCatalogHandler(s services.ChecksService) gin.HandlerFunc {
...
+		if err != nil {
+			_ = c.Error(err)
+			return
+		}
+
+		nodes := NewNodes(s, clusterItem, hosts)
+
+		selectedChecks, err := s.GetSelectedChecksById(clusterId)
+		if err != nil {
+			selectedChecks = models.SelectedChecks{SelectedChecks: pq.StringArray{}}
```
arbulu89, line 153: Why have you `hidden` this error here?
I understand that you did this because if the `GetSelectedChecksById` is used when no checks are selected, it fails.
But I guess the same will happen with the `GetConnectionSettingsById` call.

In any case, I don't think you need to initialize a new `models.SelectedChecks` here. The `GetSelectedChecksById` already returns an empty struct of those in case of failure. This would space the import of `pq`

web/checks_api.go
```diff
@@ -100,17 +122,38 @@ func ApiCreateChecksCatalogHandler(s services.ChecksService) gin.HandlerFunc {
 // @Success 200 {object} JSONChecksSettings
 // @Failure 404 {object} map[string]string
 // @Router /api/checks/{id}/settings [get]
-func ApiCheckGetSettingsByIdHandler(s services.ChecksService) gin.HandlerFunc {
+func ApiCheckGetSettingsByIdHandler(consul consul.Client, s services.ChecksService) gin.HandlerFunc {
```
arbulu89, line 125: Now, this time,  I will play the devil advocate game XD
Why don't you use the new discovered data here instead of consul data?
We already have all the cloud data in the database. This is actually by now outdated.

Anyway, we can refactor at any time.

web/checks_api.go
```diff
@@ -100,17 +122,38 @@ func ApiCreateChecksCatalogHandler(s services.ChecksService) gin.HandlerFunc {
...
 	return func(c *gin.Context) {
-		id := c.Param("id")
+		clusterId := c.Param("id")
```
arbulu89, line 127: I wouldn't rename this to `clusterId`. The checks should be agnostic on this.
I understand that right now it is only attached to clusters, but at some point it won't be.

Maybe we should have other extra api call to get the hosts and their defaults settings

arbulu89 (changes requested):
Hey @dottorblaster !
Great work in the frontend. Nothing to complain about!

I have more comments in the backend side thought hehe:

- The connection default user is hardcoded now. This is retrieved from the cloud data currently. You could get it even from consul 
- Do we have some user notifications when the `Save changes` button is clicked? Positive or negative?
- After the changes the `check_select.js` file gets deprecated, right? Could we remove it?
- A lot of leftovers in the `web/clusters.go` file. The next variables passed in the `c.HTML` function `NewClusterHandler` should be removed: `ChecksCatalog, ConnectionData, DefaultConnectionData, ChecksResult`
- The `NewSaveClusterSettingHandler` looks deprecated. We should remove it from the `app.go` and `clusters.go` files. With this, we can remove the `checkSelectionForm` struct and the `getConnectionMap` function

web/checks_api.go
```diff
@@ -28,6 +32,24 @@ type JSONCheck struct {
 	Labels         string `json:"labels,omitempty"`
 }
 
+// ApiCheckResultsHandler godoc
+// @Summary Get the whole checks' catalog
+// @Produce json
+// @Success 200 {object} map[string]interface{}
```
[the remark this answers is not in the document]
dottorblaster: Well, this isn't a `JSONChecksCatalog' per se, what gets returned is a `GroupedCheckList`.

I'll put that in the annotation if you're ok with that.

web/checks_api.go
dottorblaster: And you're actually right! I wanted to wait until we had the hosts service in place too, either we wait until the host service hits `main` or we go live with this and refactor later as it shouldn't be a big deal. It's ok for me to wait, anyway :smile:

web/checks_api.go
dottorblaster: I didn't know about this future scope and I wish I'd known :joy: anyway, you're right but as of now this endpoint has a lot of bloatware with the whole consul cluster spattered inside, so the neutrality of the thing is way too compromised by the entire code structure I adopted, with this being a tiny detail. Anyway, I can rename this to `resourceId` and expand the TODO item, what do you think?

web/checks_api.go
dottorblaster: :thinking: I think you're right, I'll work on this.

web/checks_api.go
dottorblaster: `resourceId` will play here too, I think. Good catch!

web/frontend/javascripts/cluster_check_settings.js
dottorblaster: You're right but I wanted to wait until we had the hosts projection. Maybe not the greatest choice here, I should have talked with the team before

docs/api/docs.go
```diff
@@ -154,6 +154,23 @@ var doc = `{
                 }
             }
         },
+        "/api/checks_catalog": {
```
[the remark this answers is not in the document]
dottorblaster: It should be :thinking:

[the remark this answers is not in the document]
dottorblaster: Wrong annotation. My bad.

web/checks_api.go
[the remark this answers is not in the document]
arbulu89: What I wanted to say is that the returned `json` object shouldn't be the internal struct, it should be something specifically designed for the api (by coincidence they can have the same fields).

For example, here we could have `map[string]*JSONChecksCatalog`.
In order to get that you would need to transform the `models.GroupedCheckList` to that struct. This is what we have done in other API calls at least.

web/frontend/javascripts/cluster_check_settings.js
arbulu89: This information comes from the `cloud` discovery loop in any case.

In my opinion the connection settings are something that we could project.
Once we receive the hosts, cluster and cloud data, we create the `ConnectionSettings` entry in the database (this struct already exists in the `check_settings.go` model. We could add the `DefaultUser` column.

This way, you could just consume this table instead of getting the clusters, hosts, etc lists.
The work is moved to the projecto (which actually knows this data), and the remaining services and controllers are much easier.

This is design discussion that we can have live though

dottorblaster:
@arbulu89 I addressed almost everything now. Two things are open:

> The connection default user is hardcoded now. This is retrieved from the cloud data currently. You could get it even from consul

It is, and I wanted to wait until we had the hosts projection, so I could easily take the cloud provider/user from there. For now, I was opting for a fake it till you make it strategy :grimacing: 

> Do we have some user notifications when the Save changes button is clicked? Positive or negative?

We don't, but I want to have them! I'm struggling finding a library that lets us render notification toasts, but it's harder than I thought since nearly no library supports rendering custom toasts, and I really just want to render our own EOS toasts.

web/checks_api.go
arbulu89: As commented here https://github.com/trento-project/trento/pull/363#discussion_r742731024 we can take different design decisions (load the things from other read models, project the whole connection settings, etc).
Anyway, meanwhile we can move forward with this thing, knowing that it needs some work later.

web/clusters.go
```diff
@@ -573,26 +571,6 @@ func NewClusterHandler(client consul.Client, s services.ChecksService) gin.Handl
 			return
 		}
 
-		selectedChecks, getCheckErr := s.GetSelectedChecksById(clusterId)
-		if getCheckErr != nil {
-			StoreAlert(c, NoCheckSelected())
-		}
-
-		connectionData, getConnErr := s.GetConnectionSettingsById(clusterId)
-		defaultConnectionData, getDefConnErr := getDefaultConnectionSettings(client, clusterItem)
```
arbulu89, line 582: We would need to remove the `getDefaultConnectionSettings` and `getChecksCatalogWithSelected` functions implementations, as they are done in the frontend. 
By the way, you could get the default settings using this consul code. As you already used consul data to create the connections settings.
Whether we put all the data with consul or we wait until having the new projections, but let's do not do half of the work hehe

PD: Most probably, the `Alert` functions removed here are not used elsewhere, so they could be removed too

arbulu89 (commented):
Added an additional comment about `default user` entry

web/checks_api.go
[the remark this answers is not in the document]
dottorblaster: Hm, you're right, I'll work on that

web/clusters.go
dottorblaster: I will hold the works on this because we're in the process to get the hosts projected, you're right, we shouldn't rush :muscle:

nelsonkopliku (approved):
Approved after merge :joy:

PR #396, opened 2021-10-27
Compare versions across nodes

aleksei-burlakov:
Compare versions across nodes

This PR implements the draft #344 . There are two commits. The first commit (the one with the comment "Check versions across nodes") is ready to merge. There is, however, even after polishings, a lot of repeating text. It would be nice to have something like in the second commit (the one with the comment "It would be nice if it was working like this"). The second commit unfortunately doesn't work.

@arbulu89 @brett060102 what do you think?

### DIFF
--- a/runner/ansible/roles/checks/2.2.2.identical/defaults/main.yml
+++ b/runner/ansible/roles/checks/2.2.2.identical/defaults/main.yml
@@ -1,0 +1,20 @@
+---
+
+name: 2.2.2.identical
+group: OS and package versions
+labels: hana
+description: |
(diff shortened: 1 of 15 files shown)


stefanotorresi:
@aleksei-burlakov how is this PR looking? It would be great if we delivered our first cluster-wide checks with 0.6.0.

aleksei-burlakov:
> @aleksei-burlakov how is this PR looking? It would be great if we delivered our first cluster-wide checks with 0.6.0.

@stefanotorresi  Sorry for the late response. Those are the premium checks, I will create a new PR in the trento-premium.

Issue #354, opened 2021-10-19
Restore manual triggering of the CI/CD


fabriziosestito:
This got lost in the new CI/CD iteration, let's restore it to have an easy way to re-trigger the deploy on our forks/branches.

PR #406, opened 2021-10-28
fix CI issues

### LINKED ISSUE #389: Limit concurrency to 1 in the CI
We should limit the concurrency to 1 in the CI, following -> https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#concurrency

This will save some potential ordering problems

stefanotorresi:
fixes #389 and #354 
it also make the ansible linting and python checking in parallel with the go testing, which should speed up things considerably.

### DIFF
--- a/requirements.dev.txt
+++ b/requirements.dev.txt
@@ -1,0 +1,5 @@
+ansible==4.6.0
+ansible-core==2.11.5
+ansible-lint==5.1.3
+ara==1.5.7
+PyYAML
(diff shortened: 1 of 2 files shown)


fabriziosestito (approved):
Thanks for this!

requirements.dev.txt
```diff
@@ -0,0 +1,5 @@
+ansible==4.6.0
```
arbulu89, line 1: Good idea!
Most probably, we could even have a `prod` requirements file, that is used in runner container creation.

arbulu89 (approved):
Great!

requirements.dev.txt
stefanotorresi: yeah, I thought about this! Probably in another PR though.

PR #411, opened 2021-10-29
Update Swagger usage

stefanotorresi:
- removes `make swag` in favour of `go generate`.
- moves the generated files in `docs/api`
- moves the generic annotations from `web/app.go` to `web/api.go`
- moves the documentation endpoint in `/api/docs`
- fixes the declared `basePath`
- adds some docs about the API and its generated documentation
- removes the `make generate` task from the dependencies of `make test` and `make vet-check`, as this would require to install swag in the CI, which we actually don't want. Further work on the make file will be done in a following PR.


web/app.go
```diff
@@ -174,10 +173,10 @@ func NewAppWithDeps(config *Config, deps Dependencies) (*App, error) {
 	webEngine.GET("/sapsystems/:id", NewSAPResourceHandler(deps.hostsService, deps.sapSystemsService))
 	webEngine.GET("/databases", NewHanaDatabaseListHandler(deps.consul, deps.hostsService, deps.sapSystemsService, deps.tagsService))
 	webEngine.GET("/databases/:id", NewSAPResourceHandler(deps.hostsService, deps.sapSystemsService))
-	webEngine.GET("/swagger/*any", ginSwagger.WrapHandler(swaggerFiles.Handler))
 
 	apiGroup := webEngine.Group("/api")
 	{
+		webEngine.GET("/docs/*any", ginSwagger.WrapHandler(swaggerFiles.Handler))
 		apiGroup.GET("/ping", ApiPingHandler)
```
fabriziosestito, line 167: even this is moved in the block, the route is still under `webEngine`. Was this the intention or we should move it under `apiGroup`?

web/api.go
```diff
@@ -6,6 +6,21 @@ import (
 	"github.com/gin-gonic/gin"
 )
 
+//go:generate swag init -g api.go -o ../docs/api
+// @title Trento API
```
fabriziosestito, line 10: naiz!

web/app.go
stefanotorresi: nope, that's an error, thanks for catching it :D

web/api.go
stefanotorresi: IKR? 😉

stefanotorresi:
ping @dottorblaster

dottorblaster (approved):

PR #413, opened 2021-10-29
Makefile updates

stefanotorresi:
Follow up to #411 

As discussed on Slack, this PR adds a `full-check` make phony target, which also executes the `generate` one, which is no longer a dependency of `test` and `vet-check`.

It also spreads the various PHONY target declarations near the target definition itself, which is probably more convenient than having them all grouped up in one line.

It also removes the recently introduced, frontend-related, intermediate targets `web/frontend/format`, `web/frontend/format-check` and `web/frontend/lint`, which weren't declared as PHONY but they weren't true files either, so that now only the actual phony targets remain now (respectively `web-format`, `web-format-check` and `web-lint`).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ build: trento
 trento: web-assets
 	$(GO_BUILD)
 
-.PHONY: cross-compiled
+.PHONY: cross-compiled $(ARCHS)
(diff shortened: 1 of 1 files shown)


dottorblaster (approved):
LGTM :medal_military:

PR #414, opened 2021-11-02
Prune old events

fabriziosestito:
This PR adds a web subcommand to prune events older than X days (default to `10`)

It also adds a cronjob to the helm chart to execute the command once a day.


cmd/web/web.go
```diff
@@ -63,8 +74,20 @@ func NewWebCmd() *cobra.Command {
...
+func addPruneCmd(webCmd *cobra.Command) {
+	var olderThan int
+
+	pruneCmd := &cobra.Command{
+		Use:   "prune-events",
+		Short: "Prune events older than",
+		Run:   prune,
+	}
+
+	pruneCmd.Flags().IntVar(&olderThan, "older-than", 10, "Prune events older than (in days, required)")
```
nelsonkopliku, line 88: This should not be required anymore

cmd/web/web.go
```diff
@@ -80,6 +103,24 @@ func serve(cmd *cobra.Command, args []string) {
...
 	}()
 
+	app := initApp()
+	if err := app.Start(ctx); err != nil {
+		log.Fatal("Error while running the web application server: ", err)
+	}
+}
+
+func prune(cmd *cobra.Command, args []string) {
+	olderThan := viper.GetDuration("prune-events")
```
nelsonkopliku, line 113: I think we can user `.GetInt("older-than")`

nelsonkopliku (changes requested):
Super! There's only a couple of things to fix. :wink:

cmd/web/web.go
```diff
@@ -80,6 +103,24 @@ func serve(cmd *cobra.Command, args []string) {
...
+
+	app := initApp()
```
arbulu89, line 115: I'm not sure if we should have a common `initApp` for this, as the variables checked in the `loadConfig` function don't even belong to this subcommand execution. In fact, this pruning shouldn't need any gin-gonic initialization code, which we need, as we create the app with these dependencies.
Anyway, this I guess is only if we want to be 100% purist.

cmd/web/web.go
arbulu89: I would actually be more precise on this. Something like:
`Prune auto discovered events (older than x days)`
Yes, it is difficult to define this flag definition hehe

cmd/web/web.go
arbulu89: Should we do some sanity check here? 
What happens if the provided value is a negative value or similar? Maybe `viper` already has these kind of checks

arbulu89 (commented):
Hey @fabriziosestito ,
My only comment comes with the attachment of the new `prune` function on the `app` struct. I see that in order to reuse as many as already existing code, we have coupled some "not really wanted things" (like initializing the server endpoint, for example).

PD: In fact, now I have found out that the `app` is still using viper code. At some point we should initialize the db data from the command line too

cmd/web/web.go
fabriziosestito: Let's follow up, we need to explore a bit the initialization in general

nelsonkopliku (approved):
:rocket:

cmd/web/web.go
stefanotorresi: Couldn't we instantiate only the service here? I don't see the reason why pruning events a method of `web.App`, when we only need it in a CLI command.

fabriziosestito: @stefanotorresi thanks, i agree.
I will work on moving away the db initialization in the internal package so we can depend on it on both app and the command.

I also think we should move this out from the service, because the service also needs a channel to be initialized, something like an utility function in the datapipeline package will be enough.

fabriziosestito: @stefanotorresi @arbulu89 refactored the db initialization and removed the prune method from the app

cmd/web/web.go
```diff
@@ -63,11 +77,23 @@ func NewWebCmd() *cobra.Command {
...
+func addPruneCmd(webCmd *cobra.Command) {
+	var olderThan uint
+
+	pruneCmd := &cobra.Command{
+		Use:   "prune-events",
+		Short: "Prune events older than",
+		Run:   prune,
+	}
+
+	pruneCmd.Flags().UintVar(&olderThan, "older-than", 10, "Prune data discoveryu events older than <value> days.")
```
arbulu89, line 91: Typo in `discoveryu`

cmd/web/web.go
```diff
@@ -63,11 +77,23 @@ func NewWebCmd() *cobra.Command {
...
+
+	webCmd.AddCommand(pruneCmd)
 }
 
-func serve(cmd *cobra.Command, args []string) {
+func serve(_ *cobra.Command, _ []string) {
```
arbulu89, line 96: I don't know why golang doesn't complain when the function variables are not used hehe

cmd/web/web.go
```diff
@@ -90,8 +116,23 @@ func serve(cmd *cobra.Command, args []string) {
...
+	db, err := db.InitDB(dbConfig)
+	if err != nil {
+		log.Fatal("Error while initializing the database: ", err)
+	}
+
+	log.Infof("Pruning events older than %d days.", olderThan)
+	if err := datapipeline.PruneEvents(olderThan, db); err != nil {
+		log.Fatalf("Error while pruning older events: %s", err)
+	}
+	log.New().Infof("Events older than %d days pruned.", olderThan)
```
arbulu89, line 137: Why do we have here this `log.New().Infof` instead of `log.Infof`?
Is there any difference?

arbulu89 (approved):
Super @fabriziosestito ,
Now even the cmd options are totally decoupled from the web!

I have noted some typos and small questions, but the PR is ready to go from my side

cmd/web/web.go
fabriziosestito: it does complain at linting level, but using the blackhole operator makes the linter happy (and also it is easier to understand that we are not using them)

cmd/web/web.go
fabriziosestito: mmh i think it was copilot being to smart, good catch

cmd/web/web.go
fabriziosestito: (also to say we should drop a linter in when we have some time :D :D)

arbulu89: Yes, it is much better than leaving the not used parameter. It was just an internal thought hehe

nelsonkopliku (approved):

Issue #412, opened 2021-10-29
Extract the web api code to an independent package


arbulu89:
Right now, the `web` api code belongs to the `web` package. In order to have a more structured code, we could extract this logic to an additional package, which would have its own `json` models, etc

This is not an easy work, as the golang package structure is not an easy thing, Specially, because you need to avoid the `cyclical import` inferno, which is a pretty normal thing if you don't abstract all of your models and entities properly.

PR #415, opened 2021-11-02
Store checks results on the db instead of ARA

arbulu89:
Store the checks results content in the database instead of ARA.

The major changes:
- Add a new checks results table in the database: `CheckResultsRaw` (`check_results` is the actual name on the database). This table contains the checks `GroupID` (which by now is always the cluster id) and the results json as payload. It contains the creation time and an unique ID as well.
- Split the checks catalog and checks results models in 2 files. It makes the distinction easier
- Create a new API endpoint to post the checks results. In our case from the runner to the web. This is a post operation as it creates new entries every time
- Replace all the code related to ARA. Now these records are in the database.
- Update the runner callback code to post the results in this new API endpoint rather than in ARA. The content is exactly the same though

There are a lot of ARA leftovers in the code yet. I will remove them definitely once this PR is done, as this pieces of code are the ones that really use ARA by now.

PD: I still used the `JSON` prefix to define the API structs. We already have a follow up ticket at: https://github.com/trento-project/trento/issues/412


web/checks_api.go
```diff
@@ -28,24 +27,76 @@ type JSONCheck struct {
...
+// @Summary Create a checks results entry
+// @Produce json
+// @Param id path string true "Resource Id"
+// @Param Body body JSONChecksResults true "Checks results"
+// @Success 201 {object} JSONChecksResults
 // @Failure 500 {object} map[string]string
-// @Router /api/clusters/{cluster_id}/results [get]
-func ApiClusterCheckResultsHandler(client consul.Client, s services.ChecksService) gin.HandlerFunc {
+// @Router /api/checks/{id}/results [post]
+func ApiCreateChecksResultstaHandler(s services.ChecksService) gin.HandlerFunc {
```
rtorrero, line 77: The name of the function seems to have a typo: `ApiCreateChecksResultstaHandler` -> `ApiCreateChecksResultsHandler`

web/checks_api_test.go
```diff
@@ -120,11 +120,11 @@ func TestApiClusterCheckResultsHandler(t *testing.T) {
 	assert.Equal(t, 200, resp.Code)
 }
 
-func TestApiClusterCheckResultsHandler500(t *testing.T) {
+func TestApiGetCheckResultsHandler404(t *testing.T) {
 	mockChecksService := new(services.MockChecksService)
 	mockChecksService.On(
-		"GetChecksResultAndMetadataByCluster", "47d1190ffb4f781974c8356d7f863b03").Return(
-		&models.ClusterCheckResults{}, fmt.Errorf("kaboom"))
+		"GetChecksResultAndMetadataById", "47d1190ffb4f781974c8356d7f863b03").Return(
```
rtorrero, line 126: I've noticed that sometimes you use "CheckResults" and others you move the plural: "ChecksResult". Is this intentional?

web/checks_api_test.go
```diff
@@ -137,12 +137,119 @@ func TestApiClusterCheckResultsHandler500(t *testing.T) {
...
 	resp := httptest.NewRecorder()
+	req := httptest.NewRequest("GET", "/api/checks/47d1190ffb4f781974c8356d7f863b03/results", nil)
 
-	req := httptest.NewRequest("GET", "/api/clusters/47d1190ffb4f781974c8356d7f863b03/results", nil)
+	app.webEngine.ServeHTTP(resp, req)
+
+	assert.Equal(t, 404, resp.Code)
+}
+
+func TestApiCreateChecksResultstaHandler(t *testing.T) {
```
rtorrero, line 147: Also here ofc :-)

web/checks_api_test.go
rtorrero: Also sometimes I see "ChecksResults". There might be a reason I'm missing but if not, I'd try to keep it constant across the code.

web/services/checks.go
```diff
@@ -38,11 +35,11 @@ type ChecksService interface {
 	CreateChecksCatalogEntry(check *models.Check) error
 	CreateChecksCatalog(checkList models.ChecksCatalog) error
 	// Check result services
-	GetChecksResult() (map[string]*models.Results, error)
-	GetChecksResultByCluster(clusterId string) (*models.Results, error)
-	GetChecksResultAndMetadataByCluster(clusterId string) (*models.ClusterCheckResults, error)
-	GetAggregatedChecksResultByHost(clusterId string) (map[string]*AggregatedCheckData, error)
-	GetAggregatedChecksResultByCluster(clusterId string) (*AggregatedCheckData, error)
+	GetChecksResultById(id string) (*models.Results, error)
```
rtorrero, line 38: Basically the same as already commented

rtorrero (approved):
Thanks for this! LGTM except minor cosmetical issues. Feel free to sort them out in a different PR if you agree with them (or here if you prefer)

stefanotorresi:
To unblock this, for the time being, let's use the public api Gin instance, which was a quandary left unsolved since the block.

arbulu89:
Closing this as the conflict resolution looks more difficult than opening a new fresh PR.
I will leverage most of the content, but based in the new upstream code

PR #416, opened 2021-11-02
Use the correct test request constructor

stefanotorresi:
As mentioned in Slack, `http.NewRequest()` is the client request constructor and it doesn't trigger the parsing of a lot of internal fields which some server handlers might rely on, so they would fail tests that should actually be passing.

I spotted such a false negative when testing the swagger handler, which does this kind of things:
https://github.com/swaggo/gin-swagger/blob/5ef0958c325e56d022effff2081e02430e573856/swagger.go#L82
Step debugging showed `c.Request.RequestURI` was empty during the test run, but the handler was actually working correctly when running the application in a real environment.

The correct method to create simulated server requests is `httptest.NewRequest()` as documented here:
https://pkg.go.dev/net/http/httptest#NewRequest

### DIFF
--- a/web/about_test.go
+++ b/web/about_test.go
@@ -30,9 +30,6 @@ func TestAboutHandlerPremium(t *testing.T) {
 
 	resp := httptest.NewRecorder()
 	req := httptest.NewRequest("GET", "/about", nil)
-	if err != nil {
-		t.Fatal(err)
-	}
 
 	app.webEngine.ServeHTTP(resp, req)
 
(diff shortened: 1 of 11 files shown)


web/about_test.go
```diff
@@ -30,7 +29,7 @@ func TestAboutHandlerPremium(t *testing.T) {
 	}
 
 	resp := httptest.NewRecorder()
-	req, err := http.NewRequest("GET", "/about", nil)
+	req := httptest.NewRequest("GET", "/about", nil)
```
arbulu89, line 32: @stefanotorresi I think that the brute `copy/paste` is not a good idea hehe
Here for example, we would need to remove the next `if err != nil` condition. That it doesn't break any test is a coincidence, as we create the app before getting the `err` variable

arbulu89 (changes requested):
@stefanotorresi As commented in the inline comment, we need a bit more of `copy/paste` magic hehe

web/about_test.go
stefanotorresi: gahh, you're right. I couldn't do that with regexes 😁

arbulu89 (approved):
Go go!

PR #422, opened 2021-11-04
Switch clusters page on projected read models

nelsonkopliku:
With this PR we are switching the `/clusters` view to use the projected read models.


arbulu89 (commented):
Fantastic!
I hope nobody disables the data collection, otherwise this won't work.
Maybe it is a good moment to remove that option as well.

As a small comment, I think that the `ClustersTable` and `ClustersRow` structs (and all the related functions that now belong to the cluster service and model) are not used anymore, so they could be removed.

fabriziosestito (commented):
Great! Apart from the comments above, I think we still miss the agent configuration in the install agent script. This works fine in our CI because we are using environment variables, but when the user will try to install this it is not going to work out of the box.

Let's follow up on this, it might be time to use a generated config file and to tweak viper/the install-agent script a little.

nelsonkopliku:
> Fantastic! I hope nobody disables the data collection, otherwise this won't work. Maybe it is a good moment to remove that option as well.
> 
> As a small comment, I think that the `ClustersTable` and `ClustersRow` structs (and all the related functions that now belong to the cluster service and model) are not used anymore, so they could be removed.
>
Thanks @arbulu89 cleaned up the `ClustersTable` and `ClustersRow` and removed reference to the `collector-enabled` flag, which is enabled by default now.

---

> Great! Apart from the comments above, I think we still miss the agent configuration in the install agent script. This works fine in our CI because we are using environment variables, but when the user will try to install this it is not going to work out of the box.
> 
> Let's follow up on this, it might be time to use a generated config file and to tweak viper/the install-agent script a little.
>
@fabriziosestito as always, correct.
The thing you say is that our `packaging/systemd/trento-agent.service` which gets installed in the node
runs 
```
[Service]
ExecStart=/usr/bin/trento agent start --consul-config-dir=/srv/consul/consul.d
```
which clearly is not considering the collector host, neither th mtls enabled nor the ceerificates :trollface: 

Big yes for the config file :rocket: 

Here's a `.trento-agent.yaml` that I use locally

```
enable-mtls: true
cert: ./test/certs/server-cert.pem
key: ./test/certs/server-key.pem
ca: ./test/certs/ca-cert.pem


collector-host: localhost
collector-port: 8443
```

which I use like this `./trento agent start --config=/path/to/.trento-agent.yaml`

I am not saying it has to be a `yaml` or that the filename should be that one, it's just an example.
Let's find out what works best for the users.

fabriziosestito:
> > Fantastic! I hope nobody disables the data collection, otherwise this won't work. Maybe it is a good moment to remove that option as well.
> > As a small comment, I think that the `ClustersTable` and `ClustersRow` structs (and all the related functions that now belong to the cluster service and model) are not used anymore, so they could be removed.
> 
> Thanks @arbulu89 cleaned up the `ClustersTable` and `ClustersRow` and removed reference to the `collector-enabled` flag, which is enabled by default now.
> 
> > Great! Apart from the comments above, I think we still miss the agent configuration in the install agent script. This works fine in our CI because we are using environment variables, but when the user will try to install this it is not going to work out of the box.
> > Let's follow up on this, it might be time to use a generated config file and to tweak viper/the install-agent script a little.
> 
> @fabriziosestito as always, correct. The thing you say is that our `packaging/systemd/trento-agent.service` which gets installed in the node runs
> 
> ```
> [Service]
> ExecStart=/usr/bin/trento agent start --consul-config-dir=/srv/consul/consul.d
> ```
> 
> which clearly is not considering the collector host, neither th mtls enabled nor the ceerificates :trollface:
> 
> Big yes for the config file rocket
> 
> Here's a `.trento-agent.yaml` that I use locally
> 
> ```
> enable-mtls: true
> cert: ./test/certs/server-cert.pem
> key: ./test/certs/server-key.pem
> ca: ./test/certs/ca-cert.pem
> 
> 
> collector-host: localhost
> collector-port: 8443
> ```
> 
> which I use like this `./trento agent start --config=/path/to/.trento-agent.yaml`
> 
> I am not saying it has to be a `yaml` or that the filename should be that one, it's just an example. Let's find out what works best for the users.
> 
> Followup for config #423

@nelsonkopliku i am a big fan of toml for this but we are pretty much yaml driven in the whole codebase, so i think we should go with that (sigh).
As a simple improvement we could create an example configuration in a `/etc/trento` folder (or whatever is idiomatic for SUSE) directly from the rpm, something like `/etc/trento/config.yaml`, with commented out lines that can act as documentation for the user.

Then the installer script should be able to generate a `config.yaml` by using the parameters (in this case, only the trento server IP/host would be enough).

Then we could alter the systemd file unit to read the configuration and let viper do the magic for us.


We could also use something like a `conf.d` path where several configuration that could be loaded with priorities, if this is something that can help the installation process.

arbulu89 (approved):
Nice! Now I don't see any leftover. We can remove them if we find more.
Thanks

rtorrero:
I'm loving this. Ping me if you need some help on the RPM & systemd service file changes if you want :-)

dottorblaster (approved):
HELL YEAH

fabriziosestito (approved):

PR #424, opened 2021-11-05
Using correct mocked value for discovered cloud

nelsonkopliku:
Fixing overlooked transformation that is happening in the agent side, related to Cloud discovery.

What is being published by the agent is `azure` instead of `some-weird-id`


arbulu89 (approved):
Good catch!

PR #428, opened 2021-11-05
Use a different package to extract system information during host discovery

nelsonkopliku:
On demo we are projecting the `HostTelemetry`, but the current package is not able to properly extract CPU count.

This PR replaces it with https://github.com/shirou/gopsutil.

I have doubts on the physical/logical cpu stuff. Please let know if that would work

### DIFF
--- a/agent/discovery/host.go
+++ b/agent/discovery/host.go
@@ -3,14 +3,17 @@ package discovery
 import (
 	"fmt"
 	"net"
+	"strconv"
 	"strings"
 
+	"github.com/shirou/gopsutil/cpu"
+	"github.com/shirou/gopsutil/host"
+	"github.com/shirou/gopsutil/mem"
 	log "github.com/sirupsen/logrus"
 	"github.com/trento-project/trento/agent/discovery/collector"
 	"github.com/trento-project/trento/internal/consul"
 	"github.com/trento-project/trento/internal/hosts"
 	"github.com/trento-project/trento/version"
-	"github.com/zcalusic/sysinfo"
 )
 
 const HostDiscoveryId string = "host_discovery"
@@ -46,16 +49,13 @@ func (h HostDiscovery) Discover() (string, error) {
 		return "", err
 	}
 
-	var si sysinfo.SysInfo
-	si.GetSysInfo()
-
 	host := hosts.DiscoveredHost{
-		OSVersion:       si.OS.Version,
+		OSVersion:       getOSVersion(),
 		HostIpAddresses: ipAddresses,
 		HostName:        h.discovery.host,
-		CPUCount:        int(si.CPU.Cpus) * int(si.CPU.Cores),
-		SocketCount:     int(si.CPU.Cpus),
-		TotalMemoryMB:   int(si.Memory.Size),
+		CPUCount:        getLogicalCPUs(),
+		SocketCount:     getCPUSocketCount(),
+		TotalMemoryMB:   getTotalMemoryMB(),
 		AgentVersion:    version.Version,
 	}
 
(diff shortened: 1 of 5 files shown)


agent/discovery/host.go
```diff
@@ -90,3 +89,35 @@ func getHostIpAddresses() ([]string, error) {
 
 	return ipAddrList, nil
 }
+
+func GetOSVersion() string {
+	infoStat, err := host.Info()
+	if err != nil {
+		log.Errorf("Error while getting host info: %s", err)
+	}
+	return infoStat.PlatformVersion
```
dottorblaster, line 98: What if we get an error from `host.Info()`? We'll still have `PlatformVersion` inside `infoStat`?

agent/discovery/host.go
```diff
@@ -90,3 +89,35 @@ func getHostIpAddresses() ([]string, error) {
...
+}
+
+func getTotalMemoryMB() int {
+	v, err := mem.VirtualMemory()
+	if err != nil {
+		log.Errorf("Error while getting memory info: %s", err)
+	}
+	return int(v.Total) / 1024 / 1024
```
dottorblaster, line 106: Same here, we got an empty struct `if err != nil`? I really don't know what we get

agent/discovery/host.go
nelsonkopliku: It will get its null value, in this case an empty string. I guess :sweat_smile:

agent/discovery/host.go
nelsonkopliku: It will get the int zero value `0`.

dottorblaster (approved):
LGTM

arbulu89 (commented):
I think we should handle the error on these commands, to set the discovery in a failed/warning state.
We shouldn't report at least a `0` value. Maybe `-1`?
Or not report the value is it failed.

agent/discovery/host.go
```diff
@@ -90,3 +89,35 @@ func getHostIpAddresses() ([]string, error) {
...
+
+func GetOSVersion() string {
+	infoStat, err := host.Info()
+	if err != nil {
+		log.Errorf("Error while getting host info: %s", err)
+	}
+	return infoStat.PlatformVersion
+}
+
+func getTotalMemoryMB() int {
```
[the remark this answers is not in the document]
nelsonkopliku: No particular reason. No one is using it, except for this discovery.

When neede It could very well be public, or even an Interface to a service.
I kept it really simple.

Is that good enough for now?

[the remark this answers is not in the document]
arbulu89: I was just surprised because the other `Get...` function are using capital letter. I just noticed the difference and lack of consistency. Nothing else hehe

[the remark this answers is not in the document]
nelsonkopliku: Ah, my bad, now I spotted.

They were supposed to be all private actually. Now I see I have all public except for one.
A leftover. Gonna make all private.

Thanks :pray:

nelsonkopliku:
> I think we should handle the error on these commands, to set the discovery in a failed/warning state. We shouldn't report at least a `0` value. Maybe `-1`? Or not report the value is it failed.
>
Thanks @arbulu89. I do agree with you.
Since this telemetry thing is still super alpha, is that okay to postpone that? I'll track the thing.

agent/discovery/host.go
```diff
@@ -46,16 +48,13 @@ func (h HostDiscovery) Discover() (string, error) {
...
 	host := hosts.DiscoveredHost{
-		OSVersion:       si.OS.Version,
+		OSVersion:       getOSVersion(),
 		HostIpAddresses: ipAddresses,
 		HostName:        h.discovery.host,
-		CPUCount:        int(si.CPU.Cpus) * int(si.CPU.Cores),
-		SocketCount:     int(si.CPU.Cpus),
-		TotalMemoryMB:   int(si.Memory.Size),
+		CPUCount:        getLogicalCPUs(),
+		SocketCount:     getPhysicalCPUs(),
```
rtorrero, line 56: I wouldn't match SocketCount with the Core count. In my machine, which just has one socket (as most desktops), it returns `12` which is actually the number of cores.

Appart, from that (I know this was already there before the PR), I'd take the opportunity to rename following what other common tools do. See e.g. `lscpu` output for my machine:

```
CPU(s):                  24
  On-line CPU(s) list:   0-23
Vendor ID:               AuthenticAMD
  Model name:            AMD Ryzen 9 3900X 12-Core Processor
    CPU family:          23
    Model:               113
    Thread(s) per core:  2
    Core(s) per socket:  12
    Socket(s):           1
    Stepping:            0
    Frequency boost:     enabled
    CPU max MHz:         4672.0698
    CPU min MHz:         2200.0000
    BogoMIPS:            7586.00
```
(Not saying that we need to add all that, just that for the aspects of the CPU that we are adding, we should use standard names)

Also, I'm a fan of adding `Count` as suffix when a var reflects a sum of of something, but that's just a suggestion :-)

rtorrero:
After a bit of thought, I think this was the right decision. Tracking core counts of VMs also makes sense, and in the future we could add checks to report if the agent is running behind a hypervisor.

Good job! (just left a comment at some small detail I think we should change before merging)

agent/discovery/host.go
dottorblaster: Aside:

`AMD Ryzen 9 3900X 12-Core Processor` :OOOOOOO

rtorrero:
It looks good to me now, thanks :-)!

PR #429, opened 2021-11-05
Forcing refreshing updated_at information of the HostTelemetry

nelsonkopliku:
In order to know when a projected read model has been reprojected, even if no changes applied, we force to always update also the `updated_at` property.


dottorblaster (approved):
We said we had to make smaller PRs, and you absolutely nailed it :joy:

Issue #425, opened 2021-11-05
Make agent id a uuid based on machine-id


nelsonkopliku:
Now we're identifying the agents with `/etc/machine-id`.

We'd like to uniform identifiers to `uuids` and keep uniqueness of the AgentID

https://github.com/google/uuid

PR #432, opened 2021-11-08
Use Agent UUID from machine id

nelsonkopliku:
With this PR we create anAgentID which is a UUID ceated from the content in `/etc/machine-id`.

The created UUID will be the same for the same machineID content given.

We have uuid and uniqueness of agent identifier.

Closes https://github.com/trento-project/trento/issues/425

### DIFF
--- a/agent/discovery/collector/client.go
+++ b/agent/discovery/collector/client.go
@@ -10,7 +10,9 @@ import (
 	"net/http"
 	"strings"
 
+	"github.com/google/uuid"
 	log "github.com/sirupsen/logrus"
+	"github.com/trento-project/trento/internal"
 
 	"github.com/spf13/afero"
 	"github.com/spf13/viper"
@@ -22,7 +24,7 @@ type Client interface {
 
 type client struct {
 	config     *Config
-	machineID  string
+	agentID    string
 	httpClient *http.Client
 }
 
@@ -64,10 +66,12 @@ func NewCollectorClient(config *Config) (*client, error) {
 
 	machineID := strings.TrimSpace(string(machineIDBytes))
 
+	agentID := uuid.NewSHA1(internal.TrentoNamespace, []byte(machineID))
+
 	return &client{
 		config:     config,
 		httpClient: httpClient,
-		machineID:  machineID,
+		agentID:    agentID.String(),
 	}, nil
 }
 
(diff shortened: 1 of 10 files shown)


dottorblaster (approved):

agent/discovery/collector/client.go
```diff
@@ -64,10 +65,15 @@ func NewCollectorClient(config *Config) (*client, error) {
 
 	machineID := strings.TrimSpace(string(machineIDBytes))
 
+	agentID, err := uuid.NewRandomFromReader(strings.NewReader(machineID))
+	if err != nil {
```
fabriziosestito, line 69: Is the UUID Random?
In this case what happens if the agent or the machine is restarted?

We need to find a way to have deterministic IDs, e.g. namespacing could help

fabriziosestito (changes requested):

agent/discovery/collector/client.go
nelsonkopliku: From what I could see, this line returns always the same UUID given the same machineID.

I can double check.

rtorrero: I think this is a seed for a random number so yes, given the same seed the random number should be the same.

nelsonkopliku: I'll add a test for this.

```
Pseudocode

loop n times and call
uuid.NewRandomFromReader(strings.NewReader(machineID))

check all the uuids are the same
```

Would this help?

fabriziosestito: I would still use a namespace for this one rather than the random  uuid

nelsonkopliku: I;ll give it a try :wink:

agent/discovery/collector/client.go
```diff
@@ -35,6 +36,8 @@ type Config struct {
 	CA            string
 }
 
+var UuidNamespace = uuid.Must(uuid.Parse("fb92284e-aa5e-47f6-a883-bf9469e7a0dc"))
```
nelsonkopliku, line 39: I do not really like it here, I see this at a higher level in our modules, but I am getting bitten by cyclic imports.
Gonna keep it here for now if that's okay. :innocent:

agent/discovery/collector/client.go
nelsonkopliku: @fabriziosestito changed, please take a look.

agent/discovery/collector/client.go
fabriziosestito: does moving this to `internal/uuid.go` helps? 
I would call it `TrentoNamespace` or something similar to  make it clear that is our defined namespace

fabriziosestito (approved):
Little nitpick on naming, but feel free to merge and address it later

agent/discovery/collector/client.go
nelsonkopliku: Deal.

PR #433, opened 2021-11-08
Add hosts projector

fabriziosestito:
This PR introduces the first iteration of the hosts projector and service.

It also introduce the concept of model/entity separation, where entity is the object we use to store data to the persistence layer, therefore is shaped around the gorm library, while the model is the object we use in our domain code and views. 
Also the hosts service now leverages gorm to load associations such as Tags, and it can be used in the future to load other normalized read models such as SAPSystems, health checks, etc..

As a side note
Inspiration for services design was taken from here:
https://hexdocs.pm/phoenix/contexts.html#cross-context-data

What it is still mixing and will come as followup PR:
- [ ] Add heartbeat system so we can have health of a host
- [ ] Add errors to discovered data event so we can show if the agent is working correctly
- [ ] When SAPSystems will be projected, add associations between hosts and systems so we can build the links in the hostlist table
- [ ] Refactor clusters to use entity/model separation as well, and gorm for preloading tags and filtering
- [ ] Move Tag in entities package, refactor the tagging system by moving the related create/delete/list tags to every "taggable" service (this will save us a lot of code by using just one handler instead of one specific handler per resource)


web/datapipeline/projector.go
```diff
@@ -74,3 +78,11 @@ func (p *projector) Project(dataCollectedEvent *DataCollectedEvent) error {
 		return nil
 	})
 }
+
+func payloadDecoder(payload datatypes.JSON) *json.Decoder {
```
dottorblaster, line 82: what do you think about renaming this to `getPayloadDecoder`?

web/hosts_next.go
```diff
@@ -0,0 +1,65 @@
...
+	// for _, host := range hostList {
+	// 	switch host.Health() {
+	// 	case consulApi.HealthPassing:
+	// 		h.PassingCount += 1
+	// 	case consulApi.HealthWarning:
+	// 		h.WarningCount += 1
+	// 	case consulApi.HealthCritical:
+	// 		h.CriticalCount += 1
+	// 	}
+	// }
```
dottorblaster, line 22: Leftover?

web/hosts_next.go
```diff
@@ -0,0 +1,65 @@
...
+
+		filterTags, err := hostsService.GetAllTags()
+		if err != nil {
+			_ = c.Error(err)
+			return
+		}
+		// hostList = filterHostsByTags(hostList, hostsTags, tagsFilter)
+
+		// hContainer := NewHostsHealthContainer(hostList)
+		// hContainer.Layout = "horizontal"
```
dottorblaster, line 50: Leftover?

dottorblaster (commented):
Some comments but I feel like we're ready to merge

web/datapipeline/projector.go
nelsonkopliku: good for me if it makes it clearer :wink:

web/entities/host.go
```diff
@@ -0,0 +1,41 @@
...
+	CloudProvider string
+	ClusterID     string
+	ClusterName   string
+	SIDs          pq.StringArray `gorm:"column:sids; type:text[]"`
+	AgentVersion  string
+	Tags          []models.Tag `gorm:"polymorphic:Resource;polymorphicValue:hosts"`
+	UpdatedAt     time.Time
+}
+
+func (h *Host) ToModel() *models.Host {
```
nelsonkopliku, line 23: Oh man, this feels so gooooood. :heart:

nelsonkopliku (approved):
Good good stuff.

I do not see anything strictly blocking, anyway it's not going to break current behavior, and we're going to iterate on this up next if I understood correctly.

web/app.go
```diff
@@ -143,6 +148,7 @@ func NewAppWithDeps(config *Config, deps Dependencies) (*App, error) {
 	webEngine.GET("/", HomeHandler)
 	webEngine.GET("/about", NewAboutHandler(deps.subscriptionsService))
 	webEngine.GET("/hosts", NewHostListHandler(deps.consul, deps.tagsService))
+	webEngine.GET("/hosts-next", NewHostListNextHandler(deps.hostsNextService))
```
stefanotorresi, line 151: Have you considered attaching always to the same route, but changing handler depending on a quickly baked feature flag (e.g. an environment variable or an internal viper setting)?

fabriziosestito: @stefanotorresi we could but I need both routes for development and also this is gonna go away pretty soon...

nelsonkopliku (approved):

web/app.go
stefanotorresi: All righty then, carry on!

rtorrero (approved):
This keeps getting better, lgtm!

PR #434, opened 2021-11-08
Allow Agent rpm to run with config file

nelsonkopliku:
Okay here we go.

This PR introduces usage of configuration files.
FTR: trento was already able to do so thanks to [Viper](https://github.com/spf13/viper) :snake: , now we're starting to leverage it.

What's in here:
- example `agent.yaml` (see following section `On single vs multiple config files`)
- RPM update to bundle this new file
- trento is able to load different config for different contexts (`agent`, `web`, `runner`, ...)
- `install-server.sh` ~~supports a `--generate-config` flag (`false` by default)~~ generates a configuration file in `/usr/etc/trento/agent.yaml`
- pipeline cleanup
- some documentation

<details>
  <summary><b>On single vs multiple config files</b></summary>
     
in relation to having one single config file with namespace here's a thought.
Such a file would be something like:

```
# /usr/etc/trento/trento.yaml

web:
    db-port: 5432
    ...

agent:
    enable-mtls: true
    cert: ./path/to/certs/client-cert.pem
    key: ./path/to/certs/client-key.pem
    ca: ./path/to/certs/ca-cert.pem
    collector-host: localhost
    collector-port: 8443

runner:
    ...
```

If we want to also keep supporting cli flags we're required to:
detect whether a configuration file has been providedor not
access options either by `viper.GetString("collector-host")`  or `viper.GetString("agent.collector-host")`  given the previous condition
If, otherwise, we have different config files they would look like this:

```
# /usr/etc/trento/trento-agent.yaml

enable-mtls: true
cert: ./path/to/certs/client-cert.pem
key: ./path/to/certs/client-key.pem
ca: ./path/to/certs/ca-cert.pem
collector-host: localhost
collector-port: 8443
```
with this, we transparently access to `viper.GetString("collector-host")`  and we do not really care where it comes from, cli or config.
We'd also get clear boundaries.
</details>

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ of existing clusters, rather than deploying new one.
       - [Setting up and starting ARA](#setting-up-and-starting-ara)
       - [Starting the Trento Runner](#starting-the-trento-runner)
     + [Trento Web UI](#trento-web-ui)
+- [Configuration](#configuration)
 - [Development](#development)
   * [Helm development chart](#helm-development-chart)
   * [Build system](#build-system)
@@ -444,6 +445,89 @@ At this point, we can start the web application as follows:
 ```
 
 Please consult the `help` CLI command for more insights on the various options.
+
+
+# Configuration
+
+Trento can be run with a config file in replacement of command-line arguments.
+
+## Locations
+Configuration, if not otherwise specified by the `--config=/path/to/config.yaml` option, would be searched in following locations:
+
+Note that order represents priority
+
+- `/etc/trento/` <-- first location looked
+- `/usr/etc/trento/` <-- fallback here if config not found in previous location
+- `~/.config/trento/` aka user's home <-- fallback here
+
+## Formats
+
+`yaml` is the only supported format at the moment.
+
+
+## Naming conventions
+Each component of trento supports its own configuration, so the expected config files in the chosen location must be called after the component it is supporting.
+
+So supported names are `(agent|web|runner).yaml`
+
+Example locations:
+
+`/etc/trento/agent.yaml`
+
+`/etc/trento/web.yaml`
+
+`/etc/trento/runner.yaml`
+
+or 
+
+`/usr/etc/trento/agent.yaml`
+
+`/usr/etc/trento/web.yaml`
+
+`/usr/etc/trento/runner.yaml`
+
+**Note**: `runner` still not supported for now
+
+## Examples
+
+```
+# /etc/trento/agent.yaml
+
+enable-mtls: true
+cert: /path/to/certs/client-cert.pem
+key: /path/to/certs/client-key.pem
+ca: /path/to/certs/ca-cert.pem
+collector-host: localhost
+collector-port: 8443
+```
+
+```
+# /etc/trento/web.yaml
+
+db-port: 5432
+
+collector-port: 8443
+enable-mtls: true
+cert: /path/to/certs/server-cert.pem
+key: /path/to/certs/server-key.pem
+ca: /path/to/certs/ca-cert.pem
+```
+
+## Environment Variables
+
+All of the options supported by the command line and configuration file can be provided as environment variables as well.
+
+The rule is: get the option name eg. `enable-mtls`, replace dashes `-` with underscores `_`, make it uppercase and add a `TRENTO_` prefix.
+
+Examples:
+
+`enable-mtls` -> `TRENTO_ENABLE_MTLS=true ./trento agent start`
+
+`collector-host` -> `TRENTO_COLLECTOR_HOST=localhost ./trento web serve`
+
+`cert` -> `TRENTO_CERT=/path/to/certs/server-cert.pem ./trento web serve`
+
+
 
 # Development
 
--- a/internal/config.go
+++ b/internal/config.go
@@ -1,0 +1,68 @@
+package internal
+
+import (
+	"fmt"
+	"os"
+	"path"
+	"strings"
+
+	"github.com/mitchellh/go-homedir"
+	"github.com/spf13/viper"
+)
+
+// InitConfig intializes the config for the application
+// If no config file is provided with the --config flag
+// it will look for a config in following locations:
+//
+// ${context} being one of the supported components: agent|web|runner
+//
+// Order represents priority
+// /etc/trento/${context}.yaml
+// /usr/etc/trento/${context}.yaml
+// $HOME/.config/trento/${context}.yaml
+func InitConfig(configName string) error {
+	viper.SetConfigType("yaml")
+	SetLogLevel(viper.GetString("log-level"))
+	SetLogFormatter("2006-01-02 15:04:05")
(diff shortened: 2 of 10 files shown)


rtorrero (commented):
I'm totally fine with adding the configuration file commented out, but I'd rename it to something like `agent.yaml.example`, wdyt?

Otherwise we could uncomment it and add it as defaults, and fail on first attempt with a clear:
`Please, set up the certificate and key files to use mTLS or disable it (not recommended)`

fabriziosestito:
> I'm totally fine with adding the configuration file commented out, but I'd rename it to something like `agent.yaml.example`, wdyt?
> 
> Otherwise we could uncomment it and add it as defaults, and fail on first attempt with a clear: `Please, set up the certificate and key files to use mTLS or disable it (not recommended)`

@rtorrero 
Commented out config are pretty common in `etc` folders, see for instance `/etc/ssh/sshd_config` in any Arch based distribution. I would just add comments with a brief description of the fields as a future improvement.

I wouldn't fail after the installation that would be very annoying for the user. The mTLS configuration is just an addition that we can suggest but not enforce at this level, the user could for example proxy the collector endpoint through and add its own configuration layer.

internal/config.go
```diff
@@ -0,0 +1,54 @@
...
+		// Search context config in following locations:
+		// Order represents priority
+		//
+		// /etc/trento/${context}.yaml
+		// /usr/etc/trento/${context}.yaml
+		// $HOME/${context}.yaml
+		viper.AddConfigPath("/etc/trento/")
+		viper.AddConfigPath("/usr/etc/trento/")
+		viper.AddConfigPath(home)
+		viper.SetConfigName(context)
```
dottorblaster, line 41: Could we have this comment as godoc?

nelsonkopliku: Sure, changed.

fabriziosestito (approved):

internal/config.go
```diff
@@ -0,0 +1,63 @@
...
+	viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_", ".", "_"))
+	viper.SetEnvPrefix("TRENTO")
+	viper.AutomaticEnv() // read in environment variables that match
+
+	err := viper.ReadInConfig()
+
+	if err != nil {
+		if _, notFoundError := err.(viper.ConfigFileNotFoundError); !notFoundError || shouldBreakOnConfigNotFound {
+			return err
+		}
```
stefanotorresi, line 59: I think we have some error conditions not being handled here, e.g. other types of file I/O errors that `viper.ReadInConfig()` might return.

README.md
```diff
@@ -445,6 +446,74 @@ At this point, we can start the web application as follows:
 
 Please consult the `help` CLI command for more insights on the various options.
 
+
+# Configuration
+
+To ease life of users, Trento can be run with a config file in replacement of command-line arguments.
```
stefanotorresi, line 452: We should mention here the ability to use environment variables automatically mapped to config keys.
I would also suggest to remove the sentence `To ease life of users,`, because that's a very subjective thing. 😛

README.md
```diff
@@ -445,6 +446,74 @@ At this point, we can start the web application as follows:
...
+
+Note that order represents priority
+
+- `/etc/trento/` <-- first location looked
+- `/usr/etc/trento/` <-- fallback here if config not found in previous location
+- `~` aka user's home <-- fallback here
+
+## Formats
+
+Supported formats are the ones suppoerted by [Viper](https://github.com/spf13/viper), and we encourage `yaml` usage.
```
stefanotorresi, line 465: Typo here. Also the word "supported" is repeated.

Besides that, did you actually try using some other format? I remember experiencing issues with that in the past, so we ended up locking viper to only support `yaml` and call it done.

stefanotorresi (changes requested):
This is fantastic work! Only a couple of little things.

README.md
nelsonkopliku: `To ease life of users` :trollface: 
Fixing it.

README.md
nelsonkopliku: How can I rephrase it?

`Supported formats are the ones that Viper supports`, good?

And, no, I actually didn't try other formats. Shall we communicate to only use `yaml` and remove reference to supported formats in VIper?

nelsonkopliku: Know what, gonna stick on `yaml`.

I tried out `toml` and we'd have to consider namespacing by default, for configuration groups.

```
# TOML

[mTLS]
enable-mtls = false
cert = "./test/certs/client-cert.pem"
key = "./test/certs/client-key.pem"
ca = "./test/certs/ca-cert.pem"
```

This becomes accessible by `viper.GetBool("mTLS.enable-mtls")`

In yaml it would be
```
mTLS:
    enable-mtls: false
    cert: ./test/certs/client-cert.pem
    key: ./test/certs/client-key.pem
    ca: ./test/certs/ca-cert.pem
```

and access the same way  by `viper.GetBool("mTLS.enable-mtls")`.

I do not dislike this namespacing, yet I'd keep it simple and stick on `yaml`.

Deal?

stefanotorresi: Either way works for me. Namespacing is kinda cool, but it also makes very ugly CLI flags: e.g. `--mtls.enable-mtls`. We would need to review _all of them_ and rename _a lot_ to make sure there are no funky ones.

nelsonkopliku: indeed. I wouldn't do it NAO :joy: 
So simple yaml shou;d be enough

dottorblaster (approved):
LGTM

packaging/config/agent.yaml
```diff
@@ -0,0 +1,29 @@
+# Path where Consul auto-loads configuration files: the node meta-data will be stored there as `trento-metadata.json`
```
rtorrero, line 1: Just nitpicking, but I'd add an intro in the comments describing the purpose of the file, e.g. -> "This is the configuration for trento when executed under the `agent` role ..." and describe also a bit the order of precedence. I know that this stuff is probably already described elsewhere in the docs, but it's IMO a good thing to have it here too.

Feel free to ignore this for now as we can adress it in a future PR.

packaging/config/agent.yaml
```diff
@@ -0,0 +1,29 @@
...
+# Legacy, it will be removed in the future
+# defaults to ./consul.d
+
+# consul-config-dir: /srv/consul/consul.d
```
rtorrero, line 5: More nit-picking: I'd add a double comment  `##` for actual comments and a single one for configuration values. This makes it clearer for the user to quickly identify what the configurable values are.

rtorrero (approved):
Just few minor things we can fix later (or now if you feel it's important enough)

Good stuff! :-)

packaging/config/agent.yaml
nelsonkopliku: Makes sense, I added some :wink:

packaging/config/agent.yaml
nelsonkopliku: Sure thing! Changed. Thanks for this, I might be overlooking these things, so well done.

PR #436, opened 2021-11-09
Revert 435

stefanotorresi:
Long-term alternative solution to #435.
Pending an update to the CI jobs so that we attempt to build docker images on every commit, to avoid further regressions in the make targets dependency graph.

To explicitly rerun webpack, you can now run `make clean-web-assets-js build`.
Symmetrically, if you wanted to rebuild just the go binary without re-running webpack, you could already run `make clean-binary build`.


dottorblaster (approved):
LGTM

nelsonkopliku (approved):
Goooo

PR #438, opened 2021-11-09
More ci goodies

stefanotorresi:
- optimizes `Dockerfile` so that it doesn't need to re-download dependencies for every source code change.
- more entries to `.dockerignore`
- fixes a minor issue with the `make trento` target which wouldn't properly detect when it was up-to-date already
- builds containers at every commit and run the `trento version` as a simple smoke test
- pins all jobs to the same fixed ubuntu version

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,6 @@
-web/frontend/node_modules
-web/frontend/assets
-trento
+/*.sh
+/.github
+/packaging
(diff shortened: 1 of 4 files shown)


rtorrero (approved):
Developer experience is over 9000 now :-) Thanks!

fabriziosestito (approved):

PR #442, opened 2021-11-10
Add Cypress and add a first smoke test

dottorblaster:
As we need MOAR tests and we are moving towards a stabler product and a more intensive QA phase, I thought that putting together some smoke tests could be a good idea especially when it comes to identifying regressions.

I used Cypress to do this, I like _a lot_ this tool because it's easily scriptable and it can instrument both Chromium and Firefox.


fabriziosestito (approved):

test/e2e/cypress/integration/about.js
```diff
@@ -0,0 +1,20 @@
+context('Trento About page', () => {
+  beforeEach(() => {
+    cy.visit('http://demo.trento-project.io');
+  });
```
fabriziosestito, line 4: we could make this an env variable with a default so we can use it in our local forks/envs

fabriziosestito:
Loving this!!

test/e2e/cypress/integration/about.js
dottorblaster: You're actually right, I have to make a tiny library to make the import and the setup of this kind of stuff easier

nelsonkopliku (approved):
yes-please
pg

dottorblaster:
@fabriziosestito I added the `baseUrl` configuration, Cypress supports it out of the box. Also, we can override the baseUrl using the `CYPRESS_BASE_URL` env variable. What do you think?

PR #444, opened 2021-11-11
Add heartbeat to agent

fabriziosestito:
Part 2 of the heartbeat implementation started in #441.
This PR adds a ticker loop that sends an heartbeat to the collector endpoint every `HeartbeatInterval` seconds.
(this can't be configured by the user and probably never will).

We might want to rename the collector endpoint to something like "private api" and have publish and heartbeat client methods in the `/api` folder so let's keep this on the radar, but as for now this is just enough.

### DIFF
--- a/agent/discovery/collector/client.go
+++ b/agent/discovery/collector/client.go
@@ -19,6 +19,7 @@ import (
 
 type Client interface {
 	Publish(discoveryType string, payload interface{}) error
+	Heartbeat() error
 }
 
 type client struct {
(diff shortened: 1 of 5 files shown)


dottorblaster:
image

agent/discovery/collector/client.go
```diff
@@ -19,6 +19,7 @@ import (
 
 type Client interface {
 	Publish(discoveryType string, payload interface{}) error
+	Heartbeat() error
```
nelsonkopliku, line 22: Just for reference, since I think it is relates. https://github.com/trento-project/trento/pull/441#discussion_r747884377 

As we already said, not necessary to act now

nelsonkopliku (approved):
Woah. Crazy friday.

nelsonkopliku (approved):

dottorblaster (approved):

PR #450, opened 2021-11-12
Cleanup clusters service

fabriziosestito:
Iterating on the hosts projectors we introduced the separation between persistence related structs and models.
Also we managed to use polymorphic associations for tags.
This PR refactors the clusters service with this concepts.

Next iterations will introduce variadic functional options for filtering and pagination.

### DIFF
--- a/web/datapipeline/clusters_projector.go
+++ b/web/datapipeline/clusters_projector.go
@@ -7,6 +7,7 @@ import (
 
 	log "github.com/sirupsen/logrus"
 	"github.com/trento-project/trento/internal/cluster"
+	"github.com/trento-project/trento/web/entities"
 	"github.com/trento-project/trento/web/models"
 	"gorm.io/gorm"
 	"gorm.io/gorm/clause"
@@ -43,12 +44,12 @@ func clustersProjector_ClusterDiscoveryHandler(event *DataCollectedEvent, db *go
 }
 
 // transformClusterData transforms the cluster data into the read model
-func transformClusterData(cluster *cluster.Cluster) (*models.Cluster, error) {
+func transformClusterData(cluster *cluster.Cluster) (*entities.Cluster, error) {
 	if cluster.Id == "" {
 		return nil, fmt.Errorf("no cluster ID found")
 	}
 
-	return &models.Cluster{
+	return &entities.Cluster{
 		ID:          cluster.Id,
 		Name:        cluster.Name,
 		ClusterType: detectClusterType(cluster),
--- a/web/services/clusters.go
+++ b/web/services/clusters.go
@@ -58,7 +58,7 @@ func (s *clustersService) GetAll(filters map[string][]string) (models.ClusterLis
 		}
 	}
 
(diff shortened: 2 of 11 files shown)


web/datapipeline/clusters_projector.go
```diff
@@ -43,12 +44,12 @@ func clustersProjector_ClusterDiscoveryHandler(event *DataCollectedEvent, db *go
 }
 
 // transformClusterData transforms the cluster data into the read model
```
nelsonkopliku, line 46: > transformClusterData transforms the cluster data into the read model
>
I'd probalby rephrase to `entity` instead of Read Model.
wdyt?

web/services/clusters.go
```diff
@@ -48,28 +58,76 @@ func (s *clustersService) GetAll(filters map[string][]string) (models.ClusterLis
...
+	return clusterTypes, nil
+}
+
+func (s *clustersService) GetAllSIDs() ([]string, error) {
+	var sids pq.StringArray
+
+	err := s.db.Model(&entities.Cluster{}).
+		Where("sids IS NOT NULL").
+		Distinct().
+		Pluck("unnest(sids)", &sids).
```
nelsonkopliku, line 104: Is the `Distinct` going to be applied after `unnesting`? or before? 

Asking because I don't know how `Distinct` works with `text[]`

:innocent: :flushed:

nelsonkopliku (approved):
Thanks for this! LGTM.

web/services/clusters.go
fabriziosestito: distinct just distinct to the select clause, while unnest expands the array in rows.. this is all pgsql magic and we don't have to worry about it.

nelsonkopliku: yeah sure :joy: thanks

fabriziosestito: @nelsonkopliku btw i dont think that the order matters in gorm, i believe that till you don't end the chain with an  ending method (find, first, scan, pluck, etc) the condition is added to the query builder in a smart way

nelsonkopliku: makes sense. Thanks for clarifying!

fabriziosestito:
Resolves: #455

dottorblaster (approved):
LGTM

web/services/clusters.go
```diff
@@ -58,7 +58,7 @@ func (s *clustersService) GetAll(filters map[string][]string) (models.ClusterLis
 		}
 	}
 
-	err := db.Find(&clusters).Error
+	err := db.Find(&clusters).Order("name").Order("id").Error
```
arbulu89, line 61: Hey @fabriziosestito ,
Just to be sure. Did you test if this works?
When I played a bit with this, I couldn't get the correct ordering if I didn't put the `Order` clause before the find.
Like this:
```db.Order("name").Order("id").Find(&clusters).```

Just as a comment hehe

fabriziosestito: Resolved thanks!

PR #451, opened 2021-11-15
Cleanup filtering & pagination

fabriziosestito:
This PR cleans up the filtering in the new db based services.
It also leverages the database for pagination adding a new gorm scope.

### DIFF
--- a/web/clusters.go
+++ b/web/clusters.go
@@ -3,6 +3,7 @@ package web
 import (
 	"fmt"
 	"net/http"
+	"strconv"
 	"strings"
 
 	"github.com/gin-gonic/gin"
@@ -292,7 +293,29 @@ func NewClusterListHandler(clustersService services.ClustersService) gin.Handler
 	return func(c *gin.Context) {
 		query := c.Request.URL.Query()
 
-		clusterList, err := clustersService.GetAll(query)
+		clustersFilter := &services.ClustersFilter{
+			Name:        query["name"],
+			SIDs:        query["sids"],
+			ClusterType: query["cluster_type"],
+			Health:      query["health"],
+			Tags:        query["tags"],
+		}
+
+		pageNumber, err := strconv.Atoi(c.DefaultQuery("page", "1"))
+		if err != nil {
+			pageNumber = 1
+		}
+		pageSize, err := strconv.Atoi(c.DefaultQuery("per_page", "10"))
+		if err != nil {
+			pageSize = 10
+		}
+
+		page := &services.Page{
+			Number: pageNumber,
+			Size:   pageSize,
+		}
+
+		clusterList, err := clustersService.GetAll(clustersFilter, page)
 		if err != nil {
 			_ = c.Error(err)
 			return
@@ -319,13 +342,15 @@ func NewClusterListHandler(clustersService services.ClustersService) gin.Handler
 		healthContainer := NewClustersHealthContainer(clusterList)
 		healthContainer.Layout = "horizontal"
 
-		page := c.DefaultQuery("page", "1")
-		perPage := c.DefaultQuery("per_page", "10")
-		pagination := NewPaginationWithStrings(len(clusterList), page, perPage) // not nice here
-		firstElem, lastElem := pagination.GetSliceNumbers()
+		count, err := clustersService.GetCount()
+		if err != nil {
+			_ = c.Error(err)
+			return
+		}
+		pagination := NewPagination(count, pageNumber, pageSize)
 
 		c.HTML(http.StatusOK, "clusters.html.tmpl", gin.H{
-			"ClustersTable":      clusterList[firstElem:lastElem],
+			"ClustersTable":      clusterList,
 			"AppliedFilters":     query,
 			"FilterClusterTypes": filterClusterTypes,
 			"FilterSIDs":         filterSIDs,
--- a/web/hosts_next.go
+++ b/web/hosts_next.go
@@ -2,6 +2,7 @@ package web
 
 import (
 	"net/http"
+	"strconv"
 
 	"github.com/gin-gonic/gin"
 	"github.com/trento-project/trento/web/models"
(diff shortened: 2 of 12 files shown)


web/services/hosts_next.go
```diff
@@ -59,9 +66,8 @@ func (s *hostsNextService) GetAll(filters map[string][]string) (models.HostList,
 		host := h.ToModel()
 		host.Health = computeHealth(&h)
 
-		health, ok := filters["health"]
-		if ok && len(health) > 0 {
-			if !internal.Contains(health, host.Health) {
+		if filter != nil && len(filter.Health) > 0 {
+			if !internal.Contains(filter.Health, host.Health) {
```
dottorblaster, line 70: I know what you'll say, but making a function for these conditions is always viable to me :D

dottorblaster (approved):
LGTM

web/services/clusters_test.go
```diff
@@ -96,7 +96,7 @@ func (suite *ClustersServiceTestSuite) TestClustersService_GetAll() {
 	suite.checksService.On("GetAggregatedChecksResultByCluster", "2").Return(&AggregatedCheckData{WarningCount: 1}, nil)
 	suite.checksService.On("GetAggregatedChecksResultByCluster", "3").Return(&AggregatedCheckData{CriticalCount: 1}, nil)
 
-	clusters, _ := suite.clustersService.GetAll(map[string][]string{})
+	clusters, _ := suite.clustersService.GetAll(nil, nil)
```
nelsonkopliku, line 99: I find it a bit meh having to call `.GetAll(nil. nil)` to get the complete list.
I'd maybe design differently the interface with a method that actually gets all the clusters, and one for filtered purposes.

Yet I understand why you did this, so it's a good iteration.

web/clusters.go
```diff
@@ -292,7 +293,29 @@ func NewClusterListHandler(clustersService services.ClustersService) gin.Handler
 	return func(c *gin.Context) {
 		query := c.Request.URL.Query()
 
-		clusterList, err := clustersService.GetAll(query)
+		clustersFilter := &services.ClustersFilter{
```
arbulu89, line 296: This way of binding looks quite anti gin-gonic. What about using `c.Bind` here?
Here an example: https://github.com/gin-gonic/gin/issues/742
In summary, we would need a struct (`ClustersFilter`) which would get all the information.
I would give a try at least

web/clusters.go
```diff
@@ -319,13 +342,15 @@ func NewClusterListHandler(clustersService services.ClustersService) gin.Handler
 		healthContainer := NewClustersHealthContainer(clusterList)
 		healthContainer.Layout = "horizontal"
 
-		page := c.DefaultQuery("page", "1")
-		perPage := c.DefaultQuery("per_page", "10")
-		pagination := NewPaginationWithStrings(len(clusterList), page, perPage) // not nice here
-		firstElem, lastElem := pagination.GetSliceNumbers()
+		count, err := clustersService.GetCount()
```
arbulu89, line 345: Do we really need this `GetCount` call? Why not just get the len of the `clusterList` ?

web/hosts_next.go
```diff
@@ -27,7 +28,27 @@ func NewHostListNextHandler(hostsService services.HostsNextService) gin.HandlerF
 	return func(c *gin.Context) {
 		query := c.Request.URL.Query()
 
-		hostList, err := hostsService.GetAll(query)
+		hostsFilter := &services.HostsFilter{
```
arbulu89, line 31: Same here. I would try to use the `Bind` method

nelsonkopliku (approved):
Great!

web/services/clusters_test.go
fabriziosestito: I have to say i did a bit of reasearch  and the use of nil is pretty common also in the golang standard library.
There are other tricks with variadic arugments but they mean a lot of boiler code that I'm not confident will be helpful at the end.
On the other hand, adding another method will add more code to test and mantain, so again the idea is to minimize that since we are going to use that method with filtering and pagination 99.9% of the time.

web/services/hosts_next.go
fabriziosestito: yep im not sure this will stay i will keep on the radar for the next iterations!

web/services/clusters_test.go
dottorblaster: @nelsonkopliku I think the same about it, but bear in mind that ignoring the "nil is evil" line of thoughs is a cornerstone in Golang's design. Hence, the researches done by @fabriziosestito show something that I absolutely dislike, but extremely idiomatic at the same time :smile:

arbulu89 (changes requested):
Hey @fabriziosestito ,
Nice migration!
I find a bit anti gin-gonic how the query binding is done. Even though that it is not documented properly, the query entries can be bind to an struct in the same way than the post body data.
I have put an example link in one of the comments.

web/clusters.go
arbulu89: Ok, nevermind. We cannot use the len of the `clusterList` as it already has the filters, and we need the total number of items

web/services/clusters_test.go
fabriziosestito: @nelsonkopliku @dottorblaster see the edit of my comment, i was also biased against nil eheh if we find a better solution i am all hear but i could not!

nelsonkopliku: makes sense guys, thanks for clarification. 
Sometimes I still have to wrap my head around some go quirks, but that's ok, just need to deal with it :sweat_smile:

web/clusters.go
fabriziosestito: @arbulu89 thanks i will try this out!

fabriziosestito: @arbulu89 i tried this and i am not sure since it involves adding a ``form`` annotation to the `ClustersService` struct while I would prefer to keep the to layers decoupled or at least not having the service layer depending on the controller one at all.

arbulu89: Well, I guess the ideal would be to have an additional struct that receives the query data with the `form` annotations, and it can be converted to the other page filter. Now that we have api structs, entities, models, yet another layer hehe

Anyway, feel free to implement as you wish. I just thought that using the gin gonic native way to deal with query parameters was better.

I will remove my request changes review.

PR #452, opened 2021-11-15
Uniform runner config loading to config files

nelsonkopliku:
Makes the runner load configuration the same way as the web and the agent do.

### DIFF
--- a/cmd/runner/runner.go
+++ b/cmd/runner/runner.go
@@ -11,6 +11,7 @@ import (
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
 
+	"github.com/trento-project/trento/internal"
 	"github.com/trento-project/trento/runner"
 )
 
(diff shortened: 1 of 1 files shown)


arbulu89 (approved):
Thanks man!

This should fix the regression introduce in: https://github.com/trento-project/trento/commit/99077570bf46b0417f6f4a10a005f2763e0f1b54

As a test, if the `--log-level debug` option in the runner, enables the ansible execution logs, I would consider it fixed at least hehe

nelsonkopliku:
> Thanks man!
> 
> This should fix the regression introduce in: [9907757](https://github.com/trento-project/trento/commit/99077570bf46b0417f6f4a10a005f2763e0f1b54)
> 
> As a test, if the `--log-level debug` option in the runner, enables the ansible execution logs, I would consider it fixed at least hehe
>
Yes, my bad.
If fixes the thing :pray:

rtorrero:
Awesome, thanks!

rtorrero (approved):

Issue #453, opened 2021-11-15
Swagger page calls are broken


arbulu89:
All the swagger page calls are broken. 
Swagger is adding a double `api/api` prefix in the urls, which creates the issue.
I guess this regression happened in the `api/docs` transition.

We would need to update the swagger docstrings of the handler, or find some additional fix

PR #457, opened 2021-11-15
Remove /api prefix from swagger api docstrings

arbulu89:
Fix for: https://github.com/trento-project/trento/issues/453

I guess the error was introduced in: https://github.com/trento-project/trento/pull/411

PD: I don't know why it has introduced the `github.com/alecthomas/template"` package honestly. I'm using the swag:
```
trento> swag --version
swag version v1.7.0
```

### DIFF
--- a/docs/api/docs.go
+++ b/docs/api/docs.go
@@ -31,7 +31,7 @@ var doc = `{
     "host": "{{.Host}}",
     "basePath": "{{.BasePath}}",
     "paths": {
-        "/api/checks/catalog": {
+        "/checks/catalog": {
             "get": {
                 "produces": [
                     "application/json"
@@ -90,7 +90,7 @@ var doc = `{
                 }
             }
         },
-        "/api/checks/{id}/settings": {
+        "/checks/{id}/settings": {
             "get": {
                 "consumes": [
                     "application/json"
@@ -171,7 +171,7 @@ var doc = `{
                 }
             }
         },
-        "/api/clusters/{cluster_id}/results": {
+        "/clusters/{cluster_id}/results": {
             "get": {
                 "produces": [
                     "application/json"
@@ -206,7 +206,7 @@ var doc = `{
                 }
             }
         },
-        "/api/clusters/{id}/tags": {
+        "/clusters/{id}/tags": {
             "post": {
                 "consumes": [
                     "application/json"
@@ -270,7 +270,7 @@ var doc = `{
                 }
             }
         },
-        "/api/clusters/{id}/tags/{tag}": {
+        "/clusters/{id}/tags/{tag}": {
             "delete": {
                 "consumes": [
                     "application/json"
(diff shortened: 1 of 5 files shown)


nelsonkopliku (approved):
:+1:

fabriziosestito:
package honestly. I'm using the swag:
> 
> ```
> trento> swag --version
> swag version v1.7.0
> ```

@arbulu89  swag should be pinned for everyone to 1.7.4 by the `tools/tools.go` package, could you confirm that installing it when cwd in trento repository root installs the above?

fabriziosestito (changes requested):
package honestly. I'm using the swag:
> 
> ```
> trento> swag --version
> swag version v1.7.0
> ```

@arbulu89  swag should be pinned for everyone to 1.7.4 by the `tools/tools.go` package, could you confirm that installing it when cwd in trento repository root installs the above?

I tried to run the generation and with the pinned version the file differs, and the `"github.com/alecthomas/template"` is not there anymore.


EDIT: I also tried to access `/api/doc/index.html` on the main branch build, and it works correctly, so maybe something related to the `swag` version mismatch?

nelsonkopliku:
> package honestly. I'm using the swag:
> 
> > ```
> > trento> swag --version
> > swag version v1.7.0
> > ```
> 
> @arbulu89 swag should be pinned for everyone to 1.7.4 by the `tools/tools.go` package, could you confirm that installing it when cwd in trento repository root installs the above?
> 
> I tried to run the generation and with the pinned version the file differs, and the `"github.com/alecthomas/template"` is not there anymore.
> 
> EDIT: I also tried to access `/api/doc/index.html` on the main branch build, and it works correctly, so maybe something related to the `swag` version mismatch?
>

Agreed on the version, if we can pin, that helps.

I also tried this http://demo.trento-project.io/api/docs/index.html#/default/get_api_checks_catalog and the problem I think is not the page itself, but the specific APIs execution from the Swagger UI that. breaks

image

stefanotorresi:
ok, so there are multiple things to note here:
- Xabi is using an outdated version of swag. Re-running `go install github.com/swaggo/swag/cmd/swag`.
- #411 added this: https://github.com/trento-project/trento/blob/93bf0eb1188853d2d33450cf7da685726fdcd68f/web/api.go#L20 without removing `/api` from all the endpoints, so it's effectively duplicated.

The ideal solution would be to search and remove `/api` prefix from all the endpoint paths, and let the global `basePath` apply.

fabriziosestito:
> Agreed on the version, if we can pin, that helps.

Version is already pinned, running `go install github.com/swaggo/swag/cmd/swag` in the root of the repository should do the work. 

> I also tried this http://demo.trento-project.io/api/docs/index.html#/default/get_api_checks_catalog and the problem I think is not the page itself, but the specific APIs execution from the Swagger UI that. breaks
> 

Ah alright I see now

arbulu89:
Thanks @nelsonkopliku @stefanotorresi and @fabriziosestito ,
`swag` updated to the correct version. Now everything looks fine.

fabriziosestito (approved):

PR #458, opened 2021-11-15
Add a log telling the configuration file being used

nelsonkopliku:
In order to give more information about the running instance of trento, If we found a config file and we were able lo load/read it, we also log it.

### DIFF
--- a/internal/config.go
+++ b/internal/config.go
@@ -6,6 +6,8 @@ import (
 	"path"
 	"strings"
 
+	log "github.com/sirupsen/logrus"
+
 	"github.com/mitchellh/go-homedir"
 	"github.com/spf13/viper"
 )
@@ -62,6 +64,10 @@ func InitConfig(configName string) error {
 		if _, ok := err.(viper.ConfigFileNotFoundError); !ok {
 			return err
 		}
+	}
+
+	if configFile := viper.ConfigFileUsed(); configFile != "" {
+		log.Infof("Using config file: %s", configFile)
 	}
 
 	return nil


internal/config.go
```diff
@@ -62,6 +64,8 @@ func InitConfig(configName string) error {
 		if _, ok := err.(viper.ConfigFileNotFoundError); !ok {
 			return err
 		}
+	} else {
```
arbulu89, line 67: We could print outside the `else` to reduce the a bit the complexity, as it if doesn't return here, the result is positive.
Like here: https://github.com/spf13/viper#reading-config-files

arbulu89 (approved):
Thanks @nelsonkopliku !

internal/config.go
nelsonkopliku: I can make it like 
```
err := viper.ReadInConfig()

if err != nil {
	if _, ok := err.(viper.ConfigFileNotFoundError); !ok {
		return err
	}
}

log.Infof("Using config file: %s", viper.ConfigFileUsed())
```

Then it happens that we have that line logged also in a scenario where we do not use any config file (none explicitly provided, none found in the expected locations)

That would print
```
time="2021-11-15 17:27:56" level=info msg="Using config file: "
```

Otherwise I'd need another if to check whether we have something in ` viper.ConfigFileUsed()`

wdyt?

nelsonkopliku: Or I can make a function like this

```
func usedConfigFile() string {
	configFile := viper.ConfigFileUsed()
	if configFile == "" {
		return "none"
	}
	return configFile
}
```

and use it like 
```
log.Infof("Using config file: %s", usedConfigFile())
```

That would print
```
msg="Using config file: none"

or

msg="Using config file: /etc/trento/agent.yaml"

or whatever the provided file is
```

Preferences?

fabriziosestito: I would just add
```go
if viper.ConfigFileUsed() != nil {
  log.Infof(......
}
```

nelsonkopliku: Thanks @fabriziosestito for bringing me back to earth :joy: 
Fixed that way

PR #459, opened 2021-11-15
Ci goodies 2: The Comeback

rtorrero:
This PR brings back https://github.com/trento-project/trento/pull/438 which had to be reverted due to failing builds. It should fix the issue by splitting the docker push & asset creation into two separate tasks.

Additionally, It was also required to remove the `-ti` arguments from the `docker` command line as it wasn't able to allocate a TTY when running from the GH action

### DIFF
--- a/.github/workflows/ci-cd.yaml
+++ b/.github/workflows/ci-cd.yaml
@@ -93,6 +93,80 @@ jobs:
       - name: trento checks ID sanity test
         run: python3 hack/id_checker.py
 
+  build-container-images:
+    runs-on: ubuntu-20.04
+    needs: [test-binary, test-checks]
+    strategy:
+      matrix:
+        image:
+          - trento-web
+          - trento-runner
+    permissions:
+      contents: read
+      packages: write
+    concurrency: ci-${{ github.ref }}
+    env:
+      REGISTRY: ghcr.io
+      IMAGE_REPOSITORY: ghcr.io/${{ github.repository_owner }}/${{ matrix.image }}
+      IMAGE_TAG: "${{ (github.event_name == 'release' && github.event.release.tag_name) || github.ref_name == 'main' && 'rolling' || github.sha }}"
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - uses: docker/setup-buildx-action@v1
+      - name: Log in to the Container registry
+        uses: docker/login-action@f054a8b539a109f9f41c372932f1ae047eff08c9
+        if: github.event_name == 'release' || github.ref_name == 'main'
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Extract metadata (tags, labels) for Docker
+        id: meta
+        uses: docker/metadata-action@98669ae865ea3cffbcbaa878cf57c20bbf1c6c38
+        with:
+          images: ${{ env.IMAGE_REPOSITORY }}
+      - name: Build and push container image
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          push: "${{ github.event_name == 'release' || github.ref_name == 'main' }}"
+          target: ${{ matrix.image }}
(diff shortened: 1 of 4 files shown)


dottorblaster:
Rating 10/10 just for the name

.github/workflows/ci-cd.yaml
```diff
@@ -94,8 +94,88 @@ jobs:
...
+          target: ${{ matrix.image }}
+          tags: ${{ env.IMAGE_REPOSITORY }}:${{ env.IMAGE_TAG }}
+          labels: ${{ steps.meta.outputs.labels }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+      - name: Create image artifact
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          push: false
```
stefanotorresi, line 144: we're missing the `target` field here!

.github/workflows/ci-cd.yaml
```diff
@@ -94,8 +94,88 @@ jobs:
...
+      IMAGE_REPOSITORY: ghcr.io/${{ github.repository_owner }}/${{ matrix.image }}
+      IMAGE_TAG: "${{ (github.event_name == 'release' && github.event.release.tag_name) || (github.event_name == 'push' && github.ref_name == 'main' && 'rolling') || github.sha }}"
+      PUSH: "${{ github.event_name == 'release' || (github.event_name == 'push' && github.ref_name == 'main') || github.event_name == 'workflow_dispatch' }}"
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - uses: docker/setup-buildx-action@v1
+      - name: Log in to the Container registry
+        if: ${{ env.PUSH }}
```
stefanotorresi, line 119: I did notice this also on my orignial PR, but for some reason this `if` is not doing anything.
I would suggest the following:
- try to remove the `${{}}` notation and set it to just `if: env.PUSH`, and if see it still gets run (it shouldn't. unless it is a push on main, a release, or a manual trigger)
- if the above doesn't change anything, then remove this `if` condition and let this step always run. In this case, you might want to also remove the `PUSH` environment variable and move the conditional expression in the below `push` field of the `build-and-push` action.

stefanotorresi (changes requested):
Thanks for looking into this, Rubén!
I have some feedback.

.github/workflows/ci-cd.yaml
rtorrero: Do we need this at the `step` level or could we move it to the job level? It seems to me that what we want is the whole job to get skipped (not that specific step only)

.github/workflows/ci-cd.yaml
rtorrero: Yes! good catch

.github/workflows/ci-cd.yaml
stefanotorresi: no, this can't be at the job level, because currently the same job has two unrelated group of steps:
- log in the registry, build the images, and push them to the registry. These only happens if it's a release or a merge to main.
- build the images (again, but there is cache, so takes almost no time), and export them as an artifact. This happens regardless of whether or not it's a release or a merge to main.

If the `if` condition at the step level doesn't work, leave it out and let the log-in happen all times, even if we only need for the push to registry.

Otherwise, you could try to split this entire thing into two jobs, maybe it'll be cleaner.

rtorrero: I've splitted the thing into two jobs and the `if` now works correctly. It builds the container assets but doesn't push the image unless it needs to. Could you check it again?

stefanotorresi: ok, now we have two new minor problems:
- naming: the difference between the two jobs is not super clear. I suggest something like `build-and-push-containers` and `build-and-export-containers`, but feel free to propose something else.
- dependencies: `smoke-test-container-images` now depends on `build-and-export-containers`, while `deploy-server` depends on both `smoke-test-container-images` and `build-and-push-containers`. Make sure the flow is correct.

rtorrero: should be ok now!

stefanotorresi (approved):
LGTM!

PR #461, opened 2021-11-16
Migrate about page and sles subscriptions data to the database

arbulu89:
Start using the discovered subscription raw data in the database to populate the About page, and the subscriptions table in the host details page.

It includes:
- New projection to store the SLES subscriptions as read only. It has a fixed struct, but we could move it to a jsonb style. I don't expect almost any change in these values though
- New `SlesSubsription` entity and model created. They are almost identical, but I guess it will help to keep things decoupled
- Refactor the subscriptions service to use data in the new read models instead of consul
- Use the new service functions in the About page and the host details page to show the subscriptions
- Remove outdated subscriptions consul code. This means that we don't store anymore the subscriptions code in consul in the agent side. It is not used anywhere else, so I just got rid off it

PD: I took advantage for renaming the `Free` word to `Community` as it is the name we use always


dottorblaster:
+140
-313

I LOVE this :rocket:

web/datapipeline/sles_subsciptions_projector.go
```diff
@@ -0,0 +1,49 @@
...
+    	SubscriptionStatus: subscription.SubscriptionStatus,
+  	}
+
+		subEntities = append(subEntities, subEntity)
+  }
+
+	return db.Clauses(clause.OnConflict{
+		UpdateAll: true,
+	}).Create(&subEntities).Error
+}
```
fabriziosestito, line 50: I think we need to delete also the subscription that are not present anymore here,
like just affer creating:
```go
if err := tx.Not(subEntity).Delete(&entities.SlesSubscription{}).Error; err != nil {
		return err
	}
```

fabriziosestito (approved):

web/datapipeline/sles_subsciptions_projector.go
arbulu89: Hey @fabriziosestito ,
Your suggestion was not working completely. For example in the case where the subscription list was coming empty.
I have implemented it this way: https://github.com/trento-project/trento/pull/461/commits/7ce64e8768409748aa527dfeae1acb20e6615c8a

By the way, this removal thing looks even needed in the clusters and sap systems. It is much more complicated as the cluster belongs to multiple nodes, so an straightforward removal is not an option

fabriziosestito: @arbulu89 you are right, we need to deal with the deletion (or filtering)  in clusters and sap systems as well. I will think about it and get back to you.
Maybe I misunderstood the use-case here, but I thought it could be possible that an agent has an expired subscription and it doesn't send the data regarding that subscription anymore. In that case maybe deleting the row here is the way to go. What are your thoughts?

arbulu89: Your assumption is Ok, and not only that, the user can remove their subscription letting the machine unsubscribed (for any reason they have). In this case, we would receive an empty list of subscriptions.
That's why I implemented the `1st delete all, 2nd add the new subscriptions`.
I didn't find any way of doing it, keeping in mind that we cannot remove all the subscriptions, we need to take care of the subscription of this particular host.

web/services/subscriptions.go
```diff
@@ -1,70 +1,60 @@
...
-type SubscriptionData struct {
-	Type            string
-	SubscribedCount int
-}
-
 //go:generate mockery --name=SubscriptionsService --inpackage --filename=subscriptions_mock.go
 type SubscriptionsService interface {
-	GetSubscriptionData() (*SubscriptionData, error)
-	GetHostSubscriptions(host string) (subscription.Subscriptions, error)
+	IsTrentoPremium() (bool, int64, error) // Returns if trento is premium or not and the number of subscribed SLES_SAP machines
```
dottorblaster, line 16: `IsTrentoPremium` sounds to me like a `bool, error` return, maybe this could be `GetPremiumData`? Bear in mind we can return a struct with this shape

```go
type PremiumData struct {
    Premium    bool
    Instances  int64
}
```

What do you think?

arbulu89: This was the way actually it was implemented before, and I had big doubts on changing it.
I don't have any issue changing it back (using more or less the same structure you mentioned).
Any other opinion?

nelsonkopliku: I agree with @dottorblaster.

A `IsTrentoPremium` would be useful to do something like
```
// pseudocode
if is trento premium
    do this
otherwise 
    do that
```
I see it clearer in this context to have the `GetPremiumData` or the `GetSubscriptionData` you already had.
Maybe we need also a `IsTrentoPremium`, probably not for this feature.

nelsonkopliku (approved):

arbulu89:
Hey @nelsonkopliku @dottorblaster ,
I have added a new `GetPremiumData` function as you suggested (still keeping the `IsTrentoPremium` there, it will be used in the future most probably).
Everything else remains the same.
Could you have a new look?
Thanks!

dottorblaster (approved):

nelsonkopliku (approved):
Great!

PR #462, opened 2021-11-16
Enable SAPSystems projection

dottorblaster:
This PR is a first step toward a makeover of the SAP Systems view. With this we can project the raw data onto entities persisted inside the DB, for further aggregation.

### DIFF
--- a/web/datapipeline/sap_systems_projector.go
+++ b/web/datapipeline/sap_systems_projector.go
@@ -3,11 +3,15 @@ package datapipeline
 import (
 	"fmt"
 
+	log "github.com/sirupsen/logrus"
+	"github.com/trento-project/trento/internal/sapsystem"
+	"github.com/trento-project/trento/web/entities"
 	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
 )
 
 func NewSAPSystemsProjector(db *gorm.DB) *projector {
-	SAPSystemsProjector := NewProjector("hosts", db)
+	SAPSystemsProjector := NewProjector("sapsystems", db)
 
 	SAPSystemsProjector.AddHandler(SAPsystemDiscovery, SAPSystemsProjector_SAPSystemsDiscoveryHandler)
 
@@ -15,6 +19,67 @@ func NewSAPSystemsProjector(db *gorm.DB) *projector {
 }
 
 func SAPSystemsProjector_SAPSystemsDiscoveryHandler(dataCollectedEvent *DataCollectedEvent, db *gorm.DB) error {
-	fmt.Println(dataCollectedEvent)
+	decoder := getPayloadDecoder(dataCollectedEvent.Payload)
+	var discoveredSAPSystems sapsystem.SAPSystemsList
+	if err := decoder.Decode(&discoveredSAPSystems); err != nil {
+		log.Errorf("can't decode data: %s", err)
+		return err
+	}
+
+	for _, s := range discoveredSAPSystems {
+		instance := entities.SAPSystemInstance{
+			AgentID:  dataCollectedEvent.AgentID,
+			SystemID: s.Id,
+			SID:      s.SID,
+		}
+
+		switch s.Type {
+		case 1:
+			instance.Type = sapsystem.SAPSystemsDatabase
+		case 2:
+			instance.Type = sapsystem.SAPSystemsApplication
+			instance.DBHost = fmt.Sprint(s.Profile["SAPDBHOST"])
+		}
+
+		for _, i := range s.Instances {
+			var features string
+			var instanceNumber string
+
+			if p, ok := i.SAPControl.Properties["SAPSYSTEM"]; ok {
+				instanceNumber = p.Value
+			} else {
+				log.Warnf("Instance Number not found for %s", s.SID)
+				continue
+			}
+
+			for _, SAPControlInstance := range i.SAPControl.Instances {
+				if instanceNumber == fmt.Sprintf("%02d", SAPControlInstance.InstanceNr) {
+					features = SAPControlInstance.Features
+					break
+				}
+			}
+
+			instance.Features = features
+			instance.InstanceNumber = instanceNumber
+
+			err := storeSAPInstance(db, instance, "system_id", "sid", "type", "features", "instance_number")
+			if err != nil {
+				return err
+			}
+
+		}
+	}
+
 	return nil
+}
+
+func storeSAPInstance(db *gorm.DB, sapInstance entities.SAPSystemInstance, updateColumns ...string) error {
+	return db.Clauses(clause.OnConflict{
+		Columns: []clause.Column{
+			{Name: "agent_id"},
+			{Name: "system_id"},
+			{Name: "instance_number"},
+		},
+		DoUpdates: clause.AssignmentColumns(append(updateColumns, "updated_at")),
+	}).Create(&sapInstance).Error
 }
--- a/web/datapipeline/sap_systems_projector.go
+++ b/web/datapipeline/sap_systems_projector.go
@@ -1,0 +1,20 @@
+package datapipeline
+
+import (
+	"fmt"
+
+	"gorm.io/gorm"
+)
+
+func NewSAPSystemsProjector(db *gorm.DB) *projector {
+	SAPSystemsProjector := NewProjector("hosts", db)
+
+	SAPSystemsProjector.AddHandler(SAPsystemDiscovery, SAPSystemsProjector_SAPSystemsDiscoveryHandler)
+
+	return SAPSystemsProjector
+}
+
+func SAPSystemsProjector_SAPSystemsDiscoveryHandler(dataCollectedEvent *DataCollectedEvent, db *gorm.DB) error {
+	fmt.Println(dataCollectedEvent)
+	return nil
+}
--- a/web/entities/sapsystem.go
+++ b/web/entities/sapsystem.go
@@ -1,0 +1,19 @@
+package entities
+
+import (
+	"time"
+)
+
+type SAPSystemInstance struct {
+	AgentID           string `gorm:"primaryKey"`
+	Type              string
+	SystemID          string `gorm:"primaryKey"`
+	SID               string `gorm:"column:sid"`
+	InstanceNumber    string `gorm:"primaryKey"`
+	Features          string
+	Description       string
+	SystemReplication string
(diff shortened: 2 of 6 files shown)


web/entities/sapsystem.go
```diff
@@ -0,0 +1,19 @@
+package entities
+
+import (
+	"time"
+)
+
+type SAPSystemInstance struct {
```
arbulu89, line 7: I thinks we would need to have the `SAPSystem` and `SAPInstance` structs combo.
Why?
Because we can have some SAP instance with a different SID than the SAP system itself (the database doesn't need to have the same SID than the SAP system).

Later on, 1 SAP system is composed by multiple instances, that can have different SID and type. And even different parameters depending on the type. For example, as `application` type instances don't have any System replication data.

Maybe some kind of polymorfism would be a good idea here.

arbulu89 (commented):
Hey @dottorblaster ,
I'm afraid this implementation can be too simplistic. We would need to see better how the SAP systems composition is done to relate properly the systems with the databases, and all the instances within

web/entities/sapsystem.go
fabriziosestito: @arbulu89 the idea is that this is how we store the data from the events, the `SAPSystem`, aggregation would come in the service when we will transform (group) instances to a `model.SAPSystem` which will hold many instances.
The database aggregation is quite easy since we will search for a `entities.Host` having the same IP as the `DBHost` field here.
As for the SystemReplication data or other fields the idea was to keep them empty in case of the `application` type and just ignore them when casting that to a model.
Basically, polymorphism will be handled at service level not at persistence level.

dottorblaster: @fabriziosestito said that way better than me :smile: 

I agree that we need to tell the difference between different models, but this shall be done at service level, this PR only introduces a basic projector so we can move forward with other works

arbulu89: @fabriziosestito Ok, fair enough. I don't think it will scale well in the future, because we will need to have SAP system specific data, but I guess we will be able to change it

web/datapipeline/sap_systems_projector.go
```diff
@@ -0,0 +1,81 @@
...
+	"fmt"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/trento-project/trento/internal/sapsystem"
+	"github.com/trento-project/trento/web/entities"
+	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
+)
+
+const SAPSystemsApplication = "application"
```
arbulu89, line 13: I think we could unify these names in `internal` and use the same value in the agent and here. We can even change the agent to send a string if it helps.

web/datapipeline/sap_systems_projector.go
```diff
@@ -0,0 +1,81 @@
...
+
+func SAPSystemsProjector_SAPSystemsDiscoveryHandler(dataCollectedEvent *DataCollectedEvent, db *gorm.DB) error {
+	decoder := getPayloadDecoder(dataCollectedEvent.Payload)
+	var discoveredSAPSystems sapsystem.SAPSystemsList
+	if err := decoder.Decode(&discoveredSAPSystems); err != nil {
+		log.Errorf("can't decode data: %s", err)
+		return err
+	}
+
+	for _, s := range discoveredSAPSystems {
```
arbulu89, line 32: I see some issues in this loop. 
Here, we are looping the SAP systems installed in 1 host. There might be 1..n system.
But within the SAP system, we can have multiple SAP instances installed, 1..n again. So, I miss an additional loop here, which creates the SAP instance entry itself.

Example: We have a netweaver installation that has the ASCS and PAS in the same machine. Both belong to the same SAP system, but are 2 different instance (2 rows in the UI table).

So most probably, we would need to run the `storeSAPInstance` function within the `s.Instaces` loop.

That's why I see that the instances and systems are kind of coupled. But I guess we can introduce the system data in the instance as well (as they are parent/children somehow)

arbulu89 (changes requested):
Some more comments here

web/datapipeline/sap_systems_projector.go
dottorblaster: > That's why I see that the instances and systems are kind of coupled

Absolutely, I see the point but I wanted to do this at service level, populating a `SAPSystem` model grouping many `SAPSystemsInstance` entities cast into proper models.

> Example: We have a netweaver installation that has the ASCS and PAS in the same machine. Both belong to the same SAP system, but are 2 different instance (2 rows in the UI table).

Understood, I'll try to come up with something better, thanks for the feedback

web/datapipeline/sap_systems_projector.go
```diff
@@ -0,0 +1,84 @@
...
+		instance := entities.SAPSystemInstance{
+			AgentID:  dataCollectedEvent.AgentID,
+			SystemID: s.Id,
+			SID:      s.SID,
+		}
+
+		if s.Type == 1 {
```
arbulu89, line 36: Could we put this in a `switch/cas` fashion at least. It looks the most appropriate way

fabriziosestito (approved):
LGTM! Good first iteration

web/datapipeline/sap_systems_projector.go
```diff
@@ -0,0 +1,84 @@
...
+	var discoveredSAPSystems sapsystem.SAPSystemsList
+	if err := decoder.Decode(&discoveredSAPSystems); err != nil {
+		log.Errorf("can't decode data: %s", err)
+		return err
+	}
+
+	for _, s := range discoveredSAPSystems {
+		instance := entities.SAPSystemInstance{
+			AgentID:  dataCollectedEvent.AgentID,
+			SystemID: s.Id,
```
arbulu89, line 32: I'm a bit troubled with the name, but I recognize that it is not simple hehe
Maybe, we could rename `SystemID` to simply `ID`? `SID` at the end is `system id` as well
It is the internal ID of the system at the end.
By the way, in the future we will need to convert this to `uuid` format. Nothing to be done in the this PR though

web/datapipeline/sap_systems_projector.go
```diff
@@ -0,0 +1,84 @@
...
+		}
+
+		for _, i := range s.Instances {
+			var features string
+			var instanceNumber string
+
+			if p, ok := i.SAPControl.Properties["SAPSYSTEM"]; ok {
+				instanceNumber = p.Value
+			} else {
+				continue
```
arbulu89, line 52: This `else` case is something that shouldn't never happen. Could we put some kind of `warning` logging?

web/datapipeline/sap_systems_projector.go
```diff
@@ -0,0 +1,84 @@
...
+			}
+
+			for _, SAPControlInstance := range i.SAPControl.Instances {
+				if instanceNumber == fmt.Sprintf("%02d", SAPControlInstance.InstanceNr) {
+					features = SAPControlInstance.Features
```
arbulu89, line 57: We can exit from the loop once this condition is met, as we won't never have it as true again for this sap control instance, or instance in general. Let's add a break

nelsonkopliku (approved):
Very nice! Step ahead! :rocket:

web/datapipeline/sap_systems_projector.go
```diff
@@ -0,0 +1,84 @@
...
+				}
+			}
+
+			instance.Features = features
+			instance.InstanceNumber = instanceNumber
+
+			err := storeSAPInstance(db, instance, "system_id", "sid", "type", "description", "features", "instance_number")
```
arbulu89, line 64: I would vote to **not** update the `description` as it is empty always. In fact, this field won't be auto-discovered. It is an old discussion, but it doesn't make much sense by now hehe

arbulu89 (changes requested):
Looking good!
I have added some more comments. Most of them are just my personal thought, feel free to not apply them if needed, but I think they deserve a little look

web/datapipeline/sap_systems_projector.go
fabriziosestito: @arbulu89 I asked @dottorblaster  to change this 🍡 
To be aligned, I see this expressed as SystemID in the UI, what is the SAP domain word for this one?

dottorblaster: Tough decision but `ID` is a slippery call :sweat_smile: I would prefer having little longer field names, and `ID` is just... well an ID? :smile:

web/datapipeline/sap_systems_projector.go
dottorblaster: I don't like breaks, I do this only because I love you

web/datapipeline/sap_systems_projector.go
arbulu89: The thing is that `SystemID` and `SID` are the same as `SID` stands for `System identifier`. The thing is that our `SystemID` is a an internal auto-generated ID that is **global**. The sap `SID` never was created to be global, that's why it is just a 3 chars/digits string.
In this case the `ID` is our ID, the identifier of the SAP system (not the instance).
Up to you guys, I just wanted to comment that `SystemID` and `SID` are synonims hehe One is used by SAP and the other by us

web/datapipeline/sap_systems_projector.go
```diff
@@ -49,19 +50,21 @@ func SAPSystemsProjector_SAPSystemsDiscoveryHandler(dataCollectedEvent *DataColl
 			if p, ok := i.SAPControl.Properties["SAPSYSTEM"]; ok {
 				instanceNumber = p.Value
 			} else {
+				log.Warn("Instance Number not found")
```
arbulu89, line 53: Let's improve this to be `log.Warnf("Instance Number not found for %s", s.SID)`

arbulu89 (approved):
Thanks @dottorblaster ,
Just added other 2 small comments XD but we don't need to have other round of review hehe

web/datapipeline/sap_systems_projector.go
fabriziosestito: Ah alright i thought that `SystemID` was our `ID` and `SID` was another thing :D

arbulu89: Yes, and it is.
`SystemID` our internal ID
`SID` sap system id -> `System identifier`
The difference is that one is ours and the other SAP internal hehe

Issue #454, opened 2021-11-15
Some errors or potential enhancements in the new cluster settings modal


arbulu89:
I have noticed some issues or inconsistencies in the new cluster settings modal. Some of the comments are just subjective things.
The unique major error is the first item.

- The group of checks are always shown in different order
- The width of the modal looks to narrow, it was wider with the previous version at least
- There is not any closing arrow in the top/right corner
- The modal doesn't close when we click outside (this was the old behaviour at least)
- The connection settings collapsable leaves an ugly gray row at the end when it is opened
- The right side collapse icon (down looking arrow) doesn't change the clicker to a pointer.
- The group check/uncheck works when you click in the whole row, rather than only the checkable (maybe this is wanted), but the same doesn't happen with the individual checks

ajaeger:
@arbulu89 , the first issue should be fixed with #463. The rest is still open, isn't it?

arbulu89:
> @arbulu89 , the first issue should be fixed with #463. The rest is still open, isn't it?

@dottorblaster Worked on this. He might know better the state of the ticket

PR #463, opened 2021-11-16
Address HANA cluster settings modal quirks

dottorblaster:
* now the modal closes when we click on the backdrop
* we have a close button in the upper corner
* pointer turns to a hand when hovering the accordion toggle
* catalog results are now sorted and don't get mixed up every refresh

### DIFF
--- a/web/checks_api.go
+++ b/web/checks_api.go
@@ -3,6 +3,7 @@ package web
 import (
 	"fmt"
 	"net/http"
+	"sort"
 
 	"github.com/gin-gonic/gin"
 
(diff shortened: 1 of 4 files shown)


fabriziosestito (approved):

nelsonkopliku (approved):
LGTM

dottorblaster:
This addresses a lot of #454

web/checks_api.go
```diff
@@ -60,6 +61,11 @@ func ApiChecksCatalogHandler(s services.ChecksService) gin.HandlerFunc {
 			g := JSONChecksGroup{Group: group.Group, Checks: group.Checks}
 			groupedCatalog = append(groupedCatalog, &g)
 		}
+
+		sort.SliceStable(groupedCatalog, func(i, j int) bool {
```
arbulu89, line 65: Hey @dottorblaster ,
This is not the correct way of ordering, it is not alphabetical hehe
The order of the groups is based on the name of the checks. So th
This was already codified this way here: https://github.com/trento-project/trento/blob/main/web/models/check.go#L91

The correct fix would be to add the ordering in line 59 like this:
```
for _, group := range checkGroups.OrderByName() {
```

arbulu89 (commented):
@dottorblaster A bit late, but a change is needed

PR #464, opened 2021-11-16
Add conditional in the spec file to detect TW and otherwise avoid missing macro

rtorrero:
This PR addresses the failing RPM builds in the build service due to the missing `%_distconfdir` macro for `/usr/etc/` that only exists in TW. Following [this](https://en.opensuse.org/openSUSE:Build_Service_cross_distribution_howto) guide we've introduced a if that conditionally builds the package using the new config file directory for packages or falls back to the classic `/etc/`.

Notice that the macro itself could be added easily but still, we are not allowed to own `/usr/etc` so we wouldn't be able to make an RPM package that creates `/usr/etc`.

### DIFF
--- a/packaging/suse/trento.spec
+++ b/packaging/suse/trento.spec
@@ -71,7 +71,11 @@ install -D -m 0755 %{shortname} "%{buildroot}%{_bindir}/%{shortname}"
 install -D -m 0644 packaging/systemd/trento-agent.service %{buildroot}%{_unitdir}/trento-agent.service
 
 # Install the default configuration files
+%if 0%{?suse_version} > 1500
 install -D -m 0640 packaging/config/agent.yaml %{buildroot}%{_distconfdir}/trento/agent.yaml
+%else
+install -D -m 0640 packaging/config/agent.yaml %{buildroot}%{_sysconfdir}/trento/agent.yaml
+%endif
 
 %pre
 %service_add_pre trento-agent.service
@@ -93,7 +97,12 @@ install -D -m 0640 packaging/config/agent.yaml %{buildroot}%{_distconfdir}/trent
 %{_bindir}/%{shortname}
 %{_unitdir}/trento-agent.service
 
+%if 0%{?suse_version} > 1500
 %dir %_distconfdir/trento
 %_distconfdir/trento/agent.yaml
+%else
+%dir %{_sysconfdir}/trento
+%config (noreplace) %{_sysconfdir}/trento/agent.yaml
+%endif
 
 %changelog


nelsonkopliku (approved):
Ah! Lovely ifs.
That's what I call AI

aleksei-burlakov (approved):
Perfectos!

dottorblaster (approved):
SHIP IT

arbulu89 (commented):
@rtorrero Have you tried simply using the generic way?
The `else` conditions.
In all of our other packages we have it this way (without the `_distconfigdir`).
It would make the spec file simpler at least (and functional in all the distros)

aleksei-burlakov:
> @rtorrero Have you tried simply using the generic way? The `else` conditions. In all of our other packages we have it this way (without the `_distconfigdir`). It would make the spec file simpler at least (and functional in all the distros)

It's basically choosing between `/etc` and `/usr/etc`. The motivation is that the trend is the` /usr/etc` but the sle 12 releases don't support the macro `_distconfdir`.

arbulu89:
> > @rtorrero Have you tried simply using the generic way? The `else` conditions. In all of our other packages we have it this way (without the `_distconfigdir`). It would make the spec file simpler at least (and functional in all the distros)
> 
> It's basically choosing between `/etc` and `/usr/etc`. The motivation is that the trend is the` /usr/etc` but the sle 12 releases don't support the macro `_distconfdir`.

That's why I'm asking if you just tried the `%{_sysconfdir}` the option for all the cases. This should work for sle12, sle15 and tumbleweed as far as I know (we do this in other packages at least). This would spare this `if` condition

rtorrero:
@arbulu89 yes, probably for now, `/etc` would work for TW but we will have to adopt this at some point in the near future if we want to keep publishing for openSUSE. 

Packaging guidelines change and if we want to move forward with them and have the packages available for TW as well as continue to support the older releases, we have to introduce conditionals in the build process.

@thkukuk wrote this about the topic -> https://github.com/thkukuk/atomic-updates_and_etc/blob/master/README.md
also https://en.opensuse.org/openSUSE:Packaging_UsrEtc explains a bit more on the packaging-specific details of it

arbulu89:
> @arbulu89 yes, probably for now, `/etc` would work for TW but we will have to adopt this at some point in the near future if we want to keep publishing for openSUSE.
> 
> Packaging guidelines change and if we want to move forward with them and have the packages available for TW as well as continue to support the older releases, we have to introduce conditionals in the build process.
> 
> @thkukuk wrote this about the topic -> https://github.com/thkukuk/atomic-updates_and_etc/blob/master/README.md also https://en.opensuse.org/openSUSE:Packaging_UsrEtc explains a bit more on the packaging-specific details of it

I'm not saying to use `/etc`. I'm just asking if the macros used for the `else` condition work for SLE15 as well, placing the config files in `/usr/etc`.
It is only a question. We have other packages with the same macros used by the `else` condition here shipping the files in `/usr/etc`.
Just wanting to be sure if we can spare this `if/else` condition. If the answer is `no`: go ahead

rtorrero:
> > @arbulu89 yes, probably for now, `/etc` would work for TW but we will have to adopt this at some point in the near future if we want to keep publishing for openSUSE.
> > Packaging guidelines change and if we want to move forward with them and have the packages available for TW as well as continue to support the older releases, we have to introduce conditionals in the build process.
> > @thkukuk wrote this about the topic -> https://github.com/thkukuk/atomic-updates_and_etc/blob/master/README.md also https://en.opensuse.org/openSUSE:Packaging_UsrEtc explains a bit more on the packaging-specific details of it
> 
> I'm not saying to use `/etc`. I'm just asking if the macros used for the `else` condition work for SLE15 as well, placing the config files in `/usr/etc`. It is only a question. We have other packages with the same macros used by the `else` condition here shipping the files in `/usr/etc`. Just wanting to be sure if we can spare this `if/else` condition. If the answer is `no`: go ahead

We explored together different possibilities to avoid this ugly %if stuff but we decided to move forward with it. Thanks!

PR #465, opened 2021-11-17
Order check groups by name in catalog endpoint

dottorblaster:
As above, my assumption to sort the check groups alphabetically in #463 was wrong, so thanks to @arbulu89 for pointing me to the right direction with this :smile:


arbulu89 (approved):
Thanks @dottorblaster

PR #466, opened 2021-11-17
Detect aws and gcp clouds in the agent

arbulu89:
Add `AWS` and `GCP` clouds autodetection in the agent. This data will be used in the server side and sent in telemetry.

The code is based on `crmsh`, as we have been refining there (together with the cloud providers themselves) the cloud auto-detection: https://github.com/ClusterLabs/crmsh/blob/master/crmsh/utils.py#L2055

This is how it looks like;
AWS:
image
image

GCP:
image
image

### DIFF
--- a/internal/cloud/metadata.go
+++ b/internal/cloud/metadata.go
@@ -2,6 +2,7 @@ package cloud
 
 import (
 	"os/exec"
+	"regexp"
 	"strings"
 
 	log "github.com/sirupsen/logrus"
@@ -9,7 +10,7 @@ import (
 
 const (
 	Azure = "azure"
-	Aws = "aws"
+	Aws   = "aws"
 	// DMI chassis asset tag for Azure machines, needed to identify wether or not we are running on Azure
 	// This is actually ASCII-encoded, the decoding into a string results in "MSFT AZURE VM"
 	azureDmiTag = "7783-7084-3265-9085-8269-3286-77"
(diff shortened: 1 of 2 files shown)


nelsonkopliku (approved):
Super cool!

The bare metal variant is missing, yet this is already a great improvement and I think we can add that later.
Shall we track it so we do not forget?

arbulu89:
> Super cool!
> 
> The bare metal variant is missing, yet this is already a great improvement and I think we can add that later. Shall we track it so we do not forget?

The baremetal thing is a shadowy thing. In our case, baremetal is everything that it is not in the cloud (yes, it is that generic). I don't know what kind of accuracy we need here, but from the beginning, the idea of this field was to detect the cloud itself, and everything that is not in our interest gets out.
So, the question actually is more to the stakeholders. What they expect to find here if the machines is not in `aws/gcp/azure`?

dottorblaster (approved):
LGTM, the bare metal thing can wait. @nelsonkopliku shall we open an issue for that with a specific question tag?

lee-martin:
> > > Super cool!
> > > The bare metal variant is missing, yet this is already a great improvement and I think we can add that later. Shall we track it so we do not forget?
> > 
> > 
> > The baremetal thing is a shadowy thing. In our case, baremetal is everything that it is not in the cloud (yes, it is that generic). I don't know what kind of accuracy we need here, but from the beginning, the idea of this field was to detect the cloud itself, and everything that is not in our interest gets out. So, the question actually is more to the stakeholders. What they expect to find here if the machines is not in `aws/gcp/azure`?
> 
> @arbulu89 I see, let's then discuss with stakeholders

What about rewording "cloud provider" into "provider"? To be is would still sound correct to say aws/gcp/azure are "providers" (without the cloud word) and anything that is now known for now could be labelled "unknown".

Then in the future we could progressively add in new providers like "vmware", "kvm", "dell server", "lenovo server". Not forgetting other cloud providers over time as well.

Adding new providers can wait, but I guess we should fix the name of the field soon though.

lee-martin:
@arbulu89  I validated the idea to rewording from "cloud provider" into "provider" as above, which they would like to have to make the field more flexible than cloud only. Can you please update at your convenience. (Sorry, forgot to mention you explicitly in my previous response).

arbulu89:
> @arbulu89 I validated the idea to rewording from "cloud provider" into "provider" as above, which they would like to have to make the field more flexible than cloud only. Can you please update at your convenience. (Sorry, forgot to mention you explicitly in my previous response).

Of course we can!
Could you open a ticket with the request?
Thanks

stefanotorresi:
issue collected in https://github.com/trento-project/telemetry/issues/14

PR #470, opened 2021-11-17
Include the consul-config-dir init in the agent install script

arbulu89:
Fix the `install-agent.sh` script to include the `consul-config-dir` variable in the trento agent configuration. 

Without this variable, consul-template puts the metadata file in other path (`/` or `/root` most probably) and the host list page cannot populate most of the columns without this data.

Here the result. The hana nodes have been installed with the fixed script, and the netweaver with the old one:

image


fabriziosestito (approved):

dottorblaster (approved):

nelsonkopliku (approved):
:heart:

PR #471, opened 2021-11-18
Added scrollbars to the codeblocks

### LINKED ISSUE #365: Check Catalog code blocks overflow too much and break the collapsable UX
As soon as a user expands a check with a big code block, the collapsable toggle gets pushed too far right and it needs a whole lot of scrolling before being able to interact with it with again.

Peek 2021-10-21 13-11

We should probably have overflow scroll bars around the code blocks, instead of the whole collapsable section.

MMuschner:
By adding scrollbars to the codeblocks, both verticaly and
horizontialy, the checks catalog should be more comfortable to use.

fix #365

### DIFF
--- a/web/frontend/stylesheets/stylesheets.scss
+++ b/web/frontend/stylesheets/stylesheets.scss
@@ -63,6 +63,12 @@ dt {
   pre code {
     color: $eos-bc-red-900;
   }
+  pre {
+    overflow-x: scroll;
+    overflow-y: scroll;
+    max-width: 80rem;
+    max-height: 30rem;
+  }
 }
 
 .nav-title {


dottorblaster (commented):
Thanks for the contribution, really appreciated!

Just a question, why the `max-height`?

MMuschner:
I experimented a bit and thought it would fit
scrollbars
.

dottorblaster (approved):
@MMuschner good job! LGTM

arbulu89 (commented):
@MMuschner @dottorblaster 
It looks nice! Maybe it would be better to create a new class with these styles, to avoid styling all the `pre` tags. Maybe by now we don't have others, but it would help to reduce issues in the future.

By the way, thank you for adding the gif!!!

dottorblaster:
@arbulu89 I actually agree but it's inside the `.ha-check-remediation` block, so the rule applies only in that context. Do you prefer to have a whole new class for that?

arbulu89:
> @arbulu89 I actually agree but it's inside the `.ha-check-remediation` block, so the rule applies only in that context. Do you prefer to have a whole new class for that?

@dottorblaster I didn't check that hehe
No, it is fine this way. Thanks @dottorblaster

arbulu89 (approved):

nelsonkopliku (approved):

PR #472, opened 2021-11-19
Identify trento installation

nelsonkopliku:
This PR introduces **Trento installation identification mechanism**

This becomes needed when publishing telemetry data to the telemetry service.

A random UUID is generated once at trento startup and it gets stored in the database for future usage in the system

That value would stay unchanged until the database and specifically that simple information, for any reason, gets cleaned up.
In that case we interpreted the scenario as a new installation.


dottorblaster:
There are many Trento, but which one is _the Trento_ :tm:?

web/services/installation.go
```diff
@@ -0,0 +1,39 @@
...
+type installer struct {
+	db *gorm.DB
+}
+
+func NewInstallerService(db *gorm.DB) Installer {
+	return &installer{db: db}
+}
+
+func (i *installer) InitializeIdentifier() error {
+	var trento entities.Trento
```
arbulu89, line 24: Why so many new lines in this methods hehe?
It looks strange that all lines are splited XD

Jokes apart, `gorm` offers 
- https://gorm.io/docs/advanced_query.html#FirstOrInit
- https://gorm.io/docs/advanced_query.html#FirstOrCreate
Maybe some of them would make the code prettier

web/services/installation.go
```diff
@@ -0,0 +1,39 @@
...
+
+type installer struct {
+	db *gorm.DB
+}
+
+func NewInstallerService(db *gorm.DB) Installer {
+	return &installer{db: db}
+}
+
+func (i *installer) InitializeIdentifier() error {
```
arbulu89, line 23: I think it would be nice to return the identifier in this method, just to print it or whatever reason

arbulu89 (commented):
@nelsonkopliku It looks good. Added some comments and questions before approving hehe

web/app.go
```diff
@@ -67,6 +67,7 @@ type Dependencies struct {
 	collectorService     services.CollectorService
 	clustersService      services.ClustersService
 	hostsNextService     services.HostsNextService
+	installationService  services.Installer
```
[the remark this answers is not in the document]
nelsonkopliku: ok, makes sense. Gonna do this.

web/app.go
```diff
@@ -99,11 +100,12 @@ func DefaultDependencies(config *Config) Dependencies {
 	sapSystemsService := services.NewSAPSystemsService(consulClient)
 	clustersService := services.NewClustersService(db, checksService)
 	collectorService := services.NewCollectorService(db, projectorWorkersPool.GetChannel())
+	intallationService := services.NewInstallerService(db)
 
 	return Dependencies{
 		consulClient, webEngine, collectorEngine, store, projectorWorkersPool,
 		checksService, subscriptionsService, hostsService, sapSystemsService, tagsService,
-		collectorService, clustersService, hostsServiceNext,
+		collectorService, clustersService, hostsServiceNext, intallationService,
```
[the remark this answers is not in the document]
nelsonkopliku: I see what you mean.

The alternative would be replacing in `web/app.go` this
```
if err := deps.installationService.InitializeIdentifier(); err != nil {
	log.Errorf("failed to initialize installation identifier: %s", err)
	return nil, err
}
```

with this
```
db, err := db.InitDB(config.DBConfig)
if err != nil {
	log.Fatalf("failed to connect database: %s", err)
}

installer := services.NewInstallerService(db)

if err := installer.InitializeIdentifier(); err != nil {
	log.Errorf("failed to initialize installation identifier: %s", err)
	return nil, err
}
```

and so duplicating the db initialization which is already happening in `DefaultDependencies`
```
db, err := db.InitDB(config.DBConfig)
if err != nil {
	log.Fatalf("failed to connect database: %s", err)
}
```

It might be not used elsewhere, yet I see it as a dependency for the App instantiation itself when we call `NewAppWithDeps`.

Any other option?

web/app.go
```diff
@@ -119,6 +121,7 @@ func MigrateDB(db *gorm.DB) error {
 		models.Tag{}, models.SelectedChecks{}, models.ConnectionSettings{}, models.CheckRaw{},
 		datapipeline.DataCollectedEvent{}, datapipeline.Subscription{}, models.HostTelemetry{},
 		entities.Cluster{}, entities.Host{}, entities.HostHeartbeat{}, entities.SlesSubscription{},
+		entities.Trento{},
```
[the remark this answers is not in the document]
nelsonkopliku: LOL. Deal.

web/services/installation.go
nelsonkopliku: makes sense :+1:

fabriziosestito (commented):
This looks good to me, but I have some doubts on the naming of the entity/service.
`Trento` sounds a bit generic and since @dottorblaster is working on something like `SystemSettings` maybe they could be the same Service?

web/services/installation.go
nelsonkopliku: I see, I'll remove some newlines :stuck_out_tongue_closed_eyes: 

In relation to `FirstOrInit` and `FirstOrCreate` I am not sure they fit the case :thinking: I might be wrong.

I understand those methods as: I give you this entity (Trento) with these values in it - in our case the InstallationID - and if it exists give it to me, otherwise create it.

The problem I see is that we don't know upfront the installation identifier we are looking for, that's what we are querying for.

I say: give me the very first top record of trento, if it's there I am done, otherwise I create a new instance, so next time it will exist and not recreate.

I tried something like
```
installationID := uuid.New()

db.FirstOrCreate(&trento, entities.Trento{InstallationID: installationID.String()})
```
but it clearly creates a new trento record all the times. That's not what we want I guess :joy: 

Again, I might be missing something about how that works.

nelsonkopliku:
> This looks good to me, but I have some doubts on the naming of the entity/service. `Trento` sounds a bit generic and since @dottorblaster is working on something like `SystemSettings` maybe they could be the same Service/share the same entity?
>
@fabriziosestito That's correct what you say, there's a relation with what @dottorblaster is on.
Will do a round of discussion with the team.

web/services/installation.go
arbulu89: I see, you could create other dummy primary key with a constant value maybe hehe
Or create the keys based in some seed.
Anyway, feel free to omit this comment, the current code should do the thing

web/app.go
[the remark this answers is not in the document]
arbulu89: Ok, let's go for it them!

dottorblaster:
Makes me laugh, I was just thinking "how do I manage the ID column", then I saw that basically the Trento :tm: entity overlaps with my SystemSettings work. :D Do you wanna name that so and I rebase against your work @nelsonkopliku?

web/services/installation.go
nelsonkopliku: Thanks for the tip, I'll keep those methods in mind :wink:

nelsonkopliku:
> SystemSettings
>
@dottorblaster sure thing! So my `Trento` becomes `SystemSettings` with an `InstallationID`?
Good to me! Deal?

dottorblaster:
@nelsonkopliku 

image

dottorblaster:
@nelsonkopliku

dottorblaster (approved):
SMASH THAT MERGE BUTTON

web/web_test.go
```diff
@@ -19,3 +22,11 @@ func setupTestConfig() *Config {
...
 	}
 }
+
+func newMockedSettingsService() services.SettingsService {
+	settigsService := new(services.MockSettingsService)
+
+	settigsService.On("InitializeIdentifier").Return(uuid.MustParse("59fd8017-b7fd-477b-9ebe-b658c558f3e9"), nil)
+
+	return settigsService
+}
```
fabriziosestito, line 32: @dottorblaster take a look, we'll need the same approach for the premium/telemetry middleware, by just adding a `On` clause to the mock

fabriziosestito (approved):

PR #486, opened 2021-11-22
remove the 'trento-' prefix from container artifacts

stefanotorresi:
Slipped through #459: the `trento-` prefix is already in the matrix values.


rtorrero (approved):
Woops, true, I just saw `trento-trento-runner` in the artifacts :-)

LGTM!

PR #487, opened 2021-11-22
Publish host telemetry to telemetry service

nelsonkopliku:
This PR introduces the ability to publish telemetry data to a remote service.

What comes with this:
- cleanup of the `repeat` function which was _repeated_ in two different places, now moved to `internal`
- refactor of the Agent and the Runner in order to use the common `internal.Repeat` replacement
- introduced an `Engine` for `Extracting` and `Publishing` telemetry related data
- implemented a `TelemetryPublisher` targeting the newly created telemetry collection service
- implemented the `HostTelemetryExtractor` to extract data from projected information to the needed model by the telemetry service
- Wired up with the application
- tests

**Notes**:
- for development purposes this engine is publishing to  [https://httpbin.org](https://httpbin.org/#/Anything/post_anything__anything_)
- have some doubts on endpoint naming and requests body in the collection service - going to followup on this to make the two sides coherent with each other

**Next**
- stabilize contract between trento and telemetry collection service
- find a way to set the `apiHost` for the telemetry service
- this implementation is not considering the scenario where `web` is scaled horizontally, so no locking on db so far. We'll tackle this is in following PRs.

**Plus**:
If we find it reasonable, we could also implement a `DatabasePublisher` as we have the service deployed, to publish data to a temporary table in trento database itself, just to be sure published data goes to a destination as we expect.
Otherwise we just wait for the service to be up and direct publishing traffic to this new guy

Nothing really blocking.


web/telemetry/registry.go
```diff
@@ -0,0 +1,10 @@
+package telemetry
+
+import "gorm.io/gorm"
+
+func NewTelemetryRegistry(db *gorm.DB) *TelemetryRegistry {
+	return &TelemetryRegistry{
+		"host_telemetry": NewHostTelemetryExtractor(db),
+		//"cluster": NewClusterTelemetryExtractor(),
```
dottorblaster, line 8: Leftover of showoff of this beautiful orthogonal design? :grimacing:

internal/utils.go
```diff
@@ -95,3 +97,23 @@ func CRC32hash(input []byte) int {
 	return int(crc32.Checksum(input, crc32Table))
 
 }
+
```
dottorblaster, line 100:
```suggestion
// We run the first tick immediately, then set the interval
```

dottorblaster (changes requested):
Astonishing! Some nitpicks but at a first glance I'm really confident this can land on `main`

internal/utils.go
dottorblaster: (This was a suggestion to leverage godoc for this, but I must admit I really suck with godoc itself)

web/telemetry/registry.go
nelsonkopliku: I'd call it _an example_ :innocent:

dottorblaster: :100:

internal/utils.go
```diff
@@ -95,3 +97,23 @@ func CRC32hash(input []byte) int {
...
+func Repeat(operation string, tick func(), interval time.Duration, ctx context.Context) {
+	// run the first tick immediately
```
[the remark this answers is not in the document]
nelsonkopliku: fixed

internal/utils.go
nelsonkopliku: fixed

dottorblaster (approved):
Brilliant! LGTM

fabriziosestito (approved):
LGTM! Great job, I already did a live review of this and  I found the code very clean and generic enough to be future proof for other telemetry needs.

We could store timestamped data in our db (e.g. collect checks selections and then send them in batch to the telemetry service on schedule).

runner/runner.go
```diff
@@ -226,24 +227,5 @@ func (c *Runner) startCheckRunnerTicker() {
...
-	for {
-		select {
-		case <-ticker.C:
-			tick()
-			log.Debugf("Next execution in %s", interval)
-		case <-ctx.Done():
-			return
-		}
-	}
+	internal.Repeat("runner.ara_playbook", tick, interval, c.ctx)
```
arbulu89, line 230: To be more precise `runner.ansible_playbook`

web/telemetry/host_telemetry.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"errors"
+	"time"
+
+	"github.com/trento-project/trento/web/models"
+	"gorm.io/gorm"
+)
+
+type HostTelemetry struct {
+	InstallationID string    `json:"installation_id"`
```
arbulu89, line 12: I would say that `InstallationID` and `AgentID` should go outside of this method, being them installation and agent generic. Otherwise, we will need to replicate them all over the telemetries

web/telemetry/host_telemetry.go
```diff
@@ -0,0 +1,61 @@
...
+
+type HostTelemetries []*HostTelemetry
+
+type HostTelemetryExtractor struct {
+	installationIdAwareExtractor
+	db *gorm.DB
+}
+
+func (ex *HostTelemetryExtractor) Extract() (interface{}, error) {
+	var collectedHostsTelemetry []models.HostTelemetry
```
arbulu89, line 30: I'm a bit confused with `models/entities` here. This `models.HostTelemetry` should be an entity in reality, right?
As it loads the data from the database, that later on should be transformed to a model (which in this case is the publishing data model).

web/telemetry/host_telemetry.go
```diff
@@ -0,0 +1,61 @@
...
+	var publishableHostTelemetries HostTelemetries
+
+	if err := ex.db.Find(&collectedHostsTelemetry).Error; err != nil {
+		return nil, err
+	}
+
+	if len(collectedHostsTelemetry) == 0 {
```
arbulu89, line 37: Are we sure we want to treat this as an error?
We might have some initial installation without attaching any host (strange, but could happen). Would this affect somehow?

web/telemetry/publisher.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/google/uuid"
+)
+
+type TelemetryPublisher struct {
+	apiHost    string
+	httpClient *http.Client
+}
+
+func (tp *TelemetryPublisher) Publish(telemetryName string, installationID uuid.UUID, extractedTelemetry interface{}) error {
+	endpoint := fmt.Sprintf("%s/api/collect/hosts", tp.apiHost)
```
arbulu89, line 20: I guess we could already pass this `hosts` endpoint part as variable.
Anyway, I see many questions just below. We can refine on the road

arbulu89 (approved):
Looks great!
We will continue refining on the road as we find the needs

web/telemetry/host_telemetry.go
fabriziosestito: This is what gets published to the telemetry collector, every "point" of a measurement has both installation id and agent id since it is atomic and it represents the telemetry of a single host in that precise moment. I wouldn't move this outside.

nelsonkopliku: Yup, this is going to be followed up, thanks.

web/telemetry/host_telemetry.go
nelsonkopliku: mh :thinking: I understand the feeling

`models.HostTelemetry` is what gets projected from the discoveries, that's why it went in `models`.
That package started as containing ReadModels.

I feel the need to clarify what we consider an `entity` and what a `model`/`readmodel` is
we started introducing `models` which we initially reasoned a bout as read models, then we introduced also entities.

I feel this is important to clarify ASAP as it defines the application layering as well and thus future architectural improvements.

Very shortly what I see is:
- `entities`: represent somehow the DomainModel, they have their own lifecycle, business logic and are stored somewhere. Possibly write Only (this might evolve in Aggregates, but not now)
- `readmodels`: the result of different things happening in the system, represents usually a use case (HostOverview, HostDetail, ClusterOverview...). In an upstreamdownstream relashionship usually the downstream defines its own readmodels as a result of things happening in the system ([@ziobrando](https://twitter.com/ziobrando) says it). A read model might or might not be the same representation that goes in the wire to the frontend let's say. In this case we could reason about `ViewModels` for instance...

It's a complex topic and I'd love to go deeper, refine and clarify as a team what they mean and when/how their supposed to be, not sure this is the right moment, though.

For this specific comment I'd be ok in moving `models.HostTelemetry` to `entities.HostTelemetry`, nothign against it.
@fabriziosestito @dottorblaster any thoughts on this?

web/telemetry/host_telemetry.go
nelsonkopliku: Returning an error from an `Extractor` affects the following publishing step, which wouldn't be executed.
That's by design

Look a [this line](https://github.com/trento-project/trento/pull/487/files#diff-7c76a17fd58e6a30285ef73042c129f98002c0edcfbd6fe8aed48274eddd67c5R55)
```
for telemetryName, extractor := range *e.telemetryRegistry {
	...
	extractedTelemetry, err := extractor.Extract()
	if err != nil {
		log.Errorf("Error while extracting telemetry %s: %s", telemetryName, err)
		continue
	}
        ... then publish
}
```
Does this answer and make sense? Am I overlooking anything?

web/telemetry/publisher.go
nelsonkopliku: Yep, we'll followup on those, thanks :wink:

web/telemetry/host_telemetry.go
arbulu89: My question is more about error treatment. Not having any record is not an error for me. It won't publish anything and that's all.
This is how I see it.
Anyway, so far it wouldn't change in anything the behaviour

nelsonkopliku: Okay, I see what you mean and it also makes sense.
Thanks for clarifying.

Returning an error is a way I found to communicate that for this extractor we do not need to actually publish anything.
Clearly there are other options.

Feel free to share some alternative options you see valid, so we can keep them into account next iterations.

arbulu89: I would personally do:
- If we don't have data, we send an empty list without an error
- In the `extractAndPublishFn` function, after checking if there was any error, I would check for the length of the received data. If 0, we continue the loop. Otherwise, we do what we do now.

For me it is clearer.
Anyway, as commented, both options are good enough.

nelsonkopliku: I see, that'd be valid.

However, `Extract` returns an `interface{}` so we don't really know what's in there, so first issue I see is that we are assuming that `Extract` method returns something I cal call `len` on, which might not be always the case :thinking: 

An hybrid approach would be:
- expecting that `Extract` still returns an `interface{}`, yet by design we define that one of the possible values for that return could be something of type `SkipPublishing`
- make the extractors return a `SkipPublishing` thing when they want to communicate that nothing should happen at publishing level for that extractor
- check in the engine the returned type from `Extract` method

```
type SkipPublishing struct{}

extractedTelemetry, err := extractor.Extract()
if _, ok := extractedTelemetry.(SkipPublishing); ok {
	continue
}

```
I don't dislike this either, but it seems too much.

Just to say I am open to change :wink: 
Will keep this under the radar. Thanks.

web/telemetry/host_telemetry.go
nelsonkopliku: Moved `models.HostTelemetry` to `entities.HostTelemetry`.
Eitherway it's the same for now. :stuck_out_tongue_closed_eyes:

PR #488, opened 2021-11-22
Project HANA Cluster Details

fabriziosestito:
Part of the consul removal saga,
projects the HANA cluster detail when projections clusters to a json b field for later consumption at service/controller level


nelsonkopliku (approved):
:eyes: another step forward :rocket:

web/datapipeline/clusters_projector.go
```diff
@@ -49,14 +53,17 @@ func transformClusterData(cluster *cluster.Cluster) (*entities.Cluster, error) {
 		return nil, fmt.Errorf("no cluster ID found")
 	}
 
+	clusterDetail, _ := parseClusterDetails(cluster)
+	fmt.Println(clusterDetail)
```
arbulu89, line 57: Leftover `Println`. We could leave it as debug in any case

web/datapipeline/clusters_projector.go
```diff
@@ -79,17 +86,17 @@ func detectClusterType(cluster *cluster.Cluster) string {
...
 	case hasSapHanaTopology && hasSAPHanaController:
-		return models.ClusterTypeScaleOut
+		return models.ClusterTypeHANAScaleOut
 	default:
 		return models.ClusterTypeUnknown
 	}
 }
 
-// getHanaSIDs returns the SIDs of the cluster
+// getClusterSIDs returns the SIDs of the cluster
```
arbulu89, line 97: Being honest, I think there was some misunderstanding here with this method.
As far as I know, the cluster only manages one principal SID (for scale up or scale out), and a possible cost-optimized SID.
So, 1 is always present, and the 2nd one only in cost-optimized scenarios.
As far as I know, scale out deployments only have 1 SID too.

I would go more for `SID` and `CostOptimizedSID` columns.

web/datapipeline/clusters_projector.go
```diff
@@ -49,14 +53,17 @@ func transformClusterData(cluster *cluster.Cluster) (*entities.Cluster, error) {
...
 
+	clusterDetail, _ := parseClusterDetails(cluster)
+	fmt.Println(clusterDetail)
 	return &entities.Cluster{
 		ID:          cluster.Id,
 		Name:        cluster.Name,
 		ClusterType: detectClusterType(cluster),
 		// TODO: Cost-optimized has multiple SIDs, we will need to implement this in the future
-		SIDs:            getHanaSIDs(cluster),
+		SIDs:            getClusterSIDs(cluster),
```
arbulu89, line 63: I know that this was already here, but the SID is exclusively part of SAP cluster types.
For example, a DRBD cluster won't have any SID.
Here, to be purist, the SIDs should go in the details. Anyway, I understand we will only manage SAP cluster in any case.
It was just an informative comment

web/datapipeline/clusters_projector.go
```diff
@@ -103,3 +110,229 @@ func getHanaSIDs(c *cluster.Cluster) []string {
 
 	return sids
 }
+
+// parseClusterDetails parses the cluster details depending on the cluster type
+func parseClusterDetails(c *cluster.Cluster) (json.RawMessage, error) {
+	switch detectClusterType(c) {
+	case models.ClusterTypeHANAScaleUp:
```
arbulu89, line 117: Wouldn't be more clear to have `case models.ClusterTypeHANAScaleUp, models.ClusterTypeHANAScaleOut` instead of this `fallthrough`?
Example: https://www.golangprograms.com/golang-switch-case-statements.html

web/entities/cluster.go
```diff
@@ -14,24 +15,120 @@ type Cluster struct {
...
+
+type ClusterResource struct {
+	ID        string `json:"id"`
+	Type      string `json:"type"`
+	Role      string `json:"role"`
+	Status    string `json:"status"`
+	FailCount int    `json:"fail_count"`
+}
+
+type ClusterNode struct {
```
arbulu89, line 44: This is strictly attached to a HANA cluster node. Maybe we could rename to `HANAClusterNode`. 
I'm saying this because a netweaver cluster will have different attributes

web/entities/cluster.go
```diff
@@ -14,24 +15,120 @@ type Cluster struct {
...
+type HANAClusterDetails struct {
+	SystemReplicationMode          string             `json:"system_replication_mode"`
+	SystemReplicationOperationMode string             `json:"system_replication_operation_mode"`
+	SecondarySyncState             string             `json:"secondary_sync_state"`
+	SRHealthState                  string             `json:"sr_health_state"`
+	CIBLastWritten                 time.Time          `json:"cib_last_written"`
+	StonithType                    string             `json:"stonith_type"`
+	StoppedResources               []*ClusterResource `json:"stopped_resources"`
+	Nodes                          []*ClusterNode     `json:"nodes"`
+	SBDDevices                     []*SBDDevice       `json:"sbd_devices"`
```
arbulu89, line 33: I would say that the `SBDDevices` should be referenced in the `Cluster` entity, and not in the details. All sort of clusters have sbd devices. Netweaver will have sbd devices

web/datapipeline/clusters_projector.go
```diff
@@ -103,3 +110,229 @@ func getHanaSIDs(c *cluster.Cluster) []string {
...
+	dateLayout := "Mon Jan 2 15:04:05 2006"
+	cibLastWritten, _ := time.Parse(dateLayout, c.Crmmon.Summary.LastChange.Time)
+
+	clusterDetail := &entities.HANAClusterDetails{
+		SystemReplicationMode:          systemReplicationMode,
+		SecondarySyncState:             secondarySyncState,
+		SystemReplicationOperationMode: systemReplicationOperationMode,
+		SRHealthState:                  srHealthState,
+		CIBLastWritten:                 cibLastWritten,
+		StonithType:                    getClusterFencingType(c),
```
arbulu89, line 153: Maybe better to use `parse` as method prefix in the methods where we parse the original raw data and convert it to the entity.

arbulu89 (commented):
@fabriziosestito Impressive work! Really really nice.

I have some observations with the data modelling and naming.

web/entities/cluster.go
fabriziosestito: Alright, but it is still part of the details field so we will duplicate the nesting also in the other clusters, i wouldn't put this in the generics fields

web/datapipeline/clusters_projector.go
fabriziosestito: Alright, but since we don't have the cost optimized i would just refactor to use one SID (get the first) and then we can iterate on this.

web/datapipeline/clusters_projector.go
fabriziosestito: basically they are here because they are used in filtering

rtorrero (approved):
LGTM

web/datapipeline/clusters_projector.go
fabriziosestito: https://github.com/trento-project/trento/issues/493

PR #494, opened 2021-11-23
Allow using `~` in the SSH key path used in the server installer.

### LINKED ISSUE #492: Tilde is not correctly resolved on installation script
# Situation

After I've created a public SSH key, I've started the installation script with:

```
curl -sfL https://raw.githubusercontent.com/trento-project/trento/main/install-server.sh | bash
```

After I've entered the private SSH key, I got this error messages:

```
Please provide the path of the runner private key: ~/.ssh/id_rsa
Installing trento-server on k3s...
Installing K3s...
Created symlink /etc/systemd/system/multi-user.target.wants/k3s.service → /etc/systemd/system/k3s.service.
Installing Helm...
Updating Helm dependencies...
Installing trento-server chart...
Release "trento-server" does not exist. Installing it now.
Error: failed parsing --set-file data: open ~/.ssh/id_rsa: no such file or directory
```

However, the file is there.

# Suggestion

It seems, the tilde isn't correctly resolved in the script. I would suggest to support this use case. In my case, `~` should be resolved to `/home/cloudadmin`.

tomschr:
This PR fixes #492.

The script fails if the user adds something like `~/.ssh/id_rsa` instead of a absolute path starting with `/home/...`. The fix replaces tilde by `$HOME`.

Maybe there are other places to replace this, but I haven't looked into that.

### DIFF
--- a/install-server.sh
+++ b/install-server.sh
@@ -24,6 +24,7 @@ EOF
 
 cmdline() {
     local arg=
+    local private_key_absolute_path=
     for arg
     do
         local delim=""
@@ -63,6 +64,14 @@ cmdline() {
     if [[ -z "$PRIVATE_KEY" ]]; then
         read -rp "Please provide the path of the runner private key: " PRIVATE_KEY </dev/tty
     fi
+
+    # Replace tilde with the current home:
+    PRIVATE_KEY="${PRIVATE_KEY/#\~/$HOME}"
+    private_key_absolute_path=$(realpath -q -e "$PRIVATE_KEY" || {
+         echo "Path '${PRIVATE_KEY}' to private SSH key does not exist, please try again."
+         exit 1
+    })
+    PRIVATE_KEY="$private_key_absolute_path"
 
     if [[ "$ROLLING" == "true" ]]; then
         TRENTO_VERSION="rolling"


install-server.sh
```diff
@@ -63,6 +63,8 @@ cmdline() {
     if [[ -z "$PRIVATE_KEY" ]]; then
         read -rp "Please provide the path of the runner private key: " PRIVATE_KEY </dev/tty
     fi
+    # Replace tilde with the current home:
+    PRIVATE_KEY="${PRIVATE_KEY/#\~/$HOME}"
```
fabriziosestito, line 67: @tomschr  thanks! I would use something like `PRIVATE_KEY="$(cd "$(dirname "$PRIVATE_KEY")"; pwd -P)/$(basename "$PRIVATE_KEY")"`  to make it even more generic and able to expand tilde and relative paths as well.

fabriziosestito (changes requested):

install-server.sh
tomschr: @fabriziosestito thanks for the reply!

To be honest, I don't really like the `cd` and `basename`, :wink: sorry. Two commands executed just to replace the tilde? Could we replace it with at least a single command?

After searching a bit, I found `realpath`:

```
$ realpath ~/.ssh/id_rsa
/home/tux/.ssh/id_rsa
```

You could even add the options `--logical` to resolve any `..` components or to resolve symlinks if you like. That would work with relative paths. Find all details in the [realpath man page](https://man7.org/linux/man-pages/man1/realpath.1.html).

With that, I would suggest:

```
PRIVATE_KEY=$(realpath "$PRIVATE_KEY")
```

Would that be more in line with your needs?

fabriziosestito: @tomschr `realpath` is sometimes not installed in bare-minimum systems

sionescu: @fabriziosestito /usr/bin/realpath is part of GNU coreutils. Do you mean that it's possible to have minimal systems without coreutils ?

fabriziosestito: @sionescu @tomschr  thanks! if it is in coreutils let's go for it, i had very outdated memories that it was packaged separately

sionescu: More precisely, `realpath` is part of coreutils since version 8.15, [released on 2012-01-06](https://savannah.gnu.org/forum/forum.php?forum_id=7076), so it's safe to depend on it.

tomschr: Ok, great! :+1: 

@fabriziosestito @sionescu There is only one question left: how do we deal with invalid user input?

If I'm not mistaken, the script does not have any checks yet, right? I think we should add that too to avoid non-existing paths.

According to the man page, the option `-e`/`--canonicalize-existing` checks that "all components of the path must exist". If not, the result will be empty and you get a return code of -1. We could either check for the return code or against the empty string. Any preferences?

Here is an example with checking against an empty string:

```bash
TMP=$(realpath -q -e "$PRIVATE_KEY")
if [ -z "$TMP" ]; then
   echo "Path '$PRIVATE_KEY' to private SSH key does not exist, please try again."
   exit 1
fi
PRIVATE_KEY="$TMP"
```

sionescu: We can check the command's return value, and since the interpreter is /bin/bash we can use the better-behaved operator "[[". Also, to improve code clarity, I suggest always delimiting variable references with curly braces, except for the one-letter built-ins.

```shell
TMP=$(realpath -q -e "${PRIVATE_KEY}")
if [[ $? > 0 ]]; then
   echo "Path '${PRIVATE_KEY}' to private SSH key does not exist, please try again."
   exit 1
fi
```

fabriziosestito: @tomschr this is perfect thanks! Just couple of nitpicks: I would just rename the `tmp` variable in something more meaningful `private_key_absolute_path` so we document what we are doing as well, and maybe change the message "to private SSH key does not exist, please provide a valid path."

tomschr: Thank you! I've combined the two suggestions from @sionescu and @fabriziosestito and pushed it.
Check it out and let me know if there are more things to change. Thank you! :+1:

install-server.sh
```diff
@@ -64,6 +65,14 @@ cmdline() {
         read -rp "Please provide the path of the runner private key: " PRIVATE_KEY </dev/tty
     fi
 
+    # Replace tilde with the current home:
+    private_key_absolute_path=$(realpath -q -e "$PRIVATE_KEY")
+    if [[ $? > 0 ]]; then
```
fabriziosestito, line 70: I tested it and this does not work since if the file is not present the `if` will never be reached
I would do it this way:

```bash
private_key_absolute_path=$(realpath -q -e "$PRIVATE_KEY") || { echo "Path '${PRIVATE_KEY}' to private SSH key does not exist, please try again."; exit 1; }
```

tomschr: Thanks @fabriziosestito for testing it. That's a bit surprising as I've tried it out in my shel first. Although I have to admit it was just this small snippet and not the whole script.

Anyway, I've added your line. I wasn't sure about line length so I've split it. Let me know if this is okay or if you prefer it in another way.

install-server.sh
```diff
@@ -64,6 +65,13 @@ cmdline() {
         read -rp "Please provide the path of the runner private key: " PRIVATE_KEY </dev/tty
     fi
 
+    # Replace tilde with the current home:
+    private_key_absolute_path=$(realpath -q -e "$PRIVATE_KEY" || {
+         echo "Path '${PRIVATE_KEY}' to private SSH key does not exist, please try again."
+         exit 1
+    })
+    PRIVATE_KEY="$private_key_absolute_path"
```
fabriziosestito, line 73:
```suggestion
    PRIVATE_KEY="${PRIVATE_KEY/#\~/$HOME}"
    private_key_absolute_path=$(realpath -q -e "${PRIVATE_KEY}") || {
        echo "Path '${PRIVATE_KEY}' to private SSH key does not exist, please try again."
        exit 1
    }
    PRIVATE_KEY="$private_key_absolute_path"
```

@tomschr almost there! I needed to combine your solution as well since the tilde wasn't being expanded with `realpath`.
Tilde expansion doesn't work if the string starts with quotes. The output of `read` is wrapped around quotes so this is the easiest way. You can test it by doing `realpath -q -e "~/.ssh/id_rsa"` vs `realpath -q -e ~/.ssh/id_rsa`. See also here: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_01

I tested it and this should cover all the cases.

fabriziosestito (changes requested):

install-server.sh
tomschr: Thanks @fabriziosestito! I've added now your code.
Interesting findings. I wasn't aware of the issue with the quote. Great you found out! :+1:

stefanotorresi:
closed via manual merge in a20fe9e

PR #497, opened 2021-11-24
Add EULA acceptance middleware and UI

dottorblaster:
If the user still didn't accept SUSE's EULA, we have to make him accept that to go forth. I came up with a middleware that essentially checks if the user still has to accept the EULA, and if not, presents the EULA.

I still have to add the ` if premium` condition.

About the final effect, a picture is worth a thousand words:

2021-11-24-173657_1302x1168_scrot


web/frontend/stylesheets/stylesheets.scss
```diff
@@ -205,3 +205,11 @@ td i.critical {
 .trento-pointer {
   cursor: pointer;
 }
+
+.align-right {
```
arbulu89, line 209: Strange that such a tipical things are not already loaded and available...

web/app.go
```diff
@@ -164,8 +164,10 @@ func NewAppWithDeps(config *Config, deps Dependencies) (*App, error) {
 	webEngine.Use(ErrorHandler)
 	webEngine.Use(sessions.Sessions("session", deps.store))
 	webEngine.StaticFS("/static", http.FS(assetsFS))
+	webEngine.Use(EulaMiddleware(deps.settingsService))
 	webEngine.GET("/", HomeHandler)
 	webEngine.GET("/about", NewAboutHandler(deps.subscriptionsService))
+	webEngine.GET("/accept-eula", EulaAcceptHandler(deps.settingsService))
```
arbulu89, line 170: Wouldn't be more appropriate a `POST` or `PUT` methods as we will record a value in the database?

web/eula.go
```diff
@@ -0,0 +1,21 @@
...
+
+func EulaAcceptHandler(settings services.SettingsService) gin.HandlerFunc {
+	return func(c *gin.Context) {
+		err := settings.AcceptEula()
+		if err != nil {
+			log.Error(err)
+			c.HTML(http.StatusInternalServerError, "error.html.tmpl", gin.H{"Error": "There was an error accepting the EULA. Please try again."})
+			return
+		}
+		c.Redirect(http.StatusFound, "/")
```
arbulu89, line 19: Maybe we could put toast message or an flash alert banner on the top of the home page once the Eula is accepted to give more feedback.

web/middlewares.go
```diff
@@ -0,0 +1,29 @@
+package web
+
+import (
+	"net/http"
+
+	"github.com/gin-gonic/gin"
+	"github.com/sirupsen/logrus"
```
arbulu89, line 7: I would keep the `log "github.com/sirupsen/logrus"` import to keep consistency across all the project

web/middlewares.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/trento-project/trento/web/services"
+)
+
+func EulaMiddleware(settings services.SettingsService) gin.HandlerFunc {
+	return func(c *gin.Context) {
+		eulaAccepted, err := settings.IsEulaAccepted()
+		if err != nil {
+			logrus.Error(err)
+			c.HTML(http.StatusInternalServerError, "error.html.tmpl", gin.H{"Error": "EULA Error :-("})
+			c.Abort()
```
arbulu89, line 17: Do we need these `aborts` here? I would think with the code in hand that the `c.HTML` is already returning

web/templates/eula.html.tmpl
```diff
@@ -0,0 +1,37 @@
...
+                  life of SAP Applications administrators</p>
+                <hr/>
+                <p>Trento is an open source (GPL-3.0) component that is supported as part of SUSE Linux Enterprise Server for SAP applications and therefore falls under the standard SUSE Terms and Conditions that can be reviewed at <a href="https://www.suse.com/products/terms_and_conditions.pdf">https://www.suse.com/products/terms_and_conditions.pdf</a>, SUSE EULA is available at <a href="https://www.suse.com/licensing/eula/#server">https://www.suse.com/licensing/eula/#server</a>.</p>
+
+                <p>By using Trento via a registered SUSE Linux Enterprise Server for SAP you agree to the Terms and Conditions also for Trento.</p>
+
+
+                <p>Trento collects the following KPIs on your environment and transmits it to SUSE:<p>
+                <ul>
+                  <li>SUSE Customer Center ID</li>
```
arbulu89, line 16: We don't store the `SUSE customer center ID` by now. Do we want to store this. It looks against privacy hehe

arbulu89 (approved):
@dottorblaster Approving the PR. I have left some comments that are worth checking (even though you decide to not act upon them hehe)

web/middlewares.go
arbulu89: Forget what I said, it looks a normal practice to stop the next handlings to be executed

web/middlewares.go
dottorblaster: You're right, I actually forgot that

web/frontend/stylesheets/stylesheets.scss
dottorblaster: Don't tell me that :smile: I completely agree

web/templates/eula.html.tmpl
dottorblaster: People high above me gave me this text, I pasted inside barely reading it. Maybe we can address this in a second moment, looks quite tricky to me :smile:

web/eula.go
dottorblaster: I tried to do that but actually I'm afraid both things are not viable:

- the flash alert won't be kept after the redirect, because the redirect makes the browser start a new request flow
- the toast message wouldn't live after the refresh, and we should do the whole thing with javascript to do that. I wanted to avoid javascript for such a simple feature.

web/app.go
dottorblaster: If the button was inside a form, yes, but the button to accept is a simple link, so the only way to do this is using a `GET`. I didn't think about the form, but it would be a lot of stuff in the eula page markup to achieve basically the same result.

web/templates/eula.html.tmpl
arbulu89: @stefanotorresi @lee-martin @diegoakechi What about this?

lee-martin: @arbulu89 @dottorblaster 

Below is the license agreement requested by PM and approved by legal, I'll forward internally as well.

Regarding SCC ID, isn't that actually a host detail? I mean each host could well have different SCC IDs?

"
License Agreement

Trento is an open source (GPL-3.0) component that is supported as part of SUSE Linux Enterprise Server for SAP applications and therefore falls under the standard SUSE Terms and Conditions that can be reviewed at https://www.suse.com/products/terms_and_conditions.pdf, SUSE EULA is available at https://www.suse.com/licensing/eula/#server .

By using Trento via a registered SUSE Linux Enterprise Server for SAP you agree to the Terms and Conditions also for Trento.

Trento collects the following KPIs on your environment and transmits it to SUSE:

-          SUSE Customer Center ID
-          Number of registered hosts in Trento
-          For each of those hosts:
o   SLES version
o   Number of sockets and total count of cores per host
o   Available memory per host
o   Hosting platform (Cloud service provider, Hypervisor or bare Metal)

Trento neither collects, processes or sends any personal data to SUSE.

By using Trento and updates available through SUSE channels you agree to these terms. In case you disagree, please switch to the OpenSource version of Trento.
"

arbulu89: @lee-martin No, the SCC ID is the ID you are given from SUSE to connect to SUSE customer center. It is of course private. With that ID, we could use their subscriptions.
I think that legal just re-phrased the content we just gave to them, and we might've given them this bullet list.
That's why we don't store it now. We could very well anonymize this data, but in this case, we wouldn't be storing the SCC ID itself.
At least, if it was myself, I wouldn't accept this EULA hehe

lee-martin: @arbulu89 I don't think anyone is referring to the SCC access token itself, but moreover SUSE customer ID. Let me check this, perhaps the expression "SCC ID" needs clarifying or rewording.

What I meant about the ID being a "host detail" is that we cannot assume that all hosts in Trento use the same SCC/customer ID, therefore we should surely consider the ID per host and not only once globally?

arbulu89: We don't store any SUSE customer ID, as I don't even know if this exist as such.
As an identifier, we auto-generate a random UUID based on the `/etc/machine-id` content.

web/templates/eula.html.tmpl
```diff
@@ -0,0 +1,37 @@
+{{ define "content" }}
+    <div class="mb-4">
+        <div class="row">
+            <div class="col-sm-8">
+                <h1 class='display-4 lead'>{{ .Title }}</h1>
+                <p class='subheadline'>An open cloud-native web console improving the
+                  life of SAP Applications administrators</p>
+                <hr/>
+                <p>Trento is an open source (GPL-3.0) component that is supported as part of SUSE Linux Enterprise Server for SAP applications and therefore falls under the standard SUSE Terms and Conditions that can be reviewed at <a href="https://www.suse.com/products/terms_and_conditions.pdf">https://www.suse.com/products/terms_and_conditions.pdf</a>, SUSE EULA is available at <a href="https://www.suse.com/licensing/eula/#server">https://www.suse.com/licensing/eula/#server</a>.</p>
```
stefanotorresi, line 9: It's actually Apache 2.0.

web/templates/eula.html.tmpl
stefanotorresi: I think we can remove the "SUSE Customer Center ID" line entirely.

stefanotorresi (changes requested):

web/app.go
fabriziosestito: I kinda agree that to be totally  RFC compliant, a GET should always be idempotent, while the POST usually has side-effects. One of the bad things that can generally happen is that some browsers trying to be smart and use prefetch of the links to cache nearest pages, this would mean that the EULA would be accepted without the user clicking on it.

web/eula.go
fabriziosestito: @dottorblaster  I can help  you with that I am sure we can do something about it, e.g. save the flash in session

web/templates/eula.html.tmpl
```diff
@@ -0,0 +1,37 @@
...
+                      <li>Number of sockets and total count of cores per host</li>
+                      <li>Available memory per host</li>
+                      <li>Hosting platform (Cloud service provider, Hypervisor or bare Metal)</li>
+                    </ul>
+                  </li>
+                </ul>
+
+                <p>Trento neither collects, processes or sends any personal data to SUSE.</p>
+
+                <p>By using Trento and updates available through SUSE channels you agree to these terms. In case you disagree, please switch to the Open Source version of Trento.</p>
```
stefanotorresi, line 30:
```suggestion
                <p>By using Trento Premium and its updates available through SUSE channels you agree to these terms. In case you disagree, please switch to the Community version of Trento.</p>
```

stefanotorresi (changes requested):
Moar disclaimer suggestions...

web/eula.go
arbulu89: We already have the flash message mechanism in place in `web/alert.go`

web/templates/eula.html.tmpl
stefanotorresi: @arbulu89 let's open a follow up to implement the SCC ID collection when available. We don't need to scrub it.

web/app.go
nelsonkopliku: > If the button was inside a form
>
@dottorblaster you're right and indeed we could wrap the button in a form that will only have that button, and have it submitted by POSTing.
Do you see that possible? :thinking:

web/eula.go
dottorblaster: If you folks meant `StoreAlert` I tried and nothing happened :smile:

web/templates/eula.html.tmpl
dottorblaster: Thank you so much, committed directly :sweat_smile:

web/app.go
dottorblaster: Yeah @nelsonkopliku, I didn't want to end up with a lot of markup for that but in the end...

@fabriziosestito I don't think that prefetching is something that affects us, but I totally see the point. I'll address this ASAP, thanks and thanks @arbulu89 for noticing it

web/templates/eula.html.tmpl
arbulu89: @stefanotorresi I don't think this value is collectable from within the machines. And again!!! Most of our users are PAYG users. They are not SUSE customers directly. This ID might not even exist

web/templates/eula.html.tmpl
```diff
@@ -0,0 +1,39 @@
...
+                <p>Trento collects the following KPIs on your environment and transmits it to SUSE:<p>
+                <ul>
+                  <li>SUSE Customer Center ID</li>
+                  <li>Number of registered hosts in Trento</li>
+                  <li>For each of those hosts:
+                    <ul>
+                      <li>SLES version</li>
+                      <li>Number of sockets and total count of cores per host</li>
+                      <li>Available memory per host</li>
+                      <li>Hosting platform (Cloud service provider, Hypervisor or bare Metal)</li>
```
stefanotorresi, line 23: We need to disclaim the collection of the following per-hosts data:
- List of running SAP instances
- Statistics about Trento configuration checks

stefanotorresi (changes requested):
@dottorblaster here is what's missing:

stefanotorresi (approved):

stefanotorresi:
we're gonna need changes to the `isPremium` conditional, but it's ok for now.

PR #500, opened 2021-11-26
Remove premium checks and their variables

arbulu89:
Remove the premium checks as we already have the `Premium` containers and code in place.

@stefanotorresi @lee-martin Is this the good moment to do this?


stefanotorresi:
Hmm, I'd like to have the premium version running in the demo environment first.

arbulu89:
> Hmm, I'd like to have the premium version running in the demo environment first.

Are we sure we want to have the `Premium` version on demo? It would expose already all the code

stefanotorresi:
> > Hmm, I'd like to have the premium version running in the demo environment first.
> 
> Are we sure we want to have the `Premium` version on demo? It would expose already all the code

Good question, let me ask upwards. I would personally go for it regardless, but maybe we need to remove the implementation from the catalog.

stefanotorresi:
All right, we can keep the demo as it is. No worries. Go forward with the latest list of premium checks we got from Diego.

arbulu89:
@stefanotorresi New changes based on the list provided by @diegoakechi done

stefanotorresi (approved):

PR #504, opened 2021-11-26
Add premium badge to the premium checks in catalog

arbulu89:
Add premium badge to the premium checks in the checks catalog.
**How to use**.
In order to make a check `Premium` we need to add the data in the its metadata. To do that we need to change the file `runner/ansible/roles/checks/my-check/defaults/main.tf` for the check, and include a `premium: True` in the file.
By default, the checks are not premium.

image

**Question**:
It only goes in the catalog by now. I can include it in the `Check selection` and `Checks results`. In the 1st one it gets pretty ugly, I would need to widen the whole modal (something I would do in any case). Should we add it in those places too?

### DIFF
--- a/runner/ansible/roles/post-metadata/tasks/store.yml
+++ b/runner/ansible/roles/post-metadata/tasks/store.yml
@@ -1,5 +1,15 @@
-- name: Load metadata
+- name: Load top level metadata
   include_vars: "{{ metadata_path }}"
+
+# This 2nd load is needed to get a fresh data set, without pre-loading already existing data.
+# Otherwise, we might have previous checks data here, as the first include_vars adds the variables
+# as top level (we still need that). As example, if the 1st check is premium and 2nd not, only
+# with the 1st include_vars, the premium value would be inherited, and we don't want that, otherwise,
+# we would need to define the premium entry in all the metadata files without having default options
+- name: Load specific metadata
+  include_vars:
+    file: "{{ metadata_path }}"
+    name: metadata_vars
 
 - name: Set metadata {{ id }}
   set_fact:
@@ -12,7 +22,8 @@
             'description': description,
             'remediation': remediation,
             'labels': labels,
-            'implementation': implementation
+            'implementation': implementation,
+            'premium': metadata_vars.premium|default(False)
           }]
         }, recursive=True, list_merge='append')
       }}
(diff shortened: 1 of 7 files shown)


rtorrero (approved):
Great idea, I like this. APPROVED!

fabriziosestito (approved):

dottorblaster (approved):
LGTM

rtorrero:
As to your questions: I would add it in the other sections too, but the full "Premium" badge might be a bit too much for the modals. Instead, we could use colors (if we identify a color with Premium by establish already in the checks catalog, it might be easier for the user to understand it) or maybe we could add a prefix to the checks title, e.g.: `Premium: Test if corosync.conf files are identical across all nodes `

nelsonkopliku (approved):

stefanotorresi:
The only thing I'm not convinced about this is the column where we're putting the badge.
Have you tried the description column instead of the id one?

arbulu89:
> The only thing I'm not convinced about this is the column where we're putting the badge. Have you tried the description column instead of the id one?
> 
> @rtorrero Hmmm, I don't quite like the idea of adding it as text elsewhere. The visual clue should be the same everywhere, i.e. green badge.

Well, it would look like exactly as it looks now, just moving the badge XD
I don't have any preference. It doesn't look better or worst in my opinion

arbulu89:
Hey @dottorblaster ,
Do you know any frontend trick to align the badge start width?
Now, as the chars length is not exactly the same, the badge is no perfectly aligned:
image

@stefanotorresi already propose me to use `pre`, but this completely changes the font which I don't like

dottorblaster:
@arbulu89 yeah you could have two different spans with something like a max-width, shall we look into this in a second moment? I find this idea very valuable

arbulu89:
@stefanotorresi @fabriziosestito @dottorblaster 
This is how it looks like if the badge goes in the description cell:
image

Does it look better?

dottorblaster:
It does actually look better to me, maybe we could give it a try with the premium tag before the description, but we can try further in a second moment.

arbulu89:
> It does actually look better to me, maybe we could give it a try with the premium tag before the description, but we can try further in a second moment.

This is how it looks like with the tag at the beginning:
image

arbulu89:
Finally, merging the the badge in the [description column](https://github.com/trento-project/trento/pull/504#issuecomment-981374574), in the right side. We can change if needed in the future.

PR #505, opened 2021-11-26
Add flavor to version package

rtorrero:
This PR adds a "flavor" constant that gets printed both in the "About" page and the `version` CLI subcommand. By default this constant has the "community" value for the community and will be overwritten in other variants (e.g. premium).

### DIFF
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -17,6 +17,6 @@ var versionCmd = &cobra.Command{
 	Short: "Print the version number of Trento",
 	Long:  `All software has versions. This is Trento's`,
 	Run: func(cmd *cobra.Command, args []string) {
-		fmt.Printf("Trento version %s\nbuilt with %s %s/%s\n", version.Version, runtime.Version(), runtime.GOOS, runtime.GOARCH)
+		fmt.Printf("Trento-%s version %s\nbuilt with %s %s/%s\n", version.Flavor, version.Version, runtime.Version(), runtime.GOOS, runtime.GOARCH)
 	},
 }
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -17,6 +17,6 @@ var versionCmd = &cobra.Command{
 	Short: "Print the version number of Trento",
 	Long:  `All software has versions. This is Trento's`,
 	Run: func(cmd *cobra.Command, args []string) {
-		fmt.Printf("Trento-%s version %s\nbuilt with %s %s/%s\n", version.Flavor, version.Version, runtime.Version(), runtime.GOOS, runtime.GOARCH)
+		fmt.Printf("Trento %s version %s\nbuilt with %s %s/%s\n", version.Flavor, version.Version, runtime.Version(), runtime.GOOS, runtime.GOARCH)
 	},
 }
--- a/version/flavor.go
+++ b/version/flavor.go
@@ -1,0 +1,3 @@
+package version
+
+const Flavor string = "Community"
--- a/web/templates/blocks/version.html.tmpl
+++ b/web/templates/blocks/version.html.tmpl
@@ -1,4 +1,4 @@
 {{ define "version" }}
     <a href="https://github.com/trento-project/trento/releases/tag/{{ .Version }}" target="_blank"
-       rel="noopener noreferrer">Trento v{{ .Version }}</a>
+       rel="noopener noreferrer">Trento-{{ .Flavor }} v{{ .Version }}</a>
 {{ end }}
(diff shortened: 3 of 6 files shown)


version/flavor.go
```diff
@@ -0,0 +1,3 @@
+package version
+
+const Flavor string = "community"
```
dottorblaster, line 3: :thinking: how do we replace this in Premium? `sed`?

dottorblaster (changes requested):

version/flavor.go
fabriziosestito: if we declare this as variable it could be replaced in the makefile, as we do for the version variable

stefanotorresi (changes requested):
I would enumerate the possible values as follows: `Community` and `Premium`. Capitalized.
Also I second Fabrizio's suggestion to make this a variable so that we can change it via ldflags.

version/flavor.go
rtorrero: In the premium version, as the `flavor.go` file just has this, it's a matter of adding this file in the premium repo but with a different content.

I'll give it a try with the Makefile, as long as it means we don't need to keep two separate Makefiles I'm also good with this approach

fabriziosestito: we could also use the same `version.go` file imho since it is already pretty anemic. wdyt?

rtorrero: Separating it in two files was suggested by @arbulu89 due to the way how we replace this in the premium repo (we basically overwrite the file we get from the community one). By having it separate we isolate how much code we are re-defining.

I'd like to keep it simple and avoid the ldflags change as it would mean that in order to build the premium binary, we'd need to adjust how we call `make` by adding a variable to the spec file. That would then mean having a separate spec file for the premium repo or using a %define in prjconfig in IBS which IMO is less visible and would not get applied when compiling trento-premium outside the repo.

rtorrero:
> I would enumerate the possible values as follows: `Community` and `Premium`, capitalized. Also I second Fabrizio's suggestion to make this a variable so that we can change it via ldflags.



> I would enumerate the possible values as follows: `Community` and `Premium`, capitalized. Also I second Fabrizio's suggestion to make this a variable so that we can change it via ldflags.

I updated the naming and capitalized it. About the suggestion from @fabriziosestito see the reply in the conversation

version/flavor.go
arbulu89: @rtorrero I'm starting to think as we will have 2 different packages (`trento` and `trento-premium`) we will already need a 2nd spec file. This will be needed most probably. In this case, we could just have a "forked" spec file, and do the magic there, having the community as default version.

arbulu89: And for @stefanotorresi @fabriziosestito @dottorblaster I'm afraid that you will need to start looking at the OBS system sooner or later, to realize that there are some constraints that will need have ugly patches XD

dottorblaster (approved):
LGTM

version/flavor.go
aleksei-burlakov: > @rtorrero I'm starting to think as we will have 2 different packages (trento and trento-premium) we will already need a 2nd spec file. This will be needed most probably. In this case, we could just have a "forked" spec file, and do the magic there, having the community as default version.

It makes sense, but this would give us already 3 sources of the code: trento/main, trento/premium-fork and trento-premium/master. There will be design discussions in January, I would postpone this PR until then.

nelsonkopliku (approved):

fabriziosestito (approved):

cmd/version.go
```diff
@@ -17,6 +17,6 @@ var versionCmd = &cobra.Command{
 	Short: "Print the version number of Trento",
 	Long:  `All software has versions. This is Trento's`,
 	Run: func(cmd *cobra.Command, args []string) {
-		fmt.Printf("Trento version %s\nbuilt with %s %s/%s\n", version.Version, runtime.Version(), runtime.GOOS, runtime.GOARCH)
+		fmt.Printf("Trento-%s version %s\nbuilt with %s %s/%s\n", version.Flavor, version.Version, runtime.Version(), runtime.GOOS, runtime.GOARCH)
```
stefanotorresi, line 20:
```suggestion
		fmt.Printf("Trento %s version %s\nbuilt with %s %s/%s\n", version.Flavor, version.Version, runtime.Version(), runtime.GOOS, runtime.GOARCH)
```

web/templates/blocks/version.html.tmpl
```diff
@@ -1,4 +1,4 @@
 {{ define "version" }}
     <a href="https://github.com/trento-project/trento/releases/tag/{{ .Version }}" target="_blank"
-       rel="noopener noreferrer">Trento v{{ .Version }}</a>
+       rel="noopener noreferrer">Trento-{{ .Flavor }} v{{ .Version }}</a>
```
stefanotorresi, line 3:
```suggestion
       rel="noopener noreferrer">Trento {{ .Flavor }} v{{ .Version }}</a>
```

stefanotorresi (changes requested):
Let's remove the hyphening.

stefanotorresi (approved):

PR #509, opened 2021-11-26
Move checks structs to entities and remove old checks in catalog

arbulu89:
While testing the Trento premium usage, I have noticed that if a check catalog entry is created, it is not never removed. This happened because I used the trento premium first, and the community version next with the same DB. The premium checks were still there.

During the fix, I have realized that I should move some of the data structs to entities. This is what this PR implements.

So, basically the next line is what fixes the bug, everything else is just model2entity work:
```
// Remove old not updated checks
return c.db.Not(&checkEntityList).Delete(entities.CheckList{}).Error
```

PD: Unfortunately, we cannot have `ToEntity` kind of functions, to convert models to entities, as this would require to have the enitity package imported in the models, which would create an infinite import loop. How much I love you golang... XD

### DIFF
--- a/web/entities/check.go
+++ b/web/entities/check.go
@@ -1,0 +1,41 @@
+package entities
+
+import (
+	"encoding/json"
+	"time"
+
+	"github.com/trento-project/trento/web/models"
+	"gorm.io/datatypes"
+)
+
+type Check struct {
+	ID        string `gorm:"primaryKey"`
+	CreatedAt time.Time
+	Payload   datatypes.JSON
+}
+
+type CheckList []*Check
+
+func (c *Check) ToModel() (*models.Check, error) {
+	var check models.Check
+	err := json.Unmarshal(c.Payload, &check)
+	if err != nil {
+		return nil, err
+	}
+
+	return &check, nil
+}
+
+func (c CheckList) ToModel() (models.ChecksCatalog, error) {
+	var checksCatalog models.ChecksCatalog
+
+	for _, checkRaw := range c {
+		check, err := checkRaw.ToModel()
+		if err != nil {
+			return nil, err
+		}
+		checksCatalog = append(checksCatalog, check)
+	}
+
+	return checksCatalog, nil
+}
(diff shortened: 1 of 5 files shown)


web/entities/check.go
```diff
@@ -0,0 +1,46 @@
...
+type Check struct {
+	ID        string `gorm:"primaryKey"`
+	CreatedAt time.Time
+	Payload   datatypes.JSON
+}
+
+type CheckList []*Check
+
+// TableName overrides the table name used by CheckRaw to `checks`
+func (Check) TableName() string {
```
nelsonkopliku, line 20: This shouldn't be strictly necessary as `Check` entity will be inflected by gorm to actually `checks`.
Yet, I don't see a problem in having it explicit as well.

web/services/checks.go
```diff
@@ -117,23 +107,37 @@ func (c *checksService) CreateChecksCatalogEntry(check *models.Check) error {
...
 }
 
 func (c *checksService) CreateChecksCatalog(checkList models.ChecksCatalog) error {
+
+	var checkEntityList entities.CheckList
 	for _, check := range checkList {
-		result := c.CreateChecksCatalogEntry(check)
-		if result != nil {
-			return result
+		// Unfortunately, with the current folders structure, we cannot have model to entity funcs
```
nelsonkopliku, line 122: Actually we should never have a `.ToEntity` method in a model, that would break [dependency rule](https://blog.cleancoder.com/uncle-bob/2012/08/13/the-clean-architecture.html) where higher level components - mdels in our case - do not depend/mention lower level components - entity, which is more of an infrastructural concern.

Maybe a function `toEntity(Model) Entity` unrelated from the model itself could be ok, but I am not completely sure :sweat_smile: 

I would remove this comment.

nelsonkopliku (changes requested):
Great! Just a tiny thing on a comment that I think we should remove in order to avoid confusion.

web/entities/check.go
arbulu89: You are right. Leftover from previous struct name `CheckRaw`

arbulu89:
@nelsonkopliku Changes done. Thank you for the review

nelsonkopliku (approved):
Super :rocket:

fabriziosestito (approved):

PR #514, opened 2021-11-29
Results dictionary should contain a timestamp now

### LINKED ISSUE #467: Add last execution time to the checks
Right now, it is not possible to know when the last round of checks was executed. This makes difficult to know if the runner is reporting new check results.
Having a last execution time would help us to know from the UI part if the runner has not been blocked for a while.

We could even add a execution duration time.

MMuschner:
This PR aims to add the `execution_datetime` of the checks execution to the result dictionary. This is useful to know when were the current checks execute. This is the first part of the implementation to fix #467.

### DIFF
--- a/runner/ansible/callback_plugins/trento.py
+++ b/runner/ansible/callback_plugins/trento.py
@@ -17,6 +17,7 @@ import yaml
 
 from ansible.plugins.callback import CallbackBase
 from ara.clients.http import AraHttpClient
+from datetime import datetime
 
 TRENTO_TEST_LABEL_KEY = "ara_playbook_labels"
 TRENTO_TEST_LABEL = "test"
@@ -278,6 +279,13 @@ class CallbackModule(CallbackBase):
         play = self.client.get("/api/v1/plays?uuid=%s" % self.play._uuid)
         playbook_id = play["results"][0]["playbook"]
         return playbook_id
+
+    def _return_timestamp(self):
+        """
+        Return a timestamp
+        """
+        timestap = datetime.now().strftime("%H:%M:%S")
+        return timestamp
 
     def _create_or_update_record(self, playbook, key, value, record_type):
         """


runner/ansible/callback_plugins/trento.py
```diff
@@ -279,6 +280,13 @@ def _get_playbook_id(self):
         playbook_id = play["results"][0]["playbook"]
         return playbook_id
 
+    def _return_timestamp(self):
+        """
+        Return a timestamp
+        """
+        timestap = datetime.now().strftime("%H:%M:%S")
+        return timestamp
+
```
dottorblaster, line 289: Aren't we suppose to use this function somewhere? Or it's just a conventional callback?

runner/ansible/callback_plugins/trento.py
```diff
@@ -279,6 +280,13 @@ def _get_playbook_id(self):
         playbook_id = play["results"][0]["playbook"]
         return playbook_id
 
+    def _return_timestamp(self):
+        """
+        Return a timestamp
+        """
+        timestap = datetime.now().strftime("%H:%M:%S")
+        return timestamp
```
fabriziosestito, line 288: typo: `timestamp`

arbulu89:
Hey @MMuschner ,
We will need a bit more than this hehe
- The obtained timestamp must be added to the dictionary that is sent to ara somewhere.
- Add the timestamp like `year-month-day hour:minutes:seconds`
PD: As always, try to add more information in the PR description. Something like
```
This PR aims to add the `execution_datetime` of the checks execution to the result dictionary. This is useful to know when were the current checks execute. This is the 1st part of the implementation to fix #467.
In next PRs, the current data will be used in the web part
```
Something like that. Keep in mind that the description is the entry point for the reviewer

runner/ansible/callback_plugins/trento.py
MMuschner: Just an entrypoint, babysteps

dottorblaster: Fair enough, I thought this was ready for the review, sorry :smile: if you want you can mark PRs as draft to make explicit that it's still an ongoing effort :+1:

arbulu89:
@MMuschner Other thing. Be careful linking issues to PRs. Linking means that once the PR is closed, the issue is closed as well.
This is useful when the PR completely fixes/implements the thing in the issue, but if you have partial implementations, it is a risky movement, as the issue will go away but it is not finished yet.
PD: I cannot unlink now the issue, maybe you can (or maybe it is a bug in GH side)

stefanotorresi:
It's been too long. Thanks for the effort anyway, Mike!

PR #516, opened 2021-11-29
Add changelogs in the OBS integration

rtorrero:
This PR implements two things in the CI:
- Makes sure that we also get updates on the "stable" package in OBS (devel:sap:trento/trento) while we where previously only getting them for the rolling rpms
- Adds changelogs for the rolling RPM. It's probably still not in the format that is ideal for OBS but since it's the rolling it should be fine. 

Note: The stable changelog one gets generated separately based on the release notes from GH which is also probably not what we need there, but it's better than nothing. We should look into it.

### DIFF
--- a/.github/workflows/ci-cd.yaml
+++ b/.github/workflows/ci-cd.yaml
@@ -338,14 +338,48 @@ jobs:
             ssh "$TRENTO_USER@$target_host" "sudo systemctl enable --now trento-agent.service"
           done
 
-  obs-commit:
-    needs: [test-binary, test-checks]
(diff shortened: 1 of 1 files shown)


rtorrero:
I've just noticed that it introduces a new variable in the CI/CD; I'll document it in an additional commit.

rtorrero:
Setting this to draft as I just spotted some things that I'd like to change before reviews.

rtorrero:
After a discussion with @arbulu89 we have decided to close this for now and address this in the future.

PR #517, opened 2021-11-29
Switch to hosts next

fabriziosestito:
Switches to the new hosts endpoint based on the projector systems.
Hosts related tags api endpoints have been also refactored using the new hosts service.
The old consul service is kept since it is still used in the sap system lists (not yet strangled).


nelsonkopliku (approved):
LGTMd!

dottorblaster (approved):

web/app.go
```diff
@@ -166,24 +166,23 @@ func NewAppWithDeps(config *Config, deps Dependencies) (*App, error) {
 	webEngine.StaticFS("/static", http.FS(assetsFS))
 	webEngine.GET("/", HomeHandler)
 	webEngine.GET("/about", NewAboutHandler(deps.subscriptionsService))
-	webEngine.GET("/hosts", NewHostListHandler(deps.consul, deps.tagsService))
-	webEngine.GET("/hosts-next", NewHostListNextHandler(deps.hostsNextService))
+	webEngine.GET("/hosts", NewHostListNextHandler(deps.hostsService))
```
arbulu89, line 169: Is intentional to keep the `Next` word here?

web/app.go
```diff
@@ -100,8 +100,8 @@ func DefaultDependencies(config *Config) Dependencies {
 	araService := ara.NewAraService(viper.GetString("ara-addr"))
 	checksService := services.NewChecksService(araService, db)
 	subscriptionsService := services.NewSubscriptionsService(db)
-	hostsService := services.NewHostsService(consulClient)
-	hostsServiceNext := services.NewHostsNextService(db)
+	hostsService := services.NewHostsConsulService(consulClient)
```
arbulu89, line 103: This naming is confusing as `hostsService` is used as `hostsConsulService` in the future and `hostsServiceNext` as `hostsService`.

arbulu89 (commented):
Coming late, but I have found some couple of things worth mentioning

web/app.go
fabriziosestito: thanks it is a leftover im going to address it in a separate PR

web/app.go
fabriziosestito: thanks like above i will clean it up in a separate PR

PR #519, opened 2021-11-30
Improve changes file generation

arbulu89:
Update the CI to achieve the next:
- Set the `changes` file format in the same format that `osc vc` creates. It is a small change, this highlights the date in green in obs: https://build.opensuse.org/package/view_file/devel:sap:trento/trento/trento.changes?expand=1
- Update the CI to only update the changes file in case of release. By now, we won't update it in the case of rolling as this requires much more work (we don't create the rolling tag in the same way than the stables one, and we would need some kind of additional logic to convert it to changes file mode)

**PD**: In order to start using the `changes` file correctly, I have updated the current changes file manually to have all of entries of all tags. This is the result: https://build.opensuse.org/package/view_file/devel:sap:trento/trento/trento.changes?expand=1
From now on, the release notes will be added on top of this content

### DIFF
--- a/.github/workflows/ci-cd.yaml
+++ b/.github/workflows/ci-cd.yaml
@@ -358,7 +358,12 @@ jobs:
         mkdir -p $HOME/.config/osc
         cp /root/.config/osc/oscrc $HOME/.config/osc
     - name: Prepare trento.changes file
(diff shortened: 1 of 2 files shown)


rtorrero (approved):
LGTM, thanks a lot for this @arbulu89

PR #524, opened 2021-11-30
Switch to projected HANA cluster detail view

fabriziosestito:
Switch the cluster detail view to the projected consul-free one, by adding the `GetByID` method to the cluster service and some minor improvement to the already existing Cluster model.
Minor bug fixed and some dead code removed, the old view is kept for the `Unknown` cluster type.

Some leftovers:
-  [ ] Refactor the tag api to use the new clusters service
- [ ] Refactor SIDs (plural) in just one SID in the cluster entity/model


web/app.go
```diff
@@ -170,7 +170,8 @@ func NewAppWithDeps(config *Config, deps Dependencies) (*App, error) {
 	webEngine.GET("/hosts/:name", NewHostHandler(deps.consul, deps.subscriptionsService))
 	webEngine.GET("/catalog", NewChecksCatalogHandler(deps.checksService))
 	webEngine.GET("/clusters", NewClusterListHandler(deps.clustersService))
-	webEngine.GET("/clusters/:id", NewClusterHandler(deps.consul, deps.checksService))
+	webEngine.GET("/clusters/:id", NewClusterHandler(deps.clustersService))
+	webEngine.GET("/clusters/:id/generic", NewClusterGenericHandler(deps.consul))
```
arbulu89, line 174: I understand that this is a temporary solution to split the consul based handlers with the new ones, right?
Even though it could be still useful once we remove consul to avoid having the `generic` suffix in the URLs in the templates, and manage the logic like this

web/templates/cluster_hana.html.tmpl
```diff
@@ -102,17 +98,15 @@
...
+            {{ template "sites" .Cluster.Details.Nodes.GroupBySite }}
         </div>
     </div>
     <hr>
 
     <h3>SBD/Stonith</h3>
-    {{- if .Cluster.IsFencingSBD }}
-        {{ template "sbd" .Cluster.SBD }}
-    {{- end }}
+    {{ template "sbd" .Cluster.Details.SBDDevices }}
```
arbulu89, line 107: I know that by now we always use SBD as fencing, but having the native Azure fencing system is quite normal in production systems. With this change, we would have an empty div after this `h3` header with empty values.
Not a big deal, but I just wanted to mention it

web/templates/cluster_hana.html.tmpl
```diff
@@ -25,54 +24,51 @@
...
                             {{ $badgeClass =  "badge-danger" }}
                         {{- end }}
-                        <span class="badge badge-pill {{ $badgeClass }} ml-0">{{ $secondarySyncState }}</span>
+                        <span class="badge badge-pill {{ $badgeClass }} ml-0">{{ .Cluster.Details.SecondarySyncState }}</span>
                     </div>
-
                     <div class="col-3 mt-5">
                         <strong>Stonith type:</strong><br>
-                        <span class="text-muted">{{ .Cluster.FencingType }}</span>
+                        <span class="text-muted">{{ .Cluster.Details.StonithType }}</span>
```
arbulu89, line 63: As a comment `Stonith` was renamed to `Fencing` due the offensive meaning of Stonith (political correctness). That's why currently it is used more. I know that we have stonith all over the places, but good to know why Fence word is being used lately more

arbulu89 (approved):
@fabriziosestito Great job as always!
I have noted some minor comments and questions.
Finally, I see that you completely removed the `clusters_test.go` file where the controller parts where tested. This means that we won't test the html rendering anymore for this page. Are we OK with this?

web/app.go
fabriziosestito: @arbulu89 yes this is temporary, it will be the same URL as before, i just wanted to split the two views because it was easier in the process

fabriziosestito:
> @fabriziosestito Great job as always! I have noted some minor comments and questions. Finally, I see that you completely removed the `clusters_test.go` file where the controller parts where tested. This means that we won't test the html rendering anymore for this page. Are we OK with this?

@arbulu89 mmh no the `clusters_test.go` is still there https://github.com/trento-project/trento/pull/524/commits/23469365af8d4fa2c6af553b949986b3fe635cd2 it was just under heavy refactoring

web/templates/cluster_hana.html.tmpl
fabriziosestito: ok, i will change it!

arbulu89:
> > @fabriziosestito Great job as always! I have noted some minor comments and questions. Finally, I see that you completely removed the `clusters_test.go` file where the controller parts where tested. This means that we won't test the html rendering anymore for this page. Are we OK with this?
> 
> @arbulu89 mmh no the `clusters_test.go` is still there [2346936](https://github.com/trento-project/trento/commit/23469365af8d4fa2c6af553b949986b3fe635cd2) it was just under heavy refactoring

Yes, my bad! I saw all red that my eyes told me that the file was removed hehe

web/templates/cluster_hana.html.tmpl
fabriziosestito: I will hide  the h3 + block if no SBDDevice is present

arbulu89 (approved):
Perfect! Thanks @fabriziosestito

rtorrero (approved):
LGTM, thanks!

PR #525, opened 2021-12-01
Show HANA details in its details page

arbulu89:
Show `HANA Database details` as title if the SAP system is a database:
image

PD: I didn't add any test to this, as i for see that this page must be completely refactored.


nelsonkopliku (approved):
Top

fabriziosestito (approved):

PR #526, opened 2021-12-01
Create properly the SAP system url in the host page

arbulu89:
Create properly the SAP system url in the host details page.
Right now, it was using the SID (`PRD`) instead of the internal system id to create the url.


fabriziosestito (approved):

abravosuse:
@arbulu89 @fabriziosestito I'm reviewing the results of the second test cycle and this issue still happens in our test environment with v0.6.0. How can I determine if this particular fix belongs to a later commit?

Thank you!

PR #527, opened 2021-12-01
Show SAP instance number using 2 digits

arbulu89:
Now SAP instance number in the SAP details page is shown with 2 digits:

image


fabriziosestito (approved):

dottorblaster (approved):
LGTM

PR #531, opened 2021-12-02
Refactor cluster SIDs in SID (model, entity, projector, service, handlers)

fabriziosestito:
When we will support cost optimized we will add a new field for the extra sids, as for now the cluster has just one main sid


dottorblaster (approved):
LGTM

arbulu89 (approved):
Nice!

Issue #507, opened 2021-11-26
Cluster detail view: checks selected but not showing in Health section


abravosuse:
This problem has been detected in the test environment that has been set up for the QA Team.

How to reproduce the problem:

1. Open the QA Trento Console URL.
2. Click on Pacemaker Clusters in the Menu section (or in the "Go to view" link in the Pacemaker Clusters section)
3. Filter by cluster Type = HANA scale-up
4. For any listed HANA cluster, click on the cluster name or the corresponding ID.
5. In the cluster detail view, click on the Settings button next to "Pacemaker Cluster Details".
6. In the hana_cluster window that opens up:
- Leave the Connection settings section as is (user cloudadmin has been enabled to connect via ssh from Trento Server to all hosts with Trento agents)
- In the Checks selection section, select some of the available checks.
- Click Save and go back and close the hana_cluster window.
7. Back to the cluster detail view, review the Health section on the top right corner: it shows 0 Passing, 0 Warning, 0 Critical.
8. Click on <Show check results>: a Health details window opens but it lists no checks.

abravosuse:
Step 8 above should read:
8. Click on Show check results: a Health details window opens but it lists no checks.

arbulu89:
It looks like back again the test `2C2D43` is giving troubles and blocking the whole ansible execution...
I will have a look on how we can fix it

arbulu89:
More information. The error only happens if the runner is executed within the container. And to make things worse, if you access the pod and execute the same exact command of the ansible-playbook or the trento runner itself, it works fine...
What a hell..

arbulu89:
Finally, adding `tty: true` in the trento runner container declaration in the helm chart fixes the issue.
Change here: https://github.com/trento-project/trento/pull/534/commits/62857f5ff67231ff9a837b22a9bb303c46ceb466

PR #534, opened 2021-12-02
Set a timeout in ansible tasks

arbulu89:
Enhancement of the runner to not block the whole ansible execution if some task gets blocked.
It workarounds the issue happening in: https://github.com/trento-project/trento/issues/507

With this hardening, even though the check is not properly implemented or it has some limitation that blocks the execution, the task will be set as failed and the ansible execution will continue.

As UX/UI, I have set the check as warning, and the error message provided by ansible as tooltip.

Here a short gif (in this case the check was blocked for some reason that I unknown yet). The used message is the one provide by ansible:
Peek 2021-12-02 15-24

After this I will continue with the work to fix the check `2C2D43` which is the root cause of the ansible execution blocking

**Edit**: Finally I have fixed the ansible execution blocking issue. It has been fixed adding `tty: true` to the container declaration

### DIFF
--- a/packaging/helm/trento-server/charts/trento-runner/templates/deployment.yaml
+++ b/packaging/helm/trento-server/charts/trento-runner/templates/deployment.yaml
@@ -46,6 +46,8 @@ spec:
           env:
             - name: CONSUL_HTTP_ADDR
               value: {{ .Release.Name }}-{{ .Values.dependencies.consul.name }}:{{ .Values.dependencies.consul.port }}
+          # tty is required to allow a correct ansible execution
+          tty: true
           securityContext:
             {{- toYaml .Values.securityContext | nindent 12 }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
--- a/runner/ansible/callback_plugins/trento.py
+++ b/runner/ansible/callback_plugins/trento.py
@@ -48,7 +48,8 @@ class Results(object):
                 "ABCDEF": {
                     "hosts": {
                         "host1": {
-                            "result": "passing"
+                            "result": "passing",
+                            "msg": "",
                         }
                     }
                 }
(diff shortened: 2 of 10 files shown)


packaging/helm/trento-server/charts/trento-runner/templates/deployment.yaml
```diff
@@ -46,6 +46,8 @@ spec:
           env:
             - name: CONSUL_HTTP_ADDR
               value: {{ .Release.Name }}-{{ .Values.dependencies.consul.name }}:{{ .Values.dependencies.consul.port }}
+          # tty is required to allow a correct ansible execution
+          tty: true
```
aleksei-burlakov, line 50: Very cool 🚀

runner/ansible/callback_plugins/trento.py
```diff
@@ -152,21 +163,25 @@ def v2_runner_on_ok(self, result):
         test_result = result._task_fields["args"]["test_result"]
         for group in task_vars["group_names"]:
             self.results.set_host_state(group, host, True)
+            if self.results.result_exist(group, task_vars[CHECK_ID], host):
+                continue
             self.results.add_result(group, task_vars[CHECK_ID], host, test_result)
 
-    def v2_runner_on_failed(self, result):
+    def v2_runner_on_failed(self, result, ignore_errors):
```
aleksei-burlakov, line 170: It looks like the `ignore_errors` is never used.

aleksei-burlakov (approved):
Looks awesome to me.

runner/ansible/callback_plugins/trento.py
arbulu89: Yes. This argument is coming from the placeholder function. Think about it as a golang interface.
We are just implementing the method, that it is defined somewhere else.
In our case, we don't use the argument, but It must go there

Issue #479, opened 2021-11-19
"Warning" value stays 0 even there is a problem with duplicated cluster name


Zaoliang:
see http://13.95.71.67/clusters

We have here duplicated cluster name: drbd_cluster, hana_cluster, Cluster Id are different.
So value of Warning '0' is wrong.

stefanotorresi:
It might also be worth nothing that the Alert icon should probably take the place of the grey disc at the beginning of the row.

abravosuse:
My two cents on this: for me the "issue" here is that the yellow warning sign informing about a duplicated cluster name might be interpreted as a cluster warning status. I agree with @stefanotorresi that placing the warning sign somewhere else might help but I would suggest choosing an different sign/symbol to notify about the duplicated name.

arbulu89:
First things first, as this is not an issue, it is more a UI/UX design topic. @stefanotorresi  Can we first change the labeling so we can have a clear picture of the current state of the project?

abravosuse:
@arbulu89 agreed 200% that this not a bug but an UI/UX enhancement.

stefanotorresi:
To be honest, I think this _is_ a bug.
If the condition of having duplicate cluster names is a warning, then the warning count should reflect that.

fabriziosestito:
@stefanotorresi we choose the warning icon a bit randomly for duplicated clusters without thinking this could appear as a warning, but we are considering only check results as  a warning state in the cluster details too.
I would vote to change the icon to something else, because it can really confuse users.

arbulu89:
> To be honest, I think this _is_ a bug.
> 
> If the condition of having duplicate cluster names is a warning, then the warning count should take that into account. If the users sees warning count 0, but then warnings in the rows, then something is wrong.
> 
> Why do you guys think this is not a bug?

Because it was implemented this way intentionally since the beginning. In the beginning, we didn't have the `id` column as we have now, so it was not really possible to know which cluster you were using, so to highlight that they were 2 in reality, we put the icon.
If you want, we can remove the icon completely.

stefanotorresi:
Yeah, maybe we should implement this as a check, somehow. Let's change the icon for now, so that this appears not as a warning, but something even softer, like a notice.

arbulu89:
@stefanotorresi @abravosuse Does this look OK by now?
![image](https://user-images.githubusercontent.com/36370954/144632924-79f73e6f-872f-4000-bd57-7f9ef9823d70.png)

[...]

PR #541, opened 2021-12-03
Use info icon in duplicated cluster names and sids

arbulu89:
**Edit**: Please, validate if the solution is the desired one. I have implemented this based on the issue [479](https://github.com/trento-project/trento/issues/479)

It basically changes the duplicated name warning icon by a info icon to be less invasive and have less option to confuse the user with the checks or cluster result.

Fix for: https://github.com/trento-project/trento/issues/479

image


fabriziosestito:
@arbulu89 Thanks much better like this ! 
I think tho that the teal color is a bit weird for an info tooltip, I would rather use a grey color after the name, wdyt?

stefanotorresi:
rebased and merged manually in 5bfa2ec99cb4399f4a5f9aba5f3d19b8f9d29951

PR #546, opened 2021-12-07
Add cluster settings api

nelsonkopliku:
This PR exposes an api at `GET /api/clusters/settings` to retrieve the clusters settings for the runner to provide to Ansible.

Involved settings are:
- selected checks for a cluster
- hosts for a given cluster with `address` and `user` to enable ansible connection

Example API response.
```
[
    {
        "id": "c115e7ecf5901d9a768d06a28ab0ba26",
        "selected_checks": [],
        "hosts": [
            {
                "name": "vmnetweaver01",
                "address": "x.x.x.x",
                "user": "root"
            },
            {
                "name": "vmnetweaver02",
                "address": "x.x.x.x",
                "user": "root"
            }
        ]
    },
    {
        "id": "5dfbd28f35cbfb38969f9b99243ae8d4",
        "selected_checks": [
            "A",
            "B",
            "C",
        ],
        "hosts": [
            {
                "name": "vmhana01",
                "address": "x.x.x.x",
                "user": "cloudadmin"
            },
            {
                "name": "vmhana02",
                "address": "x.x.x.x",
                "user": "cloudadmin"
            }
        ]
    }
]
```

This allows the runner to only call one API and provided useful information to Ansible and remove consul dependency from the runner itself (next PR).

### DIFF
--- a/cmd/agent/agent.go
+++ b/cmd/agent/agent.go
@@ -16,6 +16,7 @@ import (
 )
 
 func NewAgentCmd() *cobra.Command {
+	var bindIP string
 	var consulConfigDir string
 	var discoveryPeriod int
 
@@ -40,6 +41,10 @@ func NewAgentCmd() *cobra.Command {
 			return internal.InitConfig("agent")
 		},
 	}
+
+	startCmd.Flags().StringVar(&bindIP, "bind-ip", "", "Private address to which the trento-agent should be bound for internal communications. Reachable by the other hosts, including the trento server.")
+	startCmd.MarkFlagRequired("bind-ip")
+
 	startCmd.Flags().StringVarP(&consulConfigDir, "consul-config-dir", "", "consul.d", "Consul configuration directory used to store node meta-data")
 	startCmd.Flags().IntVarP(&discoveryPeriod, "discovery-period", "", 2, "Discovery mechanism loop period on minutes")
 
--- a/cmd/agent/agent.go
+++ b/cmd/agent/agent.go
@@ -16,7 +16,7 @@ import (
 )
 
 func NewAgentCmd() *cobra.Command {
-	var bindIP string
+	var sshAddress string
 	var consulConfigDir string
 	var discoveryPeriod int
 
@@ -42,8 +42,8 @@ func NewAgentCmd() *cobra.Command {
 		},
 	}
 
-	startCmd.Flags().StringVar(&bindIP, "bind-ip", "", "Private address to which the trento-agent should be bound for internal communications. Reachable by the other hosts, including the trento server.")
-	startCmd.MarkFlagRequired("bind-ip")
+	startCmd.Flags().StringVar(&sshAddress, "ssh-address", "", "The address to which the trento-agent should be reachable for ssh connection by the runner for check execution.")
+	startCmd.MarkFlagRequired("ssh-address")
 
 	startCmd.Flags().StringVarP(&consulConfigDir, "consul-config-dir", "", "consul.d", "Consul configuration directory used to store node meta-data")
 	startCmd.Flags().IntVarP(&discoveryPeriod, "discovery-period", "", 2, "Discovery mechanism loop period on minutes")
--- a/web/app.go
+++ b/web/app.go
@@ -186,6 +186,7 @@ func NewAppWithDeps(config *Config, deps Dependencies) (*App, error) {
 		apiGroup.POST("/clusters/:id/tags", ApiClusterCreateTagHandler(deps.clustersService, deps.tagsService))
 		apiGroup.DELETE("/clusters/:id/tags/:tag", ApiClusterDeleteTagHandler(deps.clustersService, deps.tagsService))
 		apiGroup.GET("/clusters/:cluster_id/results", ApiClusterCheckResultsHandler(deps.consul, deps.checksService))
+		apiGroup.GET("/clusters/settings", ApiGetClustersSettingsHandler(deps.clustersService))
 		apiGroup.POST("/sapsystems/:id/tags", ApiSAPSystemCreateTagHandler(deps.sapSystemsService, deps.tagsService))
 		apiGroup.DELETE("/sapsystems/:id/tags/:tag", ApiSAPSystemDeleteTagHandler(deps.sapSystemsService, deps.tagsService))
 		apiGroup.POST("/databases/:id/tags", ApiDatabaseCreateTagHandler(deps.sapSystemsService, deps.tagsService))
--- a/web/models/clusters_settings.go
+++ b/web/models/clusters_settings.go
@@ -1,0 +1,15 @@
+package models
+
+type ClusterSettings struct {
+	ID             string            `json:"id"`
+	SelectedChecks []string          `json:"selected_checks"`
+	Hosts          []*HostConnection `json:"hosts"`
+}
+
+type HostConnection struct {
+	Name    string `json:"name"`
+	Address string `json:"address"`
+	User    string `json:"user"`
+}
+
+type ClustersSettings []*ClusterSettings
--- a/web/services/clusters.go
+++ b/web/services/clusters.go
@@ -226,7 +226,7 @@ func (s *clustersService) GetAllClustersSettings() (models.ClustersSettings, err
 
 			hosts = append(hosts, &models.HostConnection{
 				Name:    host.Name,
-				Address: host.AgentBindIP,
+				Address: host.SSHAddress,
 				User:    username,
 			})
 		}
--- a/web/services/clusters.go
+++ b/web/services/clusters.go
@@ -5,7 +5,9 @@ import (
 	"errors"
 
 	"github.com/lib/pq"
+	log "github.com/sirupsen/logrus"
 	"github.com/trento-project/trento/internal"
+	"github.com/trento-project/trento/internal/cloud"
 	"github.com/trento-project/trento/web/entities"
 	"github.com/trento-project/trento/web/models"
 	"gorm.io/gorm"
@@ -20,6 +22,7 @@ type ClustersService interface {
 	GetAllClusterTypes() ([]string, error)
 	GetAllSIDs() ([]string, error)
 	GetAllTags() ([]string, error)
+	GetAllClustersSettings() (models.ClustersSettings, error)
 }
 
 type ClustersFilter struct {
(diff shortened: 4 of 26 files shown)


cmd/agent/agent.go
```diff
@@ -40,6 +41,10 @@ func NewAgentCmd() *cobra.Command {
 			return internal.InitConfig("agent")
 		},
 	}
+
+	startCmd.Flags().StringVar(&bindIP, "bind-ip", "", "Private address to which the trento-agent should be bound for internal communications. Reachable by the other hosts, including the trento server.")
+	startCmd.MarkFlagRequired("bind-ip")
```
stefanotorresi, line 46: This is very confusing. 
I see this configuration item only used when invoking `collector.Client.Publish()`, which issues a request towards the server.
So, as far as I can see, we're not actually binding to any network interface and, when Consul will be out of the picture, we won't have inbound traffic anymore and we won't need the agent host to be reachable at this IP address anymore. What gives?

Maybe we need to rethink the need to ask the user the ip address, because the server will inherently know this IP address from the request being issued by the agent collector client.

web/app.go
```diff
@@ -197,6 +197,11 @@ func NewAppWithDeps(config *Config, deps Dependencies) (*App, error) {
 		apiGroup.GET("/checks/catalog", ApiChecksCatalogHandler(deps.checksService))
 	}
 
+	internalApi := webEngine.Group("/api/internal")
+	{
+		internalApi.GET("/clusters/settings", ApiGetClustersSettingsHandler(deps.clustersService))
+	}
+
```
stefanotorresi, line 204: I don't think we shouldn't have an `internal` group on the `webEngine` GIN instance: we established in the past that internal APIs would go in what is currently called the `collectorEngine`, which has mTLS, and should be renamed to something like `internalEngine` soon enough.
I would suggest to put this endpoint in the `apiGroup` like all the others, for simplicity, and if we need to review grouping then let's see the global picture and rethink all of the endpoints.

web/models/clusters_settings.go
```diff
@@ -0,0 +1,15 @@
...
+	ID             string                     `json:"id"`
+	SelectedChecks []string                   `json:"selected_checks"`
+	Hosts          []*ConnectionInfoAwareHost `json:"hosts"`
+}
+
+type ConnectionInfoAwareHost struct {
+	Name    string `json:"name"`
+	Address string `json:"address"`
+	User    string `json:"user"`
+}
```
stefanotorresi, line 13: I think `ConnectionInfoAwareHost` is really hard to understand.
Hmm, maybe `ClusterSettings` just has `[]*HostConnection`. 

Then `HostConnection` will likely have additional fields for SSH keys in the future.

web/services/clusters.go
```diff
@@ -178,6 +180,63 @@ func (s *clustersService) GetAllTags() ([]string, error) {
...
+		for _, host := range cluster.Hosts {
+			username := "root"
+
+			hostConnectionSettings, found := connectionSettings[host.Name]
+			if found {
+				username = hostConnectionSettings.User
+			}
+
+			if username == "" {
+				username = "cloudadmin" // load user from cloud info
```
stefanotorresi, line 220: The default user being `cloudadmin` is an Azure-specific detail. Are we sure we can just hard-code this like that?

stefanotorresi (changes requested):

cmd/agent/agent.go
nelsonkopliku: > I see this configuration item only used when invoking collector.Client.Publish(), which issues a request towards the server.
>
mh :thinking: I am confused :sweat_smile: that is used in `./cmd/agent/config.go:52`
```
return &agent.Config{
    ...
    BindIP:          viper.GetString("bind-ip"),
    ...
}, nil
```
However.

> So, as far as I can see, we're not actually binding to any network interface 
>
Correct.

> and, when Consul will be out of the picture, we won't have inbound traffic anymore and we won't need the agent host to be reachable at this IP address anymore.
>
Partially correct I think, the runner needs to ssh to the hosts and it can be either via an IP or a resolved hostname.
When dealing with the hosts we usually have multiple IPs, yet I didn't find a way get the correct one for Ansible usage.
When installing via the `./install-agent.sh` we require a `--agent-bind-ip` which was used for consul, now we keep it and we use it for another purpose. 
Users installing the rpm have to put a `bind-ip: 0.0.0.0` in their `agent.yaml`, and this will be documented.

Do you have any other options? How can we deal with this differently, now? Maybe I am missing something :sweat_smile:

web/models/clusters_settings.go
fabriziosestito: I agree with @stefanotorresi `HostConnection` sounds good!

web/app.go
nelsonkopliku: Good to me.
shall we keep `/api/internal/clusters/settings` or do you prefer `/api/clusters/settings`?

fabriziosestito: I agree with this one, I would defer the `internalAPI` topic a bit and add it to the standard `apiGroup` for now

web/models/clusters_settings.go
nelsonkopliku: `HostConnection` SOLD!

web/services/clusters.go
nelsonkopliku: It is temporary, that would be loaded from cloud metadata if the host happens to be on a cloud environement.
It'll be fixed soon. :wink:

web/app.go
fabriziosestito: @nelsonkopliku `/api/clusters/settings` sounds good to me

nelsonkopliku: Deal!

cmd/agent/agent.go
stefanotorresi: > 
> mh thinking I am confused now sweat_smile that is used in [`./cmd/agent/config.go:52`](https://github.com/trento-project/trento/pull/546/files#diff-a22afc9279c2983d19ef8def93e25c50e43490a142d26b765296152a96c35d39R53)
> 
> ```
> return &agent.Config{
>     ...
>     BindIP:          viper.GetString("bind-ip"),
>     ...
> }, nil
> ```

Exactly, and that `BindIP` field in turn is never used to actually start any listening service bound to any network address. It's just sent as a field of a JSON to the server.

I wasn't clear in keeping SSH out of this picture: the SSH server is not bound to any particular IP address by the Agent process. How SSH is configured is out of the scope of our intervention; besides, the SSH server is usually not bound to any particular network address (i.e. 0.0.0.0:22), unlike Consul, which specifically needs a single interface.

My point is that the Agent process doesn't have to bind anything, because it's not a service, it's a client. SSH being there is a pre-existing condition we have no control over.

Yes, we can ask the user to specify which IP address to "advertise", but that's not a "bind" address, because it's actually not "bound" to anything by the Agent process.

nelsonkopliku: > Exactly, and that BindIP field in turn is never used to actually start any listening service bound to any network address. It's just sent as a field of a JSON payload to the server.
>
Okay, now its clearer what you mean.
That IP address is actually not used to _bind_ anything, it is published as part of the `HostDiscovery`.
We can change the name to something different.

> I wasn't clear in keeping SSH out of this picture: the SSH server is not bound to any particular IP address by the Agent process. How SSH is configured is out of the scope of our intervention...
>
> My point is that the Agent process doesn't have to bind anything, because it's not a service, it's a client. SSH being there is a pre-existing condition we have no control over.
>
I see, yet I need to answer the question: where should I ssh when running Ansible?

A little backgournd on the runner:
We provide an `Inventory` to Ansible for check execution, that is basically a list of hosts where to connect to and which checks to execute - nodes can also be grouped by cluster.

Here's a `Node`
```
type Node struct {
	Name        string // <-- hostname eg: vmhana01, vmnetweaver02
	AnsibleHost string <-- An IP, which? It comes to be the `BindIP` (better name to be found here)
	AnsibleUser string <-- ssh connection user
	Variables   map[string]interface{} <-- list of the checks to be executed (and maybe more, but not sure)
}
```
These Nodes get populated with the result obtained by calling `GET /api/clusters/settings` and are part of the `Inventory` used by ansible.

I don't know Ansible and I am not aware of the rules (is a `Name` always expected? What about the `AnsibleHost`? can it be empty? AnsibleUser too? Will ansible ssh to the hostname or the IP?) - and I need to do the safest thing, being requiring explicitly that IP to the user, as we already do (simply it will change semantic)

So what we can do is:
- reword `bind-ip` flag to something different (`ssh-host`, `connection-ip`, `runner-connection-hos`, whatever) pls suggest
- document the newly added flag

What do you think?

stefanotorresi: I think we can go on with the following: `bind-ip` can become `ssh-address`. 

I guess this will become the content of the `AnsibleHost` field.
I don't think we'll need the `Name` field in that struct, because I think that's currently there only because Consul's catalog reasons in terms of host names; Ansible doesn't: it needs just an address, which is not necessarily always just an IP, it could be a FQDN and it could have a `:port` suffix.

nelsonkopliku: Okay.

I made the change in [this commit](https://github.com/trento-project/trento/pull/546/commits/de55e1b65c6bcd15b18646210b4834094e1c3591) and it touches quite a lot of things.
Take a look and share your feelings.
@fabriziosestito @dottorblaster I'd like to know what you think.

We're completely changing the semantic from a more generic `bind-ip` to a really specific `ssh_address`.
If we're okay with this I'll just go ahead (refining a bit more the documentation)

Still, I feel that this change is too drastic from a semantic perspective, especially at this time. I am afraid it would add just more cognitive complexity, not sure what you think.

So here's the current definition we have for `--agent-bind-ip`
> the private address to which the trento-agent should be bound for internal communications.
> This is an IP address that should be reachable by the other hosts, including the trento server.
> 

To find a good naming we should start from that definition: what is that IP address we expect representing?

A variant based on what we have could be:
> the address to which the trento-agent should be reachable by the runner for check execution.
> 
Please feel free to contribute.

We could leave out the details of _how_ (ssh, http, whatever)
Or, if that `ssh` detail is really the key, then we go for it.

So, as a tradeoff, I see possible to change the meaning and the name to `check-execution-address`, `check-address`, `runner-check-address` or something similar, please help with naming :sweat_smile: 

Maybe `ssh_address` is just good enough, I am still not fully convinced, though.

fabriziosestito (approved):

cmd/agent/agent.go
fabriziosestito: @nelsonkopliku  imho `ssh-address` is good enough and enough explicit for now

web/app.go
```diff
@@ -186,6 +186,7 @@ func NewAppWithDeps(config *Config, deps Dependencies) (*App, error) {
 		apiGroup.POST("/clusters/:id/tags", ApiClusterCreateTagHandler(deps.clustersService, deps.tagsService))
 		apiGroup.DELETE("/clusters/:id/tags/:tag", ApiClusterDeleteTagHandler(deps.clustersService, deps.tagsService))
 		apiGroup.GET("/clusters/:cluster_id/results", ApiClusterCheckResultsHandler(deps.consul, deps.checksService))
+		apiGroup.GET("/clusters/settings", ApiGetClustersSettingsHandler(deps.clustersService))
```
arbulu89, line 189: I'm super confused with this.
We already have the `"/checks/:id/settings"` end point for this. Where the `id` is the ansible group id (as we need to group the ansible checks by group0.
This was named on purpose this way, as we were already seeing high changes to have other checks groups (like SAP system wide checks, or SAP landscapes wide checks)

arbulu89 (commented):
I have some hard feelings about did.
We already did a huge effort creating the `"/checks/:id/settings"` group of endpoint for this exact purpose.
This PR in my opinion totally overlaps this.
Did we consider simply adding the connection data information to the currently existing endpoints?
In essence, the information used in the runner, and shown in the frontend is exactly the same

cmd/agent/agent.go
nelsonkopliku: Ok for me keeping `ss-address` for now.

dottorblaster (approved):
Mega-LGTM

PR #548, opened 2021-12-08
Refactor sap system detail view

fabriziosestito:
Refactors the sap system detail view leveraging the projection,
uses the new server in sap systems tag api and cleans up consul dependencies


web/sap_systems.go
```diff
@@ -129,36 +127,33 @@ func NewHANADatabaseListHandler(sapSystemsService services.SAPSystemsService) gi
...
 			_ = c.Error(err)
 			return
 		}
 
 		// We will send the 1st entry by now, as only use the layout, which is repeated among all the
 		// SAP instances within a System. It does not resolve the HANA SR scenario in any case
-		c.HTML(http.StatusOK, "sapsystem.html.tmpl", gin.H{
-			"SAPSystem": systemList[0],
-			"Hosts":     systemHosts,
+		c.HTML(http.StatusOK, "sap_system.html.tmpl", gin.H{
```
arbulu89, line 153: We can remove the previous comments here, as we don't send anymore only the 1st system

web/services/sap_systems.go
```diff
@@ -50,6 +51,24 @@ func (s *sapSystemsService) GetAllApplications(filter *SAPSystemFilter, page *Pa
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	if len(instances) == 0 {
+		return nil, nil
+	}
+
+	return instances.ToModel()[0], nil
```
arbulu89, line 70: Why do we return only 1 item here?
Do we assume that all instances have the same data? (which in some cases is true, but as you are loading here the instance data...)

arbulu89 (approved):
Nice work!

I guess as a side work, we still need to remove all the consul part in the agent side, as we don't need to store it anymore. In a future PR I guess :)

PD: I thought we were removing/disabling some of the refactored pages in this PR

web/services/sap_systems.go
fabriziosestito: The `toModel` function takes a list of sap system instances (entities), groups them by ID, and returns a list of SAPSystems (models) by aggregating data.
Here since we specified that we want just one ID in the condition above, a list containing only one SAPSystem instance will be returned everytime.

fabriziosestito:
> Nice work!
> 
> I guess as a side work, we still need to remove all the consul part in the agent side, as we don't need to store it anymore. In a future PR I guess :)
> 
> PD: I thought we were removing/disabling some of the refactored pages in this PR

@arbulu89 we still have some leftover using consul, at the end we will remove completely the code from the agent.
This page didn't look so bad and the effort to port it to the new system was almost the same of removing the links around, so we decided to keep it.

nelsonkopliku (approved):
:eyes:

dottorblaster (approved):
YAS

Issue #385, opened 2021-10-27
Unify json formatting in communication payloads


nelsonkopliku:
As per this comment https://github.com/trento-project/trento/pull/371#discussion_r736564727 we need to unify the json formatting of the payloads we use in our APIs communication.

PR #552, opened 2021-12-10
Hosts UI revamp

dottorblaster:
An image is worth a thousand words:

2021-12-10-110245_1252x1381_scrot

Also, bonus point :star2: :arrow_right: this PR makes the single host UI page Consul-free.

image

### DIFF
--- a/web/hosts.go
+++ b/web/hosts.go
@@ -8,10 +8,7 @@ import (
 	consulApi "github.com/hashicorp/consul/api"
 	"github.com/pkg/errors"
 
-	"github.com/trento-project/trento/internal/cloud"
 	"github.com/trento-project/trento/internal/consul"
-	"github.com/trento-project/trento/internal/hosts"
-	"github.com/trento-project/trento/internal/sapsystem"
 
 	"github.com/trento-project/trento/web/models"
 	"github.com/trento-project/trento/web/services"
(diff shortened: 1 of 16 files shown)


web/hosts.go
```diff
@@ -127,35 +124,17 @@ func getTrentoAgentCheck(client consul.Client, node string) (*consulApi.HealthCh
...
+		host, err := hostsService.GetByName(name)
 		if err != nil {
-			_ = c.Error(err)
-			return
-		}
-
-		systems, err := sapsystem.Load(client, name)
-		if err != nil {
-			_ = c.Error(err)
+			_ = c.Error(NotFoundError("could not find host"))
```
fabriziosestito, line 133: The Hostname is not unique anymore, we need to change the service and the parameters (and the links around) basing everything on the ID

web/services/hosts.go
```diff
@@ -133,6 +135,36 @@ func (s *hostsService) GetAllBySAPSystemID(id string) (models.HostList, error) {
 	return hostList, nil
 }
 
+func (s *hostsService) GetByName(name string) (*models.Host, error) {
+	var host entities.Host
```
fabriziosestito, line 139: This is not really needed, the cloud privder logic can be squashed to `GetByID`
`GetByName` could have multiple results

fabriziosestito (changes requested):

web/models/host.go
```diff
@@ -18,6 +18,18 @@ type Host struct {
 	SAPSystems    []*SAPSystem
 	AgentVersion  string
 	Tags          []string
+	CloudData     interface{}
+}
+
+type AzureCloudData struct {
+	VMName          string `json:"vmname"`
+	ResourceGroup   string `json:"resourceGroup"`
```
nelsonkopliku, line 26: Aha! `camelCase` vs `snake_case` strikes again :trollface: 

Unfortunately we are not being consistent in the way we want `json` to look like.
And I am part of the problem, too :disappointed:, for instance in #546 I have used `snake_case` 

We did it differently throughout the code and I think that at a certain point we should sit, make an agreement on the style, and then commit to refactor all the codebase accordingly, which might not be trivial.
For reference: we have also an open issue for this #385 

I don't think it is the right time to tackle such a refactor, though we can find an agreement for the immediate future.

I'd like to keep #546 consistent with this PR and since I don't have strong preferences, I can fix it to make `camelCase` compliant. Would that be ok?

Then we can have a wider discussion and consider refactoring as well.

nelsonkopliku (approved):
Looks good!

fabriziosestito (approved):

web/models/host.go
fabriziosestito: @nelsonkopliku @dottorblaster in the entities/models i personally have used snake case everywhere, maybe we can keep this consistent at least for this layer and think about a clean-up later. wdyt?

nelsonkopliku: Ok, whatever it works better. Anyway we'd need to find an agreement on such things. Just not now :smile:

fabriziosestito: i like better the camel case onestly, we can refactor all at once laterz

nelsonkopliku: A bit late but ok for me.

PR #556, opened 2021-12-13
Use the %{name} macro in Provides

arbulu89:
Use the `%{name}` macro in the spec file.
This will help us to reuse this spec file in the trento-premium package, as we will only need to change the value in 1 place to change the package name.

I have done some minor editing as well


dottorblaster (approved):
LGTM

fabriziosestito (approved):

nelsonkopliku (approved):

PR #558, opened 2021-12-13
Make runner use clusters settings API instead of Consul

nelsonkopliku:
this PR cleans up the runner Inventory generation from Consul

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -15,6 +15,7 @@ import (
 type TrentoApiService interface {
 	IsWebServerUp() bool
 	GetChecksSettingsById(id string) (*webApi.JSONChecksSettings, error)
+	GetClustersSettings() (webApi.ClustersSettingsResponse, error)
 }
 
 type trentoApiService struct {
@@ -37,7 +38,7 @@ func (t *trentoApiService) getJson(query string) ([]byte, int, error) {
 
 	resp, err := t.httpClient.Get(t.composeQuery(query))
 	if err != nil {
-		return nil, resp.StatusCode, err
+		return nil, 0, err
 	}
 
 	defer resp.Body.Close()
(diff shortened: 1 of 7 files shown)


fabriziosestito (approved):

runner/inventory.go
```diff
@@ -2,23 +2,17 @@ package runner
...
-	"github.com/trento-project/trento/internal/consul"
-	"github.com/trento-project/trento/internal/hosts"
-
 	"github.com/trento-project/trento/api"
 )
 
 type InventoryContent struct {
 	Groups []*Group
-	Nodes  []*Node
+	Nodes  []*Node // this seems unused
```
arbulu89, line 15: Yes. This attribute is already there for any future circuntance where we would need to run ansible checks in ungrouped nodes

api/api.go
```diff
@@ -15,6 +15,7 @@ import (
 type TrentoApiService interface {
 	IsWebServerUp() bool
 	GetChecksSettingsById(id string) (*webApi.JSONChecksSettings, error)
+	GetClustersSettings() (webApi.ClustersSettingsResponse, error)
```
arbulu89, line 18: The `GetChecksSettingsById` is not used any longer, and it will be replaced in fact in some other PR?
Can we synchronize this to not have zombie methods here?

arbulu89 (approved):

api/api.go
fabriziosestito: @arbulu89 Good call, I will remove it in my PR

runner/inventory.go
rtorrero, line 15: Yes it does. What is this for?

rtorrero (approved):
LGTM!

runner/inventory.go
rtorrero: Saw it was asked already.

PR #561, opened 2021-12-13
Refactor cluster settings fe

fabriziosestito:
This PR removes consul completely from the web controllers by refactoring `/checks/:id/settings` leveraging the new methods on the cluster controller.
A follow-up cleanup is needed to nest the APIs under the clusters resource.


api/api.go
```diff
@@ -14,7 +14,6 @@ import (
 
 type TrentoApiService interface {
 	IsWebServerUp() bool
-	GetChecksSettingsById(id string) (*webApi.JSONChecksSettings, error)
```
nelsonkopliku, line 17: Ah great! Thanks. I should have cleaned up that.
Next to this we can also remove `api/checks.go` and `api/checks_test.go`, would you mind? :innocent:

web/checks_api.go
```diff
@@ -1,24 +1,18 @@
...
-	"github.com/trento-project/trento/internal/hosts"
 	"github.com/trento-project/trento/web/models"
 	"github.com/trento-project/trento/web/services"
 )
 
 type JSONChecksSettings struct {
 	SelectedChecks     []string          `json:"selected_checks" binding:"required"`
 	ConnectionSettings map[string]string `json:"connection_settings" binding:"required"`
-	Hosts              Nodes             `json:"hosts"`
+	Hostnames          []string          `json:"hostnames"`
```
nelsonkopliku, line 15: Do we need to also `make generate` to have swagger docs updated?

nelsonkopliku (commented):
That's great! Just a couple of minor things.

web/checks_api.go
fabriziosestito: yes you right pushing it now

nelsonkopliku (approved):
Lovely

dottorblaster (approved):
LGTM

PR #569, opened 2021-12-13
It's the final consul cleanup

nelsonkopliku:
No need for words.


fabriziosestito:
Let me find the right gif

fabriziosestito (approved):

PR #578, opened 2021-12-14
Reduce the agent discovery interval default value to 30 seconds

fabriziosestito:
We can reduce the discovery interval to 30 seconds now that locking is gone


dottorblaster (approved):
Absolutely right, let's see how it behaves but I think it will do great

nelsonkopliku (approved):
Aha! You also here guys? :smile:

PR #579, opened 2021-12-14
Revert count=1 flag

fabriziosestito:
For some reason, disabling test caching doesn't play well with our testify suite test, needs further investigation.
Reverting for now.


fabriziosestito:
Closed, found a proper solution

PR #580, opened 2021-12-14
Disable test parallelism

fabriziosestito:
After disabling the cache of tests in the make target, tests were failing in go `1.17`.
This due the added support of per-package parallelism,  which was breaking the database tests isolation.

This PR restores the caching mechanism to speed up test execution but disable parallelism to avoid concurrency in tests.


dottorblaster (approved):
LGTM

PR #581, opened 2021-12-15
Reduce discovery times to 10 seconds and do not store duplicated events

fabriziosestito:
This PR reduces the discovery time to 5 seconds and prevent the collector to store duplicated events.

If a duplicate is found, we keep sending it to the projectors in case a projection had failed before and we need to retry.
NOTE: the projectors skip old events by looking to their subscriptions.

### DIFF
--- a/web/services/collector.go
+++ b/web/services/collector.go
@@ -1,6 +1,9 @@
 package services
 
 import (
+	"errors"
+
+	log "github.com/sirupsen/logrus"
 	"github.com/trento-project/trento/web/datapipeline"
 	"gorm.io/gorm"
 )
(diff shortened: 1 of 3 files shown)


nelsonkopliku (approved):
Great

web/services/collector.go
```diff
@@ -19,11 +22,36 @@ func NewCollectorService(db *gorm.DB, projectorsChannel chan *datapipeline.DataC
...
+			}
+			c.projectorsChannel <- collectedData
+			return nil
+		}
+
+		if err != nil {
+			return err
+		}
+
+		log.Debugf("Event already exists. Agent ID: %s, DiscoveryType: %s ", collectedData.AgentID, collectedData.DiscoveryType)
```
dottorblaster, line 52: `log.Debugf` or `log.Warnf`?

dottorblaster (approved):
LGTM

web/services/collector.go
fabriziosestito: Debug i would say, this happens very often and we can't consider it a warning

nelsonkopliku (approved):

web/services/collector.go
dottorblaster: Ok!

rtorrero (approved):
LGTM

web/services/collector.go
```diff
@@ -19,11 +22,36 @@ func NewCollectorService(db *gorm.DB, projectorsChannel chan *datapipeline.DataC
...
+		c.projectorsChannel <- &event
```
stefanotorresi, line 53: I'm not quite clear, why are we projecting the last event?

fabriziosestito: We still do not have a mechanism of retrial or queue persistence in the projector system, so if the event wasn't projected for some reason we need to be sure that the projectors will receive the event and try again.
The projectors keep the state of the subscription so if the event was already projected they will just skip it.

PR #584, opened 2021-12-15
Add suse registry tag labels in the helm chart

arbulu89:
This tags are required during the helm chart creation in IBS/OBS. They simply set the version.

As this is the unique difference with the premium helm chart, I think having them is the easiest by now instead of just creating other `Chart.yaml` in the premium project only with these 3 lines of difference.

The version in `revx.y.z` must match the version of the helm chart itself 

PD: The version replacement doesn't work.. And in fact, apparently we don't need to put the trento package version there, so only the helm chart version and build version are needed


fabriziosestito (approved):

stefanotorresi:
since you changed it, you need to bump the version :)

PR #585, opened 2021-12-15
Remove hardcoded constant for the flavor to a Makefile variable

rtorrero:
Due to the recent changes that allow the trento-premium spec file to differ from the community one, it now makes more sense to set the flavor from the Makefile.

This PR does that.


nelsonkopliku (approved):

arbulu89 (commented):
Hey @rtorrero ,
Question: With this implementation if we do `FLAVOR=premium make build`, does this update the LDFLAGS properly?

Issue #582, opened 2021-12-15
Premium stuff.


stefanotorresi:
We need two functions to help implement the Premium/Community feature flags.

- `isPremiumFlavor()`
- `isPremiumActive()`

The first one is tied to the `flavor` constant hard-coded in the `flavor.go` file, which is different in IBS.

The second one is tied to the current implementation of the `SubscriptionService`, which relies on the Agent discovery
The `SubscriptionService` reports at least one SLES4SAP active subscription.

Consequences:
- The Premium EULA is only shown when `isPremiumFlavor()` returns `true`.
- The Telemetry is only sent when `isPremiumFlavor()` returns `true` and EULA is accepted.
- The Premium checks are only sent to the Runner when `isPremiumFlavor()` returns `true` and `isPremiumActive()` returns `true`. 
- The About page now shows the following (sort of):
![Screenshot from 2021-12-15 15-59-34](https://user-images.githubusercontent.com/2952427/146210111-e0544dfe-9235-4788-91ba-88e8c4e24982.png)
- The checks catalog and selection doesn't show Premium checks unless `isPremiumActive()` is true.

Follow-ups:
- Changes to EULA acceptance will not be handled for now.
- We might want to partially show Premium checks in the Premium flavor, even if Premium is not active, but just disallow selection.

PR #586, opened 2021-12-15
Add PremiumDetection capabilities

nelsonkopliku:
This PR adds PremiumDetection capabilities to:
- EULA requirement
- Telemetry Publishing
- Checks catalog
- Check settings

See issue https://github.com/trento-project/trento/issues/582


web/services/premium_detection.go
```diff
@@ -0,0 +1,65 @@
...
+		log.Warnf("Unable to determine whether the Trento Premium installation is active. Assuming it is not. Error: %s", err)
+	}
+	return isPremiumActive
+}
+
+func (premiumDetection *premiumDetection) isPremiumFlavor() bool {
+	return Premium == version.Flavor
+}
+
+func NewPremiumDetection(subscriptions SubscriptionsService, settings SettingsService) PremiumDetection {
```
arbulu89, line 60: This should return better a `*premiumDetection`.
And I would put it on the top of the file

web/services/premium_detection.go
```diff
@@ -0,0 +1,65 @@
+package services
+
+import (
+	log "github.com/sirupsen/logrus"
+	"github.com/trento-project/trento/version"
+)
+
+const Premium = "Premium"
+
+type PremiumDetection interface {
```
arbulu89, line 10: I would add the `Service` suffix here

web/services/premium_detection.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	isEulaAccepted, err := premiumDetection.settings.IsEulaAccepted()
+	if err != nil {
+		log.Warnf("Unable to determine whether the EULA for the Trento installation has been accepted. Assuming it is not. Error: %s", err)
+	}
+	return !isEulaAccepted, err
+}
+
+func (premiumDetection *premiumDetection) CanSendTelemetry() bool {
```
arbulu89, line 33: I guess, to follow the same structure in all methods, shouldn't we return `bool, error` ?

web/services/premium_detection.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	isEulaAccepted, err := premiumDetection.settings.IsEulaAccepted()
+	if err != nil {
+		log.Warnf("Unable to determine whether the EULA has been accepted for telemetry. Assuming it is not. Error: %s", err)
+	}
+	return isEulaAccepted
+}
+
+func (premiumDetection *premiumDetection) SupportsPremiumChecks() bool {
```
arbulu89, line 45: I would rename this to `IsPremiumActive`.
Most probably we will need to use it in other places rather than only the checks in the future

web/services/premium_detection.go
```diff
@@ -0,0 +1,65 @@
...
+	if !premiumDetection.isPremiumFlavor() {
+		return false
+	}
+	isPremiumActive, err := premiumDetection.subscriptions.IsTrentoPremium()
+	if err != nil {
+		log.Warnf("Unable to determine whether the Trento Premium installation is active. Assuming it is not. Error: %s", err)
```
arbulu89, line 51: I don't think we should make this assumptions, I would return the error here, and the consumer should decide what to do

arbulu89 (commented):
Nice work!
I find the core logic of the implementation good.
I have noted some small comments.

web/services/premium_detection.go
nelsonkopliku: done

web/services/premium_detection.go
nelsonkopliku: done

web/services/premium_detection.go
nelsonkopliku: done

web/services/premium_detection.go
nelsonkopliku: changed

web/services/premium_detection.go
nelsonkopliku: changed

web/app.go
```diff
@@ -70,6 +71,7 @@ type Dependencies struct {
 	settingsService      services.SettingsService
 	telemetryRegistry    *telemetry.TelemetryRegistry
 	telemetryPublisher   telemetry.Publisher
+	premiumDetection     services.PremiumDetectionService
```
arbulu89, line 74: I would add the `Service` suffix here (and in general, in the whole file)l
- `NewPremiumDetectionService`
- `premiumService`

web/services/checks.go
```diff
@@ -37,7 +38,7 @@ type ChecksService interface {
 	// Check catalog services
 	GetChecksCatalog() (models.ChecksCatalog, error)
 	GetChecksCatalogByGroup() (models.GroupedCheckList, error)
-	CreateChecksCatalogEntry(check *models.Check) error
+	CreateChecksCatalogEntry(check *models.Check) error // seems to be never used
```
arbulu89, line 41: We cuold remove it. I don't know if we are going to create individual check metadata entries at some point.
I guess with the comment is good enough for now, to now overload the PR

web/services/checks.go
```diff
@@ -55,12 +56,17 @@ type ChecksService interface {
...
+	premiumDetection PremiumDetectionService
 }
 
-func NewChecksService(araService ara.AraService, db *gorm.DB) *checksService {
-	return &checksService{araService: araService, db: db}
+func NewChecksService(db *gorm.DB, araService ara.AraService, premiumDetection PremiumDetectionService) *checksService {
+	return &checksService{
+		db:               db,
+		araService:       araService,
+		premiumDetection: premiumDetection,
```
arbulu89, line 68: I would rename it to `premiumDetectionService`

web/services/checks.go
```diff
@@ -69,7 +75,17 @@ Checks catalog services
...
 func (c *checksService) GetChecksCatalog() (models.ChecksCatalog, error) {
 	var checksEntity entities.CheckList
-	result := c.db.Order("payload->>'name'").Find(&checksEntity)
+	isPremiumActive, _ := c.premiumDetection.IsPremiumActive()
+
+	var result *gorm.DB
+	qb := c.db.Order("payload->>'name'")
+
+	if isPremiumActive {
+		result = c.db.Find(&checksEntity)
```
arbulu89, line 84: This should be `qb.Find`, right?

arbulu89: In fact, we could reduce this if/else condition to:
```
result = qb.Find(&checksEntity, datatypes.JSONQuery("payload").Equals("premium", isPremiumActive))
```

fabriziosestito (approved):

web/services/premium_detection.go
```diff
@@ -0,0 +1,72 @@
+package services
+
+import (
+	log "github.com/sirupsen/logrus"
+)
+
+const (
+	Premium   = "Premium"
+	Community = "Community"
```
stefanotorresi, line 9: I don't think these constants should be public, should they?

web/services/checks.go
```diff
@@ -266,12 +282,35 @@ func (c *checksService) GetSelectedChecksById(id string) (models.SelectedChecks,
 		SelectedChecks: []string{},
 	}
 
-	result := c.db.Where("id", id).First(&selectedChecks)
-	if errors.Is(result.Error, gorm.ErrRecordNotFound) {
-		return selectedChecks, nil
+	catalog, err := c.GetChecksCatalog()
```
arbulu89, line 285: I find this new method over complicated.
Is not there any sql join operation we could use?
Maybe join the selected checks with catalog and filter by premium

arbulu89 (commented):
Hey @nelsonkopliku ,
Nice job.
I miss having the `Service` suffix through the code, as we do with the other services.
Besides that, I added some comments that are worth checking

web/telemetry/engine.go
```diff
@@ -46,6 +48,16 @@ type Publisher interface {
...
+	canPublishTelemetry, err := e.premiumDetection.CanPublishTelemetry()
+	if err != nil {
+		log.Errorf("Unable to start Telemetry Engine. Error: %s", err)
+		return
+	}
+	if !canPublishTelemetry {
+		log.Infof("Telemetry publishing is not supported by this installation")
+		return
+	}
+
```
stefanotorresi, line 60: I'm very confused by this. It looks like a one-shot check. Is there anything ensuring that the telemetry engine is started once the `premiumDetection.CanPublishTelemetry()` changes result?

web/services/premium_detection.go
```diff
@@ -0,0 +1,72 @@
...
+	if !premiumDetection.isPremiumFlavor() {
+		return false, nil
+	}
+	isEulaAccepted, err := premiumDetection.settings.IsEulaAccepted()
+	if err != nil {
+		log.Errorf("Unable to determine whether telemetry can be published. Error: %s", err)
+		return false, err
+	}
+	return isEulaAccepted, nil
+}
```
stefanotorresi, line 56: Can't help but noticing that these two methods are basically identical except a `not` operator, but hey.

stefanotorresi (changes requested):

web/app.go
nelsonkopliku: changed.
I find the suffix a bit of redundant, buy I don't mind, it's just details. 
There are other critical things :smile:

web/services/checks.go
nelsonkopliku: kept with the comment

web/services/checks.go
nelsonkopliku: done

web/services/checks.go
nelsonkopliku: > This should be qb.Find, right?
>
good catch! thanks.

> In fact, we could reduce this if/else condition to:
> ```
> result = qb.Find(&checksEntity, datatypes.JSONQuery("payload").Equals("premium", isPremiumActive))
> ```
>
For this I see an issue :thinking: 
doing like that means that we either load checks which are premium OR checks not premium, but in reality what we need to do is either load only the community checks or load all of the checks, premium and not premium.

```
if isPremiumActive {
	result = qb.Find(&checksEntity)
} else {
	result = qb.Find(&checksEntity, datatypes.JSONQuery("payload").Equals(false, "premium"))
}
```

Makes sense?

web/services/checks.go
nelsonkopliku: Yes, you are right.
Yet the overcomplication comes from the way the models and entities were designed.
That needs to be followed up and maybe refactored in another PR, I agree.

web/services/premium_detection.go
nelsonkopliku: ok, changed. thanks

web/telemetry/engine.go
nelsonkopliku: good catch. 
Moving the checking step inside the repeat loop.
This way if `premiumDetection.CanPublishTelemetry()` changes result it will be detected and trento will start publishing telemetry.

web/services/premium_detection.go
nelsonkopliku: They are expressing two different domain invariants, and a little bit of duplication in favour of explicitness I find it acceptable.

web/services/checks.go
arbulu89: Yes, makes sense. I guess that there is a more sqliish way of doing it, but well, let's move forward XD

arbulu89 (approved):

web/services/premium_detection.go
stefanotorresi: fair enough 😉

rtorrero: I'd keep the whole function in `RequiresEulaAcceptance` and do `return !premiumDetection.RequiresEulaAcceptance` in `CanPublishTelemetry`

web/services/checks.go
stefanotorresi: let's bring this home first 🚀

stefanotorresi (approved):

web/services/premium_detection.go
stefanotorresi: nah, Nelson is right, these are two different domain concepts, if we couple them together it's gonna bite us back eventually, I was hinting at maybe find a common domain invariant, but it's good enough as it is.

PR #587, opened 2021-12-16
Fix SBD checks 0B6DB2 and 49591F remediation rendering

arbulu89:
Fix `0B6DB2` and `49591F` checks remediation rendering.

image


rtorrero (approved):
LGTM

PR #588, opened 2021-12-16
Fix attached database instances hydration

fabriziosestito:
Fixes the retrieval of the attached database instances


nelsonkopliku (approved):

arbulu89 (approved):
Looks great @fabriziosestito

fabriziosestito:
> Looks great @fabriziosestito To test is in a real environment, simply check in the SAP systems list, if any of the systems has 2 HANA entries

@arbulu89 I tested against a copy of the demo db and it works fine!

PR #591, opened 2021-12-16
"About premium" improvements

rtorrero:
This PR addresses the new template changes requested in issue #582. The other changes should have been addressed in PR #586

This is how it looks now for both cases when no subscriptions are found and when there are e.g. 2 subscriptions found.
2-found
none-found

In both cases there is now a tooltip 
tooltip

stating the number of subscriptions found. This is not exactly what the issue wanted but what I understood was the intention from the conversation we had yesterday.

### DIFF
--- a/web/about_test.go
+++ b/web/about_test.go
@@ -52,8 +52,7 @@ func TestAboutHandlerPremium(t *testing.T) {
 
 	assert.Equal(t, 200, resp.Code)
 	assert.Contains(t, minified, "About")
-	assert.Regexp(t, regexp.MustCompile("<dt.*>Subscription</dt><dd.*>Premium.*</dd>"), minified)
-	assert.Regexp(t, regexp.MustCompile("<dt.*>SLES_SAP machines</dt><dd.*>2.*</dd>"), minified)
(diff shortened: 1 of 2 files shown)


arbulu89 (commented):
With the changes @nelsonkopliku is doing, I don't know if we will still have the current `IsPremium` function. Most probably, we will need to change the backend handler to use the new service, but I'm not sure.

In the other hand, the fronted looks good, but didn't we agree to have an info icon for the tooltip?

rtorrero:
> With the changes @nelsonkopliku is doing, I don't know if we will still have the current `IsPremium` function. Most probably, we will need to change the backend handler to use the new service, but I'm not sure.
> 
> In the other hand, the fronted looks good, but didn't we agree to have an info icon for the tooltip?

Yes, the info icon is probably a good hint for the user that there is a tooltip there, adding it now. About the IsPremium, I'll have a quick chat with @nelsonkopliku Thanks!

nelsonkopliku (approved):

fabriziosestito (approved):

PR #593, opened 2021-12-16
follow official product guidelines

stefanotorresi:
change the copywrite a little and fix margin and html quoting. (never use `'`, always use `"`, we need a linter for this)


stefanotorresi:
jeez we really need a proper html dom testing library.

arbulu89 (approved):
I would appreciate a screenshot of the update to see how it looks like

stefanotorresi:
before:
Screenshot from 2021-12-17 11-34-14

after:
Screenshot from 2021-12-17 11-32-00

arbulu89:
Thanks @stefanotorresi

PR #594, opened 2021-12-16
remove repetition in the checks description

stefanotorresi:
`s/Test if//`


runner/ansible/roles/checks/1.3.2/defaults/main.yml
```diff
@@ -4,7 +4,7 @@ name: 1.3.2
 group: SBD
 labels: generic
 description: |
-  Test if the SBD_STARTMODE is set to always
+  The SBD_STARTMODE is set to always
```
arbulu89, line 7: I would remove the `The` here

runner/ansible/roles/checks/1.3.3/defaults/main.yml
```diff
@@ -4,7 +4,7 @@ name: 1.3.3
 group: SBD
 labels: generic
 description: |
-  Test if the SBD service is enabled
+  The SBD service is enabled
```
arbulu89, line 7: I would remove `The` here

runner/ansible/roles/checks/1.3.6/defaults/main.yml
```diff
@@ -4,7 +4,7 @@ name: 1.3.6
 group: SBD
 labels: generic
 description: |
-  Test if the SBD msgwait timeout value is two times the watchdog timeout
+  The SBD msgwait timeout value is two times the watchdog timeout
```
arbulu89, line 7: I would remove `The` here

arbulu89 (approved):

rtorrero:
In general, I agree with Xabi that we could also remove "The" from the description of a few tests. But after reading it, I feel that the text lacks a bit of context; see e.g. "SBD msgwait timeout value is two times the watchdog timeout". It is unclear if the SBD msgwait timeout *should*  be two times the watchdog timeout of if it *currently is* while it shouldn't. 

What about re-writing these using "should" in the description? That way, on a failed check scenario, the user is able to understand without a doubt what the cause of the failed check is.

arbulu89:
> In general, I agree with Xabi that we could also remove "The" from the description of a few tests. But after reading it, I feel that the text lacks a bit of context; see e.g. "SBD msgwait timeout value is two times the watchdog timeout". It is unclear if the SBD msgwait timeout _should_ be two times the watchdog timeout of if it _currently is_ while it shouldn't.
> 
> What about re-writing these using "should" in the description? That way, on a failed check scenario, the user is able to understand without a doubt what the cause of the failed check is.

I personally don't like the `should` option here.
They are checks. It is obvious (at least to me) that they check what the say, not that the created description is the reality of the checks result. And they expect that the things being True.
I understand this like any programming language function naming: `IsTrue`. I understand that this aims to check if the checked value is True. The same with our descriptions.

arbulu89:
@stefanotorresi In fact, after checking it again, I would remove the `The` word from all the descriptions.
What do you think?

stefanotorresi:
I kinda agree with you both.

I have thought about the `should` or `must` wording, which is more conversational and easier to read, but then we would need to establish a clear convention for all the checks, e.g. `should` is used in checks that would only yield a warning, while `must` would be used for those that are critical.

Then again, I don't want to go through this right now: I just wanted to introduce a little improvement, so I'm gonna follow Xabi's suggestion, which doesn't open doors for ambiguity, although we'll keep the wording a little bit weird and mechanical.

stefanotorresi:
I went down a rabbit hole and ended up with changing a few things in the catalog:
Screenshot from 2021-12-17 14-37-28

PR #595, opened 2021-12-16
Move About under Settings

stefanotorresi:
...and change some icons.

before:
Screenshot from 2021-12-16 19-10-34

after:
Screenshot from 2021-12-16 19-08-26


arbulu89 (approved):
I preferred to have the `About` page in the main tree

stefanotorresi:
I disagree 😝. It also makes more sense to have more than 1 item under Settings, otherwise it's kinda pointless.

PR #599, opened 2021-12-17
Revert collector deduplication

fabriziosestito:
This was causing performances problem due to the big jsonb field comparison in the QA env with 26 nodes.
We need to find a better solution (hashing/caching), reverting the premature optimization for now.

### DIFF
--- a/web/services/collector.go
+++ b/web/services/collector.go
@@ -1,9 +1,6 @@
 package services
 
 import (
-	"errors"
-
-	log "github.com/sirupsen/logrus"
 	"github.com/trento-project/trento/web/datapipeline"
 	"gorm.io/gorm"
 )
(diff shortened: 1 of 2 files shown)


nelsonkopliku (approved):

arbulu89 (approved):

arbulu89:
@fabriziosestito Question.

fabriziosestito:
> @fabriziosestito Question. How is this affecting to the reading of models? I thought that they were totally independent as this storing operation is async. Does this query slow down the complete database?

Yes it was affecting database performances, it is a heavy query done every often, keep in mind also that  the database we deploy in k8s is not production grade. We can improve this with a cache or by hashing the json file so we don't have to query by it. My big mistake was believing that comparing the two jsonb fields would be not an heavy operation.

PR #600, opened 2021-12-17
Upgrade from eos 2.5.1 to 5.0.0

rtorrero:
This PR upgrades trento to use the latest available version of the EOS design system. Since this also comes with a bump in the bootstrap version, some issues are expected.


rtorrero:
Removing the draft flag as it *seems* to be working decently enough.

web/templates/blocks/health_container.html.tmpl
```diff
@@ -2,16 +2,16 @@
     <h5>Health</h5>
     <div class="health-container{{ if eq .Layout "horizontal" }} horizontal-container{{ end }}">
         <div class="alert alert-inline alert-success">
-            <i class="eos-icons eos-18 alert-icon">check_circle</i>
+            <i class="eos-icons-outlined eos-18 alert-icon">check_circle</i>
```
arbulu89, line 5: Is there some reason you are using the `outlined` version here?
How does it look like?
Could you upload a screenshot?

The same for the badges. I would like to see them working before approving hehe

arbulu89 (commented):
Hey @rtorrero ,
Could you upload some gif/screenshots to see how it looks like?

dottorblaster (approved):
LGTM :heart_eyes:

rtorrero:
woops! I planned to add the screenshots when removed the draft flag but I forgot. Here I go:
Notice the difference in the fonts and the outlined icons in the menu (which now follows the eos design guidelines). Also notice that the feedback for a selected section in the menu is also slightly different; following also the EOS design template.
menu-collapsed

I also took the opportunity to use this from the EOS template to make the alert counts look a bit better:
label

dottorblaster:
Tables look good as well.

2021-12-21-165049_1334x535_scrot

2021-12-21-165133_1505x668_scrot

nelsonkopliku (approved):
LGTM.
the links feel just a little too bold, but it might be the screenshots.

PR #604, opened 2021-12-17
Fix cluster details resource issues

arbulu89:
WIP as I want to add some additional tests

Fix the HANA cluster page sync state and health state fields.
Besides that, I have added a HANA failed state to show it in the sites. Otherwise, it could be missleading as it is common to have 2 Primary sites once the failover is done.

Here some pics:
image
image
image
image


fabriziosestito (approved):
Great! Thanks

nelsonkopliku (approved):

PR #607, opened 2021-12-17
Accept relative paths in install-server.sh private key parameter

fabriziosestito:
Follow up PR of #494 which was missing some stuff


stefanotorresi:
what are all these whitespaces?! Is your editor picking up `.editorconfig` correctly?

PR #626, opened 2021-12-20
Paginate correctly sap systems

arbulu89:
The sap systems list page was not paginating properly, as it was paginating based on the instances number rather than the systems number, which was hiding some instances within visualized systems.

This implementation limits the pagination to the systems instead of instances.

Besides this:
- It includes the pagination controls in the sap systems page.
- Sorts the SAP systems before showing them in the UI to have always the same order based in sid and instance number

image

### DIFF
--- a/web/entities/sap_system.go
+++ b/web/entities/sap_system.go
@@ -2,6 +2,7 @@ package entities
 
 import (
 	"time"
+	"sort"
 
 	"github.com/lib/pq"
 	"github.com/trento-project/trento/web/models"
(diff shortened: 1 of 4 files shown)


web/entities/sap_system.go
```diff
@@ -84,5 +85,13 @@ func (s SAPSystemInstances) ToModel() []*models.SAPSystem {
 	for _, sapSystem := range set {
 		sapSystems = append(sapSystems, sapSystem)
 	}
+	sortBySid(sapSystems)
+
```
fabriziosestito, line 89: nitpick, we are using `SID` capital everywhere

nelsonkopliku (approved):
LGTM

dottorblaster (approved):
LGTM

fabriziosestito (approved):

fabriziosestito:
@arbulu89 good call!

Issue #447, opened 2021-11-12
The fencing timeout check fails if the time unit is set


diegoakechi:
The fencing timeout parameter defaults to seconds, and therefore can be set without informing the unit. When the using the unit, pacemaker persists it on the CIB as a string containing the unit, like `144s`. The check defaults to `144` and does not consider such scenario, reporting a false failure in such case. 

https://github.com/trento-project/trento/blob/main/runner/ansible/vars/azure_env.yml#L13-L14

The check needs to be adjusted to be more resilient in such cases.

PR #639, opened 2021-12-21
Fix test 373DB8 to pass if timeout unit is set

arbulu89:
Fix for https://github.com/trento-project/trento/issues/447

Now the checks runs a regular expression to consider the `s` unit besides the value.

PD: 
@diegoakechi I guess this issue doesn't only happen in this check, as the usage of the unit is accepted in pacemaker. I guess that this situation is possible in any time based parameter.


diegoakechi:
@arbulu89 yes, you are right. In the future we might need to abstract it on a proper Ansible module (if none exists yet) to make simpler the checks building.

PR #640, opened 2021-12-21
Point the telemetry service to telemetry.trento.suse.com

dottorblaster:
As above :sweat_smile:

### DIFF
--- a/web/telemetry/publisher.go
+++ b/web/telemetry/publisher.go
@@ -40,7 +40,7 @@ func (tp *TelemetryPublisher) Publish(telemetryName string, installationID uuid.
 	return nil
 }
 
-var apiHost = "https://telemetry.trento-project.io" // this service still does not exist
+var apiHost = "https://telemetry.trento.suse.com"
 
 func NewTelemetryPublisher() Publisher {
 	return &TelemetryPublisher{


web/telemetry/publisher.go
```diff
@@ -40,7 +40,7 @@ func (tp *TelemetryPublisher) Publish(telemetryName string, installationID uuid.
 	return nil
 }
 
-var apiHost = "https://telemetry.trento-project.io" // this service still does not exist
+var apiHost = "https://telemetry.trento.suse.com" // this service still does not exist
```
arbulu89, line 43: Hey, @dottorblaster ,
I would remove the comment besides the new domain, as now it will exists hehe

arbulu89 (approved):

web/telemetry/publisher.go
nelsonkopliku, line 43: We can remove the comment I guess, that service is being spawned up as I understood.

nelsonkopliku (approved):

web/telemetry/publisher.go
arbulu89: By the way, now I notice, `apiHost` looks a quite inaccurate name for this, doesn't it? XD

dottorblaster: Fair 'nuff

nelsonkopliku: Bad naming here is my fault :sweat_smile: 
Feel free to change or not to something like `telemetryApiHost` `telemetryEndpoint` or whatever makes sense.

dottorblaster: Comment removed, thanks for that! `apiHost` was quite clear to me but feel free to suggest something else, I think this line is going to be edited a lot to allow configuration via build flags in the future so... :smile:

stefanotorresi: 👍🏼 for renaming. I suggest `telemetryServiceUrl`.

Issue #573, opened 2021-12-14
Duplicate database after unregistration and registration process


abravosuse:
Trento version: v0.6.0
Trento environment: QA testing cycle 2
Description: following an issue with HANA system replication in the Trento testing environment, I had to unregister and register the HANA secondary database of one of the SAP systems. The cluster is now working fine but when I access the Trento console and go to the HANA Databases view, I see two databases with the same SID. One of them is obsolete, the other one is up to date. I have restarted the Trento agents in both nodes but that didn't fix the issue. See attached screenshot.

![image](https://user-images.githubusercontent.com/91731013/145978081-018e98c6-ba9f-4e28-8c47-07a1fbc1e803.png)

fabriziosestito:
Need to be confirmed on the `rolling` version, it might be already fixed

abravosuse:
v0.7.0: Same issue. HANA database listed twice after un_register even followed by registration with force_full_replica option:

![image](https://user-images.githubusercontent.com/91731013/146776870-028a28ed-4159-4ba9-8bdd-7471026b2c97.png)

PR #644, opened 2021-12-21
Delete obsolete sap system instances

### LINKED ISSUE #576: Obsolete database info in Hosts detail view after un_registration
Trento version: v0.6.0
Trento environment: QA testing cycle 2
Description: following an issue with HANA system replication in the Trento testing environment, I had to unregister and register the HANA secondary database of one of the SAP systems. The cluster is now working fine but when I access the Trento console, go to the Hosts view and click on the secondary node, I see that the all registered instance is still showing with the hdbdaemon process Stopped (along the processes of the newly registered instance which are all Running). See screenshot:

image

fabriziosestito:
Closes #576 #573

### DIFF
--- a/web/datapipeline/sap_systems_projector.go
+++ b/web/datapipeline/sap_systems_projector.go
@@ -27,35 +27,45 @@ func SAPSystemsProjector_SAPSystemsDiscoveryHandler(dataCollectedEvent *DataColl
 		return err
 	}
 
+	// deletes all obsolete instances if no sap system was discovered
+	if len(discoveredSAPSystems) == 0 {
+		return db.
+			Where("agent_id = ?", dataCollectedEvent.AgentID).
+			Delete(&entities.SAPSystemInstance{}).
+			Error
+	}
+
 	for _, s := range discoveredSAPSystems {
-		instance := entities.SAPSystemInstance{
-			AgentID: dataCollectedEvent.AgentID,
-			ID:      s.Id,
-			SID:     s.SID,
-		}
+		var sapSystemType, dbHost, dbName string
+		var tenants []string
 
 		switch s.Type {
 		case 1:
-			var tenants []string
-
 			for _, tenant := range s.Databases {
 				tenants = append(tenants, tenant.Database)
 			}
-
-			instance.Type = models.SAPSystemTypeDatabase
-			instance.Tenants = tenants
+			sapSystemType = models.SAPSystemTypeDatabase
 		case 2:
-			instance.Type = models.SAPSystemTypeApplication
-			instance.DBHost = fmt.Sprint(s.Profile["SAPDBHOST"])
+			sapSystemType = models.SAPSystemTypeApplication
+			dbHost = fmt.Sprint(s.Profile["SAPDBHOST"])
 
 			if hdb, ok := s.Profile["dbs/hdb/dbname"]; ok {
-				if dbName, ok := hdb.(string); ok {
-					instance.DBName = dbName
-				}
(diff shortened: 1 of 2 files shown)


web/datapipeline/sap_systems_projector.go
```diff
@@ -79,31 +81,43 @@ func SAPSystemsProjector_SAPSystemsDiscoveryHandler(dataCollectedEvent *DataColl
...
+			"system_replication", "system_replication_status",
+			"sap_hostname", "start_priority", "http_port", "https_port", "status",
+			"tenants", "db_host", "db_name")
+		if err != nil {
+			return err
+		}
 
+		if len(instances) > 0 {
+			for _, instance := range instances {
+				db = db.Where("NOT (id = ? AND agent_id = ? AND instance_number = ?)", s.Id, dataCollectedEvent.AgentID, instance.InstanceNumber)
```
arbulu89, line 99: It looks like we still have some remaining `Id` instead of `ID`...
It would be totally the same, but maybe `... instance.ID, instance.AgentID, instance.InstanceNumber)` looks better

web/datapipeline/sap_systems_projector.go
```diff
@@ -79,31 +81,43 @@ func SAPSystemsProjector_SAPSystemsDiscoveryHandler(dataCollectedEvent *DataColl
...
+			instances,
+			"id", "sid", "type", "features", "instance_number",
+			"system_replication", "system_replication_status",
+			"sap_hostname", "start_priority", "http_port", "https_port", "status",
+			"tenants", "db_host", "db_name")
+		if err != nil {
+			return err
+		}
 
+		if len(instances) > 0 {
```
arbulu89, line 97: What happens if we receive 0 instances suddenly?
We wouldn't remove anything, which looks wrong.
For instance, if at some point we uninstall the SAP system

arbulu89 (changes requested):
@fabriziosestito Nice!
I think the scenario where we don't receive any instance (SAP installation uninstalled) wouldn't be covered, as the code doesn't remove anything in that case.
In fact, I don't know what the agent would send in this case...

web/datapipeline/sap_systems_projector.go
fabriziosestito: ah yes good call i totally forgot about this scenario i will add the case

arbulu89: @fabriziosestito We could still have no SAP systems at all. This means that it wouldn't go within the 1st loop (`for _, s := range discoveredSAPSystems {`).

fabriziosestito: @arbulu89 it should cover now both scenarios

arbulu89 (approved):
Go!

dottorblaster (approved):
:shipit:

nelsonkopliku (approved):

abravosuse:
Hey guys, please refer to the comment I just added to issue #576 because something very strange happened when I updated Trento server and agents to the latest version in the QA environment. Essentially all SAP instances disappeared.

arbulu89:
> Hey guys, please refer to the comment I just added to issue #576 because something very strange happened when I updated Trento server and agents to the latest version in the QA environment. Essentially all SAP instances disappeared.

Thank you for reporting @abravosuse ,
It looks like @fabriziosestito fixed it in: #645

abravosuse:
Yes, @arbulu89. Sorry, I gave my feedback at issue #576 and I should have done it here as well. All good now.

Issue #410, opened 2021-10-29
Add tests to the cmd line and env variables usage


arbulu89:
After suffering a quite tricky issue: https://github.com/trento-project/trento/pull/409

The `cobra/viper` combination is super interesting to attach command lines flags and configuration files to variables, but it has the risk that it includes a high degree of auto-magic (black-box operations).

In order to deal with this, we would need to start adding tests to the command line and environment variables loading. Basically we need to know that all existing variables are used properly, if they are loaded of course.
This would help us a lot to avoid regressions. 

Besides this, tests on the `LoadConfig` functions would be interesting too.

Issue #380, opened 2021-10-26
Add mTLS agent/server configuration to the installers and the helm chart


fabriziosestito:
Add mTLS agent/server configuration to the installers and the helm chart, allow the user to pass certificates and enabled the authentication if needed

PR #662, opened 2021-12-27
Add mTLS to the helm chart

fabriziosestito:
This PR adds mTLS values to the helm chart and fixes several flag/configuration issues in our cli that emerged during the task.

It also adds configuration tests for agent/runner/web commands.

Closes: #410 and partially  #380

### DIFF
--- a/cmd/runner/runner.go
+++ b/cmd/runner/runner.go
@@ -15,12 +15,12 @@ import (
 	"github.com/trento-project/trento/runner"
 )
 
-var apiHost string
-var apiPort int
-var interval int
-var ansibleFolder string
-
 func NewRunnerCmd() *cobra.Command {
+	var apiHost string
+	var apiPort int
+	var interval int
+	var ansibleFolder string
+
 	runnerCmd := &cobra.Command{
 		Use:   "runner",
 		Short: "Command tree related to the runner component",
@@ -49,7 +49,7 @@ func NewRunnerCmd() *cobra.Command {
 	return runnerCmd
 }
 
-func start(cmd *cobra.Command, args []string) {
+func start(*cobra.Command, []string) {
 	var err error
 
 	signals := make(chan os.Signal, 1)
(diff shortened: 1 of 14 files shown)


cmd/runner/runner.go
```diff
@@ -15,42 +16,42 @@ import (
...
-	})
-
 	runnerCmd.AddCommand(startCmd)
 
 	return runnerCmd
 }
 
-func start(cmd *cobra.Command, args []string) {
+func start(*cobra.Command, []string) {
+	fmt.Println("jere")
```
nelsonkopliku, line 54: Leftover?

packaging/helm/trento-server/charts/trento-web/templates/deployment.yaml
```diff
@@ -43,6 +43,15 @@ spec:
             - "{{ .Values.webService.port }}"
             - --collector-port
             - "{{ .Values.collectorService.port }}"
+            {{ if .Values.mTLS.enabled }}
+            - --enable-mtls
+            {{ end }}
+            - --cert
```
nelsonkopliku, line 49: shouldn't `--cert` `--key` and `--ca` be inside the `if` ? :thinking:

nelsonkopliku (approved):
LGTM. just a small leftover.

packaging/helm/trento-server/charts/trento-web/templates/deployment.yaml
fabriziosestito: doesn't change much because if the flag is not set they are being ignored but i will change it for clarity, thanks!

cmd/runner/runner.go
fabriziosestito: woops

packaging/helm/trento-server/charts/trento-web/templates/deployment.yaml
nelsonkopliku: True, I remember the implementation.
Yet, my OCD thanks you :joy:

fabriziosestito: no you are right it is clearer, i just didn't think of it

dottorblaster (approved):

PR #665, opened 2021-12-28
Add a photofinish POC

dottorblaster:
Since we had the need to replay some events and get fixtures, I tried to get together all the needs we had in a tiny CLI tool.

Photofinish reads a `.photofinish.toml` file in the current working directory and:

- It outputs the fixture sets in the TOML file;
- It issues POST requests against the endpoint we give (default: `http://localhost:8081/api/collect`) with the content of the fixture files as request body.

I think the README summarizes pretty well what photofinish does.

The objective is to keep photofinish in this repository as we could integrate further features and keep it as a _dev cli_ but no hard feelings, we can split the two things. There are some things missing like a subcommand to drop the current database and recreate it, and the code is really a POC full of `unwrap`s, but I wanted to reach some point where we had something usable as quickly as possible.


tools/photofinish/README.md
```diff
@@ -0,0 +1,53 @@
+# photofinish - a little, handy tool to replay events
+
+Since we had the need to replay some events and get fixtures, I tried to get together all the needs we had in a tiny CLI tool.
```
nelsonkopliku, line 3: I'd rephrase it to something along these lines
> This tiny CLI tool aims to fulfill the need to replay some events and get fixtures

nelsonkopliku (approved):
Noice!

tools/photofinish/.photofinish.toml
```diff
@@ -0,0 +1,12 @@
+[first-test-scenario]
+files = [
+  "../../test/fixtures/discovery/host/expected_published_host_discovery.json",
+  "../../test/fixtures/discovery/sap_system/sap_system_discovery_application.json",
```
fabriziosestito, line 4: Ah this is smart, when you told me about the toml file i imagined a super complicated stuff with the nodes definition which actually we don't really need. I like this

fabriziosestito (approved):
I like how tiny it is, we need to remove the unwraps and work on the CI integration

rtorrero:
We have discussed a bit about this on slack but basically I see 3 things:
 - Missing a bit of documentation on how to compile/run this using `cargo` (maybe adding a job in the Makefile?)
 - The `sap_system_discovery_*.json` files seem to be missing the wrapping with the `agent_id` and `discovery_type` which causes the sap system discovery POST to fail. I'll sent a PR to address this as it falls outside of the scope of this PR
 - I noticed that it will report `Successfully POSTed` always when it is able to find an endpoint listening, even when it's the wrong endpoint (e.g. an endpoint which falls outside of the collector engine)
 
 Thanks for this, I think it's going to be quite useful :-)

tools/photofinish/.photofinish.toml
nelsonkopliku, line 4: As discussed with @rtorrero,  `../../test/fixtures/discovery/sap_system/sap_system_discovery_application.json` cannot be POSTed to the collector.
We're probably missing something like `expected_published_sap_system_discovery_application.json` with `agent_id` and `discovery_type`

nelsonkopliku (changes requested):
Changed my mind after @rtorrero review and discussion with him.

dottorblaster:
Ah sorry about that, the `.photofinish.toml` file that we have right now was meant mainly to test photofinish itself, I thought we could add another `.photofinish.toml` at the root level of the repo that we could populate with the actual suites :sweat_smile: anyway you're right, we just have to decide if we're gonna stick with this file or put the definitive one at the root level

tools/photofinish/.photofinish.toml
fabriziosestito: @nelsonkopliku @rtorrero I believe we need a new the fixtures folders with the acceptance testing scenarios, and then see if those fixtures could be used in the unit tests, maybe even improving them.
So, as for now i would keep the test fixtures as-is, and maybe comment out the toml till we don't have real scenarios to wire-up.

Wdyt?

rtorrero: @fabriziosestito yep, agreed, I'm putting something together, I'll ping you if I get lost on the way :-)

stefanotorresi:
This is indeed super-cool!! 😎 
I'm wondering if we should move this into a dedicated repo, though, as it doesn't really share a single line of code with the root project.

dottorblaster:
@stefanotorresi it's something I would evaluate, actually we talked about it with @fabriziosestito and I was pretty much convinced that it needed a separate repo, but we decided then to leverage our monorepo-like structure to host the tools we write for ourselves :smile:

stefanotorresi:
Let's move this to https://github.com/trento-project/photofinish right away.

PR #692, opened 2022-01-11
Add ctl commands

fabriziosestito:
This PR add a `debug` commands tree and refactor the existing prune events and prune checks utility functions inside this new command.

It also adds:
- reset db command
- export scenario command

The first is useful for developing and in out CI, so it can be used by cypress to reset the database between tests.
The second is an utility that we can use to bake fixtures from real clusters.

A follow up PR will add a script to export scenario from a running k3s installation.

image


cmd/debug/debug.go
```diff
@@ -0,0 +1,215 @@
...
+	"github.com/trento-project/trento/web"
+	"github.com/trento-project/trento/web/datapipeline"
+	"github.com/trento-project/trento/web/entities"
+	"gorm.io/gorm"
+)
+
+func NewDebugCmd() *cobra.Command {
+	debugCmd := &cobra.Command{
+		Use:   "debug",
+		Short: "Maintanance commands, USE WITH CAUTION.",
```
arbulu89, line 26: Typo `Maintenance`

cmd/debug/debug.go
```diff
@@ -0,0 +1,215 @@
...
+		PersistentPreRunE: func(debugCmd *cobra.Command, _ []string) error {
+			debugCmd.Flags().VisitAll(func(f *pflag.Flag) {
+				viper.BindPFlag(f.Name, f)
+			})
+			debugCmd.PersistentFlags().VisitAll(func(f *pflag.Flag) {
+				viper.BindPFlag(f.Name, f)
+			})
+			return internal.InitConfig("web")
```
arbulu89, line 34: Is this `web` a leftover, or something we need to have there?

cmd/debug/debug.go
```diff
@@ -0,0 +1,215 @@
...
+	})
+
+	if err != nil {
+		log.Fatal("Error while resetting the database: ", err)
+	}
+
+	log.Info("Database reset.")
+}
+
+func dumpScenario(db *gorm.DB, exportPath string, scenarioName string) {
```
arbulu89, line 181: Nice. Would have any value to dump n number of entries? Instead of the latest one?
I would find this really useful for bug debugging.
Anyway, with the super big amount of records that we would have, it would still be difficult

cmd/debug/debug.go
```diff
@@ -0,0 +1,215 @@
...
+		data, err := json.MarshalIndent(map[string]interface{}{
+			"agent_id":       event.AgentID,
+			"discovery_type": event.DiscoveryType,
+			"payload":        event.Payload,
+		}, "", " ")
+		if err != nil {
+			log.Fatal("Error while marshaling event: ", err)
+		}
+
+		filePath := filepath.Join(path, event.DiscoveryType, event.AgentID)
```
arbulu89, line 209: Here is hard to tell, but maybe grouping 1st with Agents rather than types would make the organization easier to filter.
In any case, both are totally OK

cmd/debug/debug.go
```diff
@@ -0,0 +1,215 @@
...
+func dbReset(db *gorm.DB) {
+	log.Info("Resetting database...")
+	err := db.Transaction(func(tx *gorm.DB) error {
+		for _, t := range web.DBTables {
+			stmt := &gorm.Statement{DB: db}
+			stmt.Parse(t)
+			tableName := stmt.Schema.Table
+
+			err := tx.Raw("TRUNCATE TABLE ?", tableName).Error
+			log.Infof("Table %s truncated.", tableName)
```
arbulu89, line 166: This log shouldn't go after the error checking?

arbulu89 (approved):
Really nice!
I'm not sure if `debug` is the best name here, but I reckon that it is a really difficult decision.
Besides that, the code looks fantastic

cmd/debug/debug.go
fabriziosestito: good call, usual leftover thanks!

cmd/debug/debug.go
fabriziosestito: This actually export the latest events needed to build the latest state of the application, one per type per agent, by looking to the latest projected event from the subscriptions.

fabriziosestito:
> Really nice! I'm not sure if `debug` is the best name here, but I reckon that it is a really difficult decision. Besides that, the code looks fantastic

I'm also not sure about `debug` and I am more than happy to change it, I would have used something like `tasks` but as @stefanotorresi this could be confusing. What do you guys think could be a proper name?

cmd/debug/debug.go
fabriziosestito: nice catch!

nelsonkopliku (approved):
Great stuff! :rocket: 

I also agree that `debug` might not be the best naming.

I like the `tasks` idea, how would it look like?
```
$ trento tasks db-reset
```

What about something like
```
$ trento run db-rest
```
or
```
$ trento run-task db-reset
```

stefanotorresi:
What about `trento system foo|bar`?

stefanotorresi:
`trento ctl' ? :trollface:

PR #693, opened 2022-01-12
Fix error in prune checks chart declaration

arbulu89:
I made a mistake in https://github.com/trento-project/trento/pull/661

```
Error: UPGRADE FAILED: failed to create resource: CronJob in version "v1" cannot be handled as a CronJob: v1.CronJob.Spec: v1.CronJobSpec.JobTemplate: v1.JobTemplateSpec.Spec: v1.JobSpec.Template: v1.PodTemplateSpec.Spec: v1.PodSpec.Containers: []v1.Container: v1.Container.Args: []string: ReadString: expects " or n, but found 3, error found in #10 byte of ...|er-than",30,"--db-ho|..., bigger context ...|":[{"args":["web","prune-checks","--olther-than",30,"--db-host","trento-server-postgresql"],"image":|...
```

This change fixes it

### DIFF
--- a/packaging/helm/trento-server/Chart.yaml
+++ b/packaging/helm/trento-server/Chart.yaml
@@ -1,12 +1,12 @@
-#!BuildTag: trento/trento-server:0.2.8
-#!BuildTag: trento/trento-server:0.2.8-build%RELEASE%
+#!BuildTag: trento/trento-server:0.2.9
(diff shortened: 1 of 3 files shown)


rtorrero (approved):
Thanks @arbulu89 was just looking into this :-)

PR #697, opened 2022-01-13
Fix prune jobs

fabriziosestito:
Fix leftover debug command in helm chart prune jobs


arbulu89 (approved):
I guess we should upgrade the `trento-web` chart version as well

dottorblaster (approved):
`:triggered-fabrizio:`

PR #703, opened 2022-01-14
Switch to the SLE BCI images

dcermak:


Dockerfile
```diff
@@ -1,36 +1,33 @@
...
-FROM golang:1.16 as go-build
+FROM registry.suse.com/bci/golang:1.16 as go-build
 WORKDIR /build
 # we add what's needed to download go modules so that dependencies can be cached in a dedicate layer
 ADD go.mod go.sum /build/
 RUN go mod download
 ADD . /build
 COPY --from=node-build /build /build
-RUN make build
+RUN zypper -n in make git && make build
```
arbulu89, line 21: Do we need to install `make` here back again?

Dockerfile
```diff
@@ -1,36 +1,33 @@
-FROM node:16 AS node-build
+FROM registry.suse.com/bci/nodejs:14 AS node-build
```
arbulu89, line 1: @dottorblaster Do you know if this can bring as some potential error?
Moving from node16 backwards to node14 (which I guess is the version packaged by SUSE now)

arbulu89 (approved):
Hey @dcermak ,
So far so good.
I have created and tested the containers, and I don't see any difference compared with the old builds.
As a comment, we already have one base container which doesn't come from SUSE:
`FROM gcr.io/distroless/base:debug AS trento-web`
https://github.com/trento-project/trento/blob/main/Dockerfile#L44

Should we do something about it?

Dockerfile
dottorblaster: @arbulu89 I don't know if this brings potential errors but I wouldn't go backwards this way, node 16 is the latest LTS and I would try to maintain at least v16. @dcermak do you know if SUSE has node16 images?

dcermak: Not yet, we have a node 16 image in the works, but it hasn't been created & published yet.

Dockerfile
```diff
@@ -1,36 +1,33 @@
...
-RUN ln -s /usr/local/bin/python /usr/bin/python \
-    && /usr/bin/python -m venv /venv \
-    && /venv/bin/pip install 'ansible~=4.6.0' 'requests~=2.26.0' 'rpm==0.0.2' 'pyparsing~=2.0' \
-    && apt-get update && apt-get install -y --no-install-recommends \
-      ssh \
-    && apt-get purge -y --auto-remove -o APT::AutoRemove::RecommendsImportant=false \
-    && rm -rf /var/lib/apt/lists/*
+FROM registry.suse.com/bci/python:3.9 AS trento-runner
+RUN /usr/bin/python3 -m venv /venv \
+    && /venv/bin/pip install 'ansible~=4.6.0' 'ara~=1.5.7' 'rpm==0.0.2' 'pyparsing~=2.0' \
```
dirkmueller, line 25: it was requests, now its ara, is this an intentional change?

fcrozat:
> Thanks for the review @arbulu89! Unfortunately we have nothing really comparable to the distroless container. The smallest image that we have is the micro image (registry.suse.com/bci/micro:15.3), which comes only with the barest minimum required to run bash (no rpm, no zypper).

bci/micro container is the equivalent to google distroless:base container (glibc / openssl / passwd).

Dockerfile
arbulu89: Really true! This is an error. It should be `requests`!

arbulu89 (changes requested):

Dockerfile
dirkmueller, line 21: most likely installing `git-core` should be good enough, not the much larger `git`

stefanotorresi (approved):
LGTM

Dockerfile
```diff
@@ -41,7 +38,7 @@ COPY --from=go-build /build/trento /app/trento
 LABEL org.opencontainers.image.source="https://github.com/trento-project/trento"
 ENTRYPOINT ["/tini", "--", "/app/trento"]
 
-FROM gcr.io/distroless/base:debug AS trento-web
+FROM registry.suse.com/bci/micro:15.3 AS trento-web
```
rtorrero, line 41: Did you mean here `registry.suse.com/bci/bci-micro:15.3`?

dcermak: It was still `bci/micro` when I created this PR :see_no_evil:

Dockerfile
dcermak: FYI, the nodejs 16 image will probably be published as a beta release with the 15 SP4 beta

dottorblaster: @dcermak can we use it just referencing it in the dockerfile when the beta comes out? Or do we need additional work? This would be great!

dcermak: I think we would just have to reference it and that would be it.

dottorblaster: Perfect, thank you!

ajaeger:
Should this be updated to the new nodejs16 image that was just released?

dcermak:
Andreas Jaeger ***@***.***> writes:

> Should this be updated to the new nodejs16 image that was just released?

Done!

arbulu89:
> Andreas Jaeger ***@***.***> writes:
> Should this be updated to the new nodejs16 image that was just released?
> Done!

Hey @dcermak ,

ajaeger:
> Andreas Jaeger ***@***.***> writes:
> Should this be updated to the new nodejs16 image that was just released?
> Done!

You're fast ;) Thanks!

arbulu89 (approved):
Tested the new container images, and they worked fine

Issue #655, opened 2021-12-22
Premium badge in the checks catalog out of place


arbulu89:
The badge which says that the check is premium has gone out of place. It is totally miss aligned.
![image](https://user-images.githubusercontent.com/36370954/147096483-70afb1c5-5667-456c-9776-2b7df70add7e.png)

This needs some CSS work or move the badge to some other column

PR #707, opened 2022-01-14
Replace premium check position to description column

arbulu89:
Fix for: https://github.com/trento-project/trento/issues/655

In order to show the premium badge properly, this PR moves the badge to the description column:

image


stefanotorresi:
Hmm I really don't like this placement esthetically, but I understand we need to fix the current one that is broken.
We need to follow up with a UI enhancement about this.

fabriziosestito (approved):

PR #712, opened 2022-01-17
Skip non dc cluster projection

fabriziosestito:
This PR adds the `DC` (designated controller) to the cluster discovery struct and payload from the agent, so we can skip whatever discovery eventwas  not sent from the elected designated controller in the cluster projector.
This is needed to have a single source of truth.

We still accept non-`DC` events in the other projectors (e.g. the host projector uses this to populate the cluster field in the host model).


web/datapipeline/clusters_projector.go
```diff
@@ -36,6 +36,10 @@ func clustersProjector_ClusterDiscoveryHandler(event *DataCollectedEvent, db *go
 		return err
 	}
 
+	if !cluster.DC {
+		return nil
```
arbulu89, line 40: Could we add a debugging trace here saying `Cluster node is not a DC. Skipping projection` or similar

arbulu89 (approved):
Nice!

nelsonkopliku (approved):

web/datapipeline/clusters_projector.go
fabriziosestito: done

PR #713, opened 2022-01-17
Fix dump scenario command and add tests

fabriziosestito:
This PR fixes the dump scenario query, which was wrong because it was considering only the latest event from the projector subscriptions, while a projector can handle multiple events, resulting in an incomplete scenario dump.

It also adds tests.

### DIFF
--- a/cmd/ctl/ctl_test.go
+++ b/cmd/ctl/ctl_test.go
@@ -103,3 +103,65 @@ func (suite *CtlTestSuite) TestPruneChecksResults() {
 	suite.Equal(1, len(prunedChecksResults))
 	suite.Equal(int64(3), prunedChecksResults[0].ID)
 }
+
+func (suite *CtlTestSuite) TestGetLatestEvents() {
+	suite.tx.AutoMigrate(&datapipeline.DataCollectedEvent{})
+
+	events := []datapipeline.DataCollectedEvent{
+		{
+			ID:            1,
+			AgentID:       "agent_id_1",
+			DiscoveryType: "discovery_type_1",
+			Payload:       []byte("{}"),
+		},
+		{
+			ID:            2,
+			AgentID:       "agent_id_1",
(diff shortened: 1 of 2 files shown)


cmd/ctl/ctl_test.go
```diff
@@ -103,3 +103,51 @@ func (suite *CtlTestSuite) TestPruneChecksResults() {
...
+			AgentID:       "agent_id_2",
+			DiscoveryType: "discovery_type_2",
+			Payload:       []byte("{}"),
+		},
+		{
+			ID:            5,
+			AgentID:       "agent_id_2",
+			DiscoveryType: "discovery_type_2",
+			Payload:       []byte("{}"),
+		},
```
arbulu89, line 141: Can we add an additional discovery here in the test with a different discovery type?
To test if the latest event with other discovery type is retrieved as well.
Simply add a new entry with `discovery_type_3` for any of this 2 agents, and check if it is obtained

arbulu89 (approved):
Thanks!
To understand it better.
The 1st subquery gets all the ids for each `agent_id/discovery_type` tuple, right?
PD: I suggest to improve a bit the added test to include a normal use case, which is to have other discovery types per agent

dottorblaster (approved):
LGTM

nelsonkopliku (approved):

fabriziosestito:
> Thanks! To understand it better. The 1st subquery gets all the ids for each `agent_id/discovery_type` tuple, right? PD: I suggest to improve a bit the added test to include a normal use case, which is to have other discovery types per agent

The subquery select the max id from after grouping by agent_id and discovery_type. The max is important since it is the latest one.
The other query just select everything that has the id returned by the subquery.

cmd/ctl/ctl_test.go
fabriziosestito: @arbulu89 test added

Issue #709, opened 2022-01-14
Individual checks are not properly highlighted when selected in the cluster settings modal


arbulu89:
I guess that after the EOS update, we got this regression.
The individual checkboxes in the cluster settings modal are not correctly highlighted. It only changes the border color but it doesn't add any marker.

![image](https://user-images.githubusercontent.com/36370954/149517811-45f0f2bf-8b85-4986-a19c-e9dc0e0e07ba.png)

arbulu89:
Apparently this is something on EOS framework. They have the same issue in the example page:
https://productbrand.suse.com/tables

PR #714, opened 2022-01-18
Fix checkbox not shown as selected inside tables

dottorblaster:
One picture is worth a thousand words:

image

The faulty one can be found in #709


arbulu89 (approved):
Love you dude!

nelsonkopliku (approved):

PR #725, opened 2022-01-20
Fix health container numbers and pagination numbers

arbulu89:
- Fix how the health container numbers are shown, to show all the numbers except the current page numbers.
- Filter properly the health with pagination. Now, the health filter is done together with the others using SQL, so it is included in the pagination. Before this, as the health filter was applied after the pagination, we could get the elements in pages that don't really exist (like only having 2 hosts, but being them in page 4).
- Fix the pagination total numbers and page count

In order to improve the code structure, I have moved some structs to models
test

### DIFF
--- a/test/e2e/cypress/integration/hosts_overview.js
+++ b/test/e2e/cypress/integration/hosts_overview.js
@@ -5,7 +5,7 @@ context('Hosts Overview', () => {
     before(() => {
         cy.resetDatabase()
         cy.loadScenario('healthy-27-node-SAP-cluster')
-        
+
         cy.task('startAgentHeartbeat', agents())
         cy.visit('/');
         cy.navigateToItem('Hosts')
@@ -15,6 +15,12 @@ context('Hosts Overview', () => {
     describe('Registered Hosts should be available in the overview', () => {
         it('should show 10 of the 27 registered hosts with default pagination settings', () => {
             cy.get('.tn-hostname').its('length').should('eq', 10)
+        })
+        it('should show 27 as total items in the pagination controls', () => {
+            cy.get('.pagination-count').should('contain', '27 items')
+        })
+        it('should have 5 pages', () => {
+            cy.get('.page-item').its('length').should('eq', 5)
         })
         it('should show all of the all 27 registered hosts when increasing pagination limit to 100', () => {
             cy.reloadList('hosts', 100)
@@ -34,17 +40,16 @@ context('Hosts Overview', () => {
 
     describe('Health Detection', () => {
         describe('Health Container shows the health overview of the entire cluster', () => {
-            it('should show health status of the first 10 visible hosts', () => {
-                cy.log("This test needs to be removed in favor of having the Health Container showing information of the entire cluster")
+            it('should show health status of the entire cluster of 27 hosts with partial pagination', () => {
                 cy.reloadList('hosts', 10)
-                cy.get('.health-container .health-passing').should('contain', 10)
+                cy.get('.health-container .health-passing').should('contain', 27)
             })
             it('should show health status of the entire cluster of 27 hosts', () => {
                 cy.reloadList('hosts', 100)
                 cy.get('.health-container .health-passing').should('contain', 27)
             })
         })
-    
+
         describe('Detected hosts Health matches deployed server status', () => {
             it('all 27 hosts in the cluster should be up', () => {
                 availableHosts.forEach((hostName) => cy.get(`#host-${hostName} > .row-status > i`).should('have.class', 'text-success'))
@@ -93,7 +98,7 @@ context('Hosts Overview', () => {
             cy.wait('@resetFilter')
         }
 
-        describe('Filtering by health', () => { 
+        describe('Filtering by health', () => {
             before(() => {
                 cy.get('.tn-filters > :nth-child(2) > .btn').click()
             })
@@ -104,12 +109,14 @@ context('Hosts Overview', () => {
             ]
             healthScenarios.forEach(([health, expectedHostsWithThisHealth], index) => {
                 it(`should show ${expectedHostsWithThisHealth || 'an empty list of'} hosts when filtering by health '${health}'`, () => {
-                    cy.intercept('GET', `/hosts?per_page=100&health=${health}`).as('filterByHealthStatus') 
+                    cy.intercept('GET', `/hosts?per_page=100&health=${health}`).as('filterByHealthStatus')
                     const selectedOption = `#bs-select-1-${index}`
                     cy.get(selectedOption).click()
                     cy.wait('@filterByHealthStatus').then(() => {
                         expectedHostsWithThisHealth == 0 && cy.get('.table.eos-table').contains('There are currently no records to be shown')
                         expectedHostsWithThisHealth > 0 && cy.get('.tn-hostname').its('length').should('eq', expectedHostsWithThisHealth)
+                        cy.get('.pagination-count').should('contain', `${expectedHostsWithThisHealth} items`)
+                        cy.get('.page-item').its('length').should('eq', Math.ceil(expectedHostsWithThisHealth/100)+2)
                         resetFilter(selectedOption)
                     })
                 })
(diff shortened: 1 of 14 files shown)


nelsonkopliku (commented):
Looks good @arbulu89! Thanks for addressing this!

There's a couple of little comments, tho! eheh :smile: Nothing major.

Keep this in mind: If you rebase on `main`, the `hosts_overview.js` test should fail, as the behavior changed.
Specifically this one `should show health status of the first 10 visible hosts`, so the test needs to be updated, and the `feature` file too :smile: 

As an extra reminder for future discussions: I am wondering whether we are doing too much things in our controllers and if the application might need some refactor as we go further refining requirements.

arbulu89:
> Looks good @arbulu89! Thanks for addressing this!
> 
> There's a couple of little comments, tho! eheh smile Nothing major.
> 
> Keep this in mind: If you rebase on `main`, the `hosts_overview.js` test should fail, as the behavior changed. Specifically this one `should show health status of the first 10 visible hosts`, so the test needs to be updated, and the `feature` file too smile
> 
> As an extra reminder for future discussions: I am wondering whether we are doing too much things in our controllers and if the application might need some refactor as we go further refining requirements.

Yes. I'm aware of this divergence in the recently added tests, and I'm working on them. Thank you for the reminder though.

web/services/clusters.go
```diff
@@ -49,6 +49,28 @@ func NewClustersService(db *gorm.DB, checksService ChecksService) *clustersServi
 
 func (s *clustersService) GetAll(filter *ClustersFilter, page *Page) (models.ClusterList, error) {
 	var clusters []entities.Cluster
+	var healthFilteredClusters []string
+
+	// Filter the clusters by Health
+	if filter != nil && len(filter.Health) > 0 {
+		var checksResults []entities.ChecksResult
```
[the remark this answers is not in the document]
arbulu89: Yes, i had the same internal discussion. It could go in `checks_result` service as well, but I didn't want to mix the filtering even there. But I will have a look in any case!
Maybe I can wrap the SQL query in that file at least

arbulu89:
> > Yes. I'm aware of this divergence in the recently added tests, and I'm working on them. Thank you for the reminder though.
> 
> Great!
> 
> If you are experiencing randomic failures during `Health Detection` maybe this #726 could help wink

By the way, do you prefer guys if I rebase already existing e2e and add the updates here, or a new PR with that? There won't be that many changes in cypress, so maybe we can put them all together here.

nelsonkopliku:
> By the way, do you prefer guys if I rebase already existing e2e and add the updates here, or a new PR with that? There won't be that many changes in cypress, so maybe we can put them all together here.

I think the test needs to be updated in this PR, otherwise e2e will fail and pipeline won't proceed

web/services/clusters.go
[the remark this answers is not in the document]
arbulu89: Done in https://github.com/trento-project/trento/pull/725/commits/1c80b6e7e9221b72f642dcc9e8ab23a88fe17260

arbulu89:
@nelsonkopliku Changes applied, main branch rebased and e2e test adapted!

web/clusters.go
```diff
@@ -81,18 +87,13 @@ func NewClusterListHandler(clustersService services.ClustersService) gin.Handler
...
+		healthContainer := NewClustersHealthContainer(clusterListAll)
 		healthContainer.Layout = "horizontal"
 
-		count, err := clustersService.GetCount()
-		if err != nil {
-			_ = c.Error(err)
-			return
-		}
-		pagination := NewPagination(count, pageNumber, pageSize)
+		pagination := NewPagination(len(clusterListAll), pageNumber, pageSize)
```
fabriziosestito: I would instead introduce another health count service method that gives as an output the health counts directly, and the delegate that to the service instead of loading again the whole list in the controller, and use the count method for the pagination lenght.

web/hosts.go
```diff
@@ -49,36 +49,37 @@ func NewHostListHandler(hostsService services.HostsService) gin.HandlerFunc {
...
 		if err != nil {
 			_ = c.Error(err)
 			return
 		}
-		pagination := NewPagination(count, pageNumber, pageSize)
 
-		hContainer := NewHostsHealthContainer(hostList)
+		pagination := NewPagination(len(hostListAll), pageNumber, pageSize)
+
+		hContainer := NewHostsHealthContainer(hostListAll)
```
fabriziosestito: Same as for the cluster comment, we fetch the list twice, while we just need the health count. I understand that queries are to be done to aggregate the data in the service anyway, but fetching the whole list with preloaded data seems overkill. It also helps with clarity in the controller which I feel is doing too much stuff ATM.

fabriziosestito (changes requested):

web/sap_systems.go
```diff
@@ -91,34 +92,35 @@ func NewHANADatabaseListHandler(sapSystemsService services.SAPSystemsService) gi
...
 
-		filterTags, err := sapSystemsService.GetAllDatabasesTags()
+		filterSIDs, err := sapSystemsService.GetAllDatabasesSIDs()
 		if err != nil {
 			_ = c.Error(err)
 			return
 		}
 
-		count, err := sapSystemsService.GetDatabasesCount()
+		filterTags, err := sapSystemsService.GetAllDatabasesTags()
```
fabriziosestito, line 113: Same as above, I don't think we need to do the whole table fetching.

fabriziosestito:
@arbulu89 thanks for addressing this! I have some doubts about how we fetch the paginated / non-paginated list twice in the contoller, since it seems that we can delegate some stuff to the service and maybe improve the fact that we load the whole preloaded list twice.

web/clusters.go
arbulu89: But what would this improve? In order to get the health numbers, we need to get all the clusters with the applied filters and get the health data. 
We could create other service method, but it would do exactly the same as far as I see. I don't know any way to save any query to the database. 
It might look cleaner, yes. But the code efficiency and in fact almost all the code would be the same.
Or am I missing something?

PD: And we cannot use the currently existing `GetCount` method, as we need to get the count with the filtered values, which at the end, would require to use the same sql where conditions we have now. Does it make sense to complicate the code yet more?

arbulu89: PD2: The unique way I see to reduce this to a unique call is to do the pagination within the code instead of using sql conditions

fabriziosestito: > It might look cleaner, yes. But the code efficiency and in fact almost all the code would be the same.
Or am I missing something?

If I am not mistaken, I think that to get the health status we don't need a "select *" query that also does various preload (e.g. Tags) but just the IDs of  the clusters (a select id + pluck) should be enough to be used in the check service `GetAggregatedChecksResultByCluster(cluster.ID)`

> PD: And we cannot use the currently existing GetCount method, as we need to get the count with the filtered values, which at the end, would require to use the same sql where conditions we have now.

By abstracting the filtering to a new method we can shape a GetCount with the filters, I don't think this would complicate the code a lot.

arbulu89: No, the get the health we need to preload the tags, as we need to filter them. And we would need to get the `id`, `name`, `cluster_type` and `sid`. And we would need to get the checks results too. So the unique difference would to we discard only the number of hosts and the number of resources

nelsonkopliku:
> @nelsonkopliku Changes applied, main branch rebased and e2e test adapted!
>

Ah great stuff! Hope you did't encounter too many quirks :sweat_smile: 

I also took a look at the `test/e2e/cypress/fixtures/hosts-overview/hosts_overview.feature` and actually it does seem to still properly enough describe the feature, so no actual change is required there.

test/e2e/cypress/integration/hosts_overview.js
```diff
@@ -104,12 +109,14 @@ context('Hosts Overview', () => {
...
             healthScenarios.forEach(([health, expectedHostsWithThisHealth], index) => {
                 it(`should show ${expectedHostsWithThisHealth || 'an empty list of'} hosts when filtering by health '${health}'`, () => {
-                    cy.intercept('GET', `/hosts?per_page=100&health=${health}`).as('filterByHealthStatus') 
+                    cy.intercept('GET', `/hosts?per_page=100&health=${health}`).as('filterByHealthStatus')
                     const selectedOption = `#bs-select-1-${index}`
                     cy.get(selectedOption).click()
                     cy.wait('@filterByHealthStatus').then(() => {
                         expectedHostsWithThisHealth == 0 && cy.get('.table.eos-table').contains('There are currently no records to be shown')
                         expectedHostsWithThisHealth > 0 && cy.get('.tn-hostname').its('length').should('eq', expectedHostsWithThisHealth)
+                        cy.get('.pagination-count').should('contain', `${expectedHostsWithThisHealth} items`)
```
nelsonkopliku, line 118: As a suggestion for the future:

This could be a custom `cy` command https://docs.cypress.io/api/cypress-api/custom-commands, and we can think of it as a custom assertion for pagination.

Something along these lines:
```
Cypress.Commands.add('paginationShouldMatch', (itemsCount) => {
    cy.get('.pagination-count').should('contain', `${itemsCount} items`)
    cy.get('.page-item').its('length').should('eq', Math.ceil(itemsCount/100)+2)
})
```

which then becomes accessible as
```
cy.paginationShouldMatch(expectedHostsWithThisHealth)

cy.paginationShouldMatch(expectedRelatedHosts)

cy.paginationShouldMatch(expectedTaggedHosts)
```
in the different variants.

Look, don't bother changing it now, as I understand it wouldn't add much value right now :smile: 
Just wanted to share the cypress concept of custom command that might become more and more useful as we go.

nelsonkopliku (approved):
For me it's a green light as long as the discussion with @fabriziosestito is sorted out.

In that regards, here's my two cents:
I think we all agree that we need to discuss about the codebase and code design itself, related also to the domain we're modeling.

So whatever we do here now, it might be anyway subject of change again, so let's do just the good enough.

arbulu89:
Hey @fabriziosestito ,
I would need you to remove your blocker to merge this PR

fabriziosestito (approved):

fabriziosestito:
> Hey @fabriziosestito , I would need you to remove your blocker to merge this PR

@arbulu89 sorry thought i did it already, now it is green

PR #726, opened 2022-01-20
Hearbeat immediately once when starting agent simulation in e2e tests

nelsonkopliku:
An improvement that might save some headaches :smile: 

Basically we heartebeat immediately once before starting the interval.


fabriziosestito (approved):

dottorblaster (approved):
Oh well, fair enough

Issue #382, opened 2021-10-26
Cross reference the related variables between the helm charts


arbulu89:
Based on the comment: https://github.com/trento-project/trento/pull/375#issuecomment-951998633

Now that the `runner` depends on the `web` to get and post information, the `runner` needs to know where the web is hosting the API service, specifically the port.

In order to this, helm offers the possibility to use global variables, as the sibling charts cannot share information. 

Some references: https://docs.helm.sh/docs/topics/charts/#scope-dependencies-and-values

PR #732, opened 2022-01-24
Helm: move `dependencies` variables to `global`

flaviodsr:
Declaring variables as global enables cross-referencing those variables
from lower level charts through the parent helm chart.

This PR also includes the following changes:
- Fix trento-web secret name
- Enable linting on subcharts

Fixes: https://github.com/trento-project/trento/issues/382

### DIFF
--- a/ct.yaml
+++ b/ct.yaml
@@ -1,5 +1,6 @@
 # https://github.com/helm/chart-testing
 chart-dirs:
+  - packaging/helm/trento-server/charts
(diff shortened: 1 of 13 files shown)


arbulu89 (approved):
Hi @flaviodsr ,
The changes look good to me!
I would wait just in case for other reviewers, but green light from my side!

fabriziosestito (approved):
LGTM to me as well

Issue #478, opened 2021-11-19
page 'Pacemaker Clusters' not reloaded automatically after tag removed


Zaoliang:
Pacemaker Clusters 
open Pacemaker Cluster for Pacemaker Cluster details, Sites details (button Details show more details), SBD/Stonith
Issue: page not reloaded automatically after tag removed

How to reproduce:
add a tag 'hana' for hana_cluster for example, go to Tags field and select tag hana, the page shows then only one correct item.
now delete the tag 'hana' from hana_cluster, go to Tags field and the tag is not there anymore, Deselect All doesn't work now.
Or go to Cluster type field and select HANA scaled-up, nothing happened. The page 'Pacemaker Clusters' could not be reloaded automatically.

abravosuse:
Issue persist on v0.6.0 (cycle 2).

arbulu89:
Here how the error looks like:
![Peek 2022-01-21 09-04](https://user-images.githubusercontent.com/36370954/150489713-47d260ad-6a0e-4a48-b2f8-06dc4c9e97d9.gif)

abravosuse:
@arbulu89 may I ask what you used to record that short clip? I see it can be very handy to show case behaviors...

arbulu89:
> @arbulu89 may I ask what you used to record that short clip? I see it can be very handy to show case behaviors...

I use `peek`. In openSUSE you can install using `zypper in peek`.
It only works fine to record short gifs, Long ones fail to render many times

arbulu89:
Hey @dottorblaster !
Do you think it would be worthy to fix this issue moving the tags system to react? Or it would be better a on place fix in the current js code?
I would myself take care of react transition if you think it is the best idea

dottorblaster:
@arbulu89 actually we can't move the tags system to React, we should move the whole view to React in order to make it work properly :persevere:

arbulu89:
> @arbulu89 actually we can't move the tags system to React, we should move the whole view to React in order to make it work properly persevere

Ok, understood. So I will apply a "hacky" solution meanwhile

PR #733, opened 2022-01-24
Fix tag removal when filtering

arbulu89:
Fix for: https://github.com/trento-project/trento/issues/478
Peek 2022-01-21 16-25


arbulu89:
Finally, the e2e test added in this PR works 100% of the times. I just needed to add many intercepts as the tags select element gets rendered every time a new tag is added, re-created the DOM

nelsonkopliku (approved):
LGTM

PR #737, opened 2022-01-25
Update the cluster discovery fixtures for cypress to match the new model

rtorrero:
Update the fixtures for the ha_clusters to match what we expect after https://github.com/trento-project/trento/pull/712


dottorblaster (approved):
CI is green so I guess we can merge this :innocent:

PR #738, opened 2022-01-25
Add SAP Systems overview acceptance tests

fabriziosestito:
This PR adds the SAP Systems overview acceptance tests.

It also solves couple of quirks with the sap system table html and removes the link to "unknown" cluster in the instances details.

Bonus: adds prettier dep and configuration for the e2e folder, reformat the e2e tests and bumps cypress version


web/templates/blocks/sap_systems_table.html.tmpl
```diff
@@ -77,15 +77,22 @@
                                     {{- else }}
                                         <td></td>
                                     {{- end }}
-                                    <td><a href="/clusters/{{ .ClusterID }}">{{ .ClusterName }}</a></td>
+                                    <td>
+                                    {{- if ne .ClusterType "Unknown" }}
```
arbulu89, line 81: Good catch! It wouldn't be safer to check if the cluster is of HANA type?

arbulu89 (approved):
Super nice!
Talking honestly, I didn't have a deep look in the js code of the checks, but it looked totally fine.
By the way, I see that you even check if the link work, rather than only checking the url itself. Should be take this as a good practice?

fabriziosestito:
> Super nice! Talking honestly, I didn't have a deep look in the js code of the checks, but it looked totally fine. By the way, I see that you even check if the link work, rather than only checking the url itself. Should be take this as a good practice?

@arbulu89 I found this as a common practice in the internet and in the cypress documentation as well.
Not sure it is the best practice but it happened to help me spot the broken cluster link that I fixed in this PR, and some other quirks.
My reasoning was simply that a human would have done the same so I kept it.
What is your opinion?

web/templates/blocks/sap_systems_table.html.tmpl
fabriziosestito: This is actually future proofer, since we do not have to change this condition when we will introduce netweaver cluster detail pages

arbulu89:
> > Super nice! Talking honestly, I didn't have a deep look in the js code of the checks, but it looked totally fine. By the way, I see that you even check if the link work, rather than only checking the url itself. Should be take this as a good practice?
> 
> @arbulu89 I found this as a common practice in the internet and in the cypress documentation as well. Not sure it is the best practice but it happened to help me spot the broken cluster link that I fixed in this PR, and some other quirks. My reasoning was simply that a human would have done the same so I kept it. What is your opinion?

Fine for me. I will add the same concepts to the test I'm working on

nelsonkopliku (approved):
Amazing!

Issue #349, opened 2021-10-19
Add licenses checking to the test phase of the build pipeline


stefanotorresi:
Check this out: https://github.com/google/go-licenses

flaviodsr:
I have tried to use [google/go-licenses](https://github.com/google/go-licenses) with no success, even the examples from the README fail to run, there are also multiple issues (including the broken examples) open. Seems like there is an effort to fix many of the issues on [version 2](https://github.com/google/go-licenses/issues/70) which is under development.

[frapposelli/wwhrd](https://github.com/frapposelli/wwhrd) works, however it requires a configuration file where you need to specify the allowed licenses, while `go-licenses` uses [google/licenseclassifier](https://github.com/google/licenseclassifier/blob/main/license_type.go) to automatically classify the license. Besides that, wwhrd only checks packages stored under `vendor/`, so when using Go modules (go mod), you need to run `go mod vendor` before running wwhrd.

While the check is being implemented using `wwhrd` it might be worth keeping an eye for the next version of [google/go-licenses](https://github.com/google/go-licenses) for a solution where we do not need to maintain a list of allowed licenses.

Bobgy:
Hey @flaviodsr! JFYI, go-licenses v2 efforts have been released as v1.1.0 with doc updated. Welcome having a try : )

PR #741, opened 2022-01-26
CI: add job for checking licenses from dependencies

flaviodsr:
The check is implemented using wwhrd [1] and the allowed
licenses are listed on the `.wwhrd.yml` file.

1. https://github.com/frapposelli/wwhrd

Fix: https://github.com/trento-project/trento/issues/349


dottorblaster (approved):
LGTM

.wwhrd.yml
```diff
@@ -0,0 +1,7 @@
+allowlist:
```
arbulu89, line 1: Can we rename the file to have `yaml` extension rather than `yml`, as we already have files with the 2nd extesion

.github/workflows/ci-cd.yaml
```diff
@@ -88,6 +88,18 @@ jobs:
...
         run: python3 hack/id_checker.py
 
+  check-licenses:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: 1.16
+      - name: Install WWHRD
```
arbulu89, line 98: We don't really need to run this in 2 steps, right?
`check-licenses` already depend on `install-wwhrd`, so calling the 2nd would call the 1st by default

Makefile
```diff
@@ -148,3 +148,12 @@ web/frontend/assets/images:
 .PHONY: helm-lint
 helm-lint:
 	docker run --rm -ti --name trento-chart-test -w /workdir -v $(shell pwd):/workdir quay.io/helmpack/chart-testing:v3.4.0 ct lint
+
+.PHONY: install-wwhrd
+install-wwhrd:
+	which wwhrd || go install github.com/frapposelli/wwhrd@latest
```
arbulu89, line 154: I don't know that much about golang dependencies, but I wonder if there is some option to define this dependency as a `development` dependency

arbulu89 (changes requested):
Hey @flaviodsr ,
So comments worth checking.
As a side note, if you don't have your golang `bin` directory in your path the `wwhrd` execution fails

.wwhrd.yml
flaviodsr: done.

.github/workflows/ci-cd.yaml
flaviodsr: you are right, I just thought that in case of failure of the install it would be more clear having a step specifically doing that.

Anyways, I have removed the step.

Makefile
flaviodsr: I am not aware of such option. Did some research and could not find it either.

flaviodsr:
@arbulu89 thanks for the review.

Regarding the golang bin directory, do you have a suggestion on how to fix that? I could add `$GOPATH/bin` to the path of the binary, however there is also the case where it is not defined which defaults to `$HOME/go/bin`. Not sure if it would be worthy going over all the options.

Makefile
dottorblaster: I would @fabriziosestito since he is the lord of that pattern but I think we can actually declare `wwhrd` in the `tools/tools.go` file having it fetched inside deps

PR #743, opened 2022-01-26
SAP system details page tests

arbulu89:
This PR implements the e2e tests for the SAP system details page.

I have added a bit more of testing than the strictly required. For example, I have started playing with state changes, to see that the status badge is properly change. For that, I have added more scenario. They are sent by photofinish. Actually I thought to implement a standalone method to send specific jsons, but well, this way it works too hehe

In essence, the HANA database details page is exactly the same, but I prefer to open a new PR with that, to make the review a bit easier.

### DIFF
--- a/test/e2e/cypress/fixtures/sap-system-details/sap_system_details.feature
+++ b/test/e2e/cypress/fixtures/sap-system-details/sap_system_details.feature
@@ -1,4 +1,4 @@
-Feature: SAP system details vies
+Feature: SAP system details view
     This is where the user has a detailed view of the status of one specific discovered SAP system
 
     Background:
@@ -7,7 +7,6 @@ Feature: SAP system details vies
         # Sid: 'NWD',
         # Hosts: ['vmnwdev01', 'vmnwdev02', 'vmnwdev03', 'vmnwdev04']
         And 4 hosts associated to the SAP system
-
 
     Scenario: Detailed view of one specific SAP system is available
         When I navigate to a specific SAP system ('/sapsystems/a1e80e3e152a903662f7882fb3f8a851')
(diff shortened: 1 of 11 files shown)


test/e2e/cypress/fixtures/sap-system-details/sap_system_details.feature
```diff
@@ -0,0 +1,49 @@
...
+        Given I navigate to a specific SAP system ('/sapsystems/a1e80e3e152a903662f7882fb3f8a851')
+        Then 4 instances are displayed
+        And the data of each instance is correct
+        And the status of each instance is GREEN
+
+    Scenario: SAP system instances status change event is received
+        Given I navigate to a specific SAP system ('/sapsystems/a1e80e3e152a903662f7882fb3f8a851')
+        When a new SAP system event for this system with the 1st instance with a GRAY status is received
+        And the page is refreshed
+        And the status of the 1st instance is GRAY
```
nelsonkopliku, line 30: Is this an intended `And` or is it a typo and it is actually a `Then`?

nelsonkopliku (commented):
Wow man, this is really nice!

I do really appreciate the effort in writing `feature` files, you're doing a great job, thanks!
Helps a lot in understanding the test.

Waiting for the build to succeed :wink:

test/e2e/cypress/fixtures/sap-system-details/sap_system_details.feature
arbulu89: Good catch, The last one should be a `Then`

arbulu89:
More things:
- I have added an `artifact-uploads` steps if the e22 tests fail. It comes handy to check the video
- Update `sap_system` service to return the instances ordered by instance_number. This was the reason why the tests were failing, Now, they are always coming in the same order, and ordering by instance number makes sense in my opinion

nelsonkopliku (approved):
Aha! Ordering! We all had the same issue :smile:

Issue #742, opened 2022-01-26
Projected events are skipped if events are coming almost in parallel


arbulu89:
If the discovery events of 1 agent are coming in a extreme fast pace, so events can be discarded as the sequentialiation of the events fails. This can lead to store incorrect data.

Here an example:
```
INFO[2022-01-26 15:02:28] Projector: clusters is interested in ha_cluster_discovery. Projecting event: 51925 

2022/01/26 15:02:28 github.com/trento-project/trento/web/datapipeline/projector.go:54 record not found
[0.279ms] [rows:0] SELECT * FROM "subscriptions" WHERE "subscriptions"."agent_id" = '7269ee51-5007-5849-aaa7-7c4a98b0c9ce' AND "subscriptions"."projector_id" = 'clusters' ORDER BY "subscriptions"."last_projected_event_id" LIMIT 1 FOR UPDATE
INFO[2022-01-26 15:02:28] HTTP request                                  client="::1" instance=internal latency=4.737824ms method=POST status=202 uri=/api/collect user_agent=
INFO[2022-01-26 15:02:28] Projecting event: 51926                      
INFO[2022-01-26 15:02:28] Projector: hosts is interested in host_discovery. Projecting event: 51926 
INFO[2022-01-26 15:02:28] Projector: host_telemetry is interested in host_discovery. Projecting event: 51926 
INFO[2022-01-26 15:02:28] Projector: hosts is interested in ha_cluster_discovery. Projecting event: 51925 
WARN[2022-01-26 15:02:28] Projector: hosts received an old event: ha_cluster_discovery 7269ee51-5007-5849-aaa7-7c4a98b0c9ce 51925. Skipping 
INFO[2022-01-26 15:02:28] HTTP request                                  client="::1" instance=internal latency=18.34933ms method=POST status=202 uri=/api/collect user_agent=
INFO[2022-01-26 15:02:28] Projecting event: 51927                      
INFO[2022-01-26 15:02:28] Projector: sapsystems is interested in sap_system_discovery. Projecting event: 51927 

```

In this logs, it is visible that the events `51925` and `51926` are coming almost in parallel. So, the second event is handled 1st, making the second an old event, which is skipped.

This scenario is pretty easy to reproduce populating the DB with `photofinish` with the `healthy-27-node-SAP-cluster` scenario.
Mostly check the cluster entry in the hosts, which is the one that fails. You would notice that some host elements don't have an associated cluster, even though they should.
Run `photofinish` several times to reproduce it.

PR #744, opened 2022-01-26
Remove skipping for an older event

fabriziosestito:
To deal with the unlikely scenario of an agent sending multiple events of the same type at the same time, a skip event condition was introduced.
This wasn't working properly since it did not consider that an agent can send different types of events almost in parallel and it was just causing side effects when loading fixtures with `photofinish`.

Closes: #742


arbulu89 (approved):
Thanks!
I will apply this change in my local env to see if it totally remove the issues during my new tests

fabriziosestito:
> Thanks! I will apply this change in my local env to see if it totally remove the issues during my new tests

@arbulu89 l will hold the merge till confirmation

nelsonkopliku (approved):
LGTM

dottorblaster (approved):
LGTM as well

PR #745, opened 2022-01-26
Write down some documentation in the README about Cypress

dottorblaster:
We always said to properly give dignity to our E2E testing setup inside the README, let's just begin! If I forgot something, shoot and I'll add it :smile:


nelsonkopliku (approved):
Thanks for this :pray:

arbulu89 (approved):
Perfect! Thanks

PR #747, opened 2022-01-27
Run npx prettier formatting on e2e test files

arbulu89:
Run `npx prettier --write .` to format all the files within `test/e2e` to apply the latest `prettierrc` configuration file options.

I don't think it needs a real review, as it doesn't have more changes than formatting ones

**For reviewers: Review only the latest 2 commits. The 1st one is only the formatting**


dottorblaster:
Nice catchj actually!

dottorblaster (approved):
eslint too! Great addition! :muscle: LGTM

PR #748, opened 2022-01-27
Add load config to install-server script

fabriziosestito:
This PR adds the possibility to configure the server installer script by a `/etc/trento/installer.conf` file, by overriding env variables.
This is needed in order change the behavior of the installer script in different trento flavors (e.g. Premium)

### DIFF
--- a/install-server.sh
+++ b/install-server.sh
@@ -77,7 +77,7 @@ cmdline() {
             ;;
 
         r)
-            readonly ROLLING=true
+            ROLLING=true
             ;;
 
         *)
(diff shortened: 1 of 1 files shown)


nelsonkopliku (approved):
LGTM

If I understand correctly in `/etc/trento/installer.conf` we can override any of the env variable for the installer script.
For instance the place where to get the helm chart from, thus allowing the premium installation to get helm charts from a non IBS place.

:+1:

arbulu89 (approved):
Do you have some example of the `installer.conf` file?
I guess that it specially matters the `trento_chart_path` variable for this case, right?

dottorblaster (approved):
Cool :0

fabriziosestito:
> Do you have some example of the `installer.conf` file? I guess that it specially matters the `trento_chart_path` variable for this case, right?

@arbulu89 sent you an example on slack

PR #753, opened 2022-01-28
Add connection retry when starting Web and Runner

flaviodsr:
* In case of starting trento-web and the connection with the database
fails, instead of failing immediately, retry to connect to the database
8 times with a variable backoff starting with 3 seconds.

* In case of starting the trento-runner and the connection with the
trento server fails, instead of failing immediately, retry to connect
to the trento server 8 times with a variable backoff starting with 5
seconds.

The above retries allows configuring the trento web and runner pods
with probes so the container does not fail and reports errors while
the database is still initializing.


fabriziosestito (approved):
LGTM!

nelsonkopliku (approved):

rtorrero:
LGTM, thanks @flaviodsr !

Issue #755, opened 2022-01-31
web pod crashing when receiving unexpected data


abravosuse:
A ramp-up customer has reported the agent is giving “connection refused to :8081” errors during discovery:
```
Jan 28 19:53:03 ua00118t trento[9984]: time="2022-01-28 19:53:03" level=info msg="Identifying if the VM is running in a cloud environment..."

Jan 28 19:53:03 ua00118t trento[9984]: time="2022-01-28 19:53:03" level=info msg="VM is running on azure"

Jan 28 19:53:03 ua00118t trento[9984]: time="2022-01-28 19:53:03" level=error msg="Error while running discovery 'cloud_discovery': Post \http://ua00114t.xxxxxxx.com:8081/api/collect\: dial tcp 10.89.199.43:8081: connect: connection refused"
```

On the server side, they observe the web pod is in status CrashLoopBackOff:
```
ua00114t:/sapsw/unix/trento/server # k3s kubectl get pods --all-namespaces

NAMESPACE     NAME                                     READY   STATUS             RESTARTS        AGE

...

default       ua00114t-web-558c65c9cc-kt7rq            0/1     CrashLoopBackOff   11 (2m3s ago)   3h1m
```

And the log file of the web pod shows a panic message:

```
time="2022-01-28 20:01:28" level=info msg="Projecting event: 33"

time="2022-01-28 20:01:28" level=info msg="Projector: hosts is interested in cloud_discovery. Projecting event: 33"

panic: interface conversion: interface {} is nil, not string
```

PR #756, opened 2022-01-31
Fix parse azure cloud data

fabriziosestito:
This PR refactors the azure cloud data parsers function in the hosts projector to use `mapstructure` instead of manually casting the nested interface/map fields.
It also adds asserts to the host projector cloud discovery handler  test.

Related to: #755

### DIFF
--- a/web/datapipeline/hosts_projector.go
+++ b/web/datapipeline/hosts_projector.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"net"
 
+	"github.com/mitchellh/mapstructure"
 	log "github.com/sirupsen/logrus"
 
 	"github.com/trento-project/trento/internal/cloud"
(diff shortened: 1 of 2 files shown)


dottorblaster (approved):
Half of the lines and a test added too. Love this!

nelsonkopliku (approved):
:rocket:

web/datapipeline/hosts_projector.go
```diff
@@ -128,67 +129,22 @@ func parseCloudData(provider string, metadata interface{}) *entities.AzureCloudD
...
+		log.Errorf("can't decode azure metadata: %s", err)
+		return entities.AzureCloudData{}
 	}
 
-	storageProfile, ok := computeInfo["storageProfile"].(map[string]interface{})
-	if !ok {
-		log.Errorf("can't decode storage profile")
-		return cloudData
+	return entities.AzureCloudData{
+		VMName:          azureMetadata.Compute.Name,
```
arbulu89, line 141: mmm, I have some doubts about this.
Could we have the scenario where some of this fields don't exist and we have some other issue.
Or they are automatically filled with empty data?
The added test doesn't cover this situation at least

fabriziosestito: it should work because the fields are not pointers so it will take the golang "zero value".

fabriziosestito: @arbulu89 I've added a test, i still believe it is a bit anemic and we are basically testing that golang works but we can keep it as safety helmet

PR #762, opened 2022-02-01
E2e test cluster overview

rtorrero:
This PR mainly covers e2e tests for the cluster details (`/clusters`) view in Trento, but also
  - Fixes the issue where the SIDs filter input would show an "empty" SID from the clusters that don't have one.
  - Improves the fixtures, so that not all the cluster names are duplicated
  - Adds loadCheckResults to the commands so that we can submit test results from the fixtures to simulate a finished check run.
  
Also, in order to not always follow the happy path, some `critical` or `warning` checkResults have been added. 

Note that still more tests can be added for the content but for now, this already provides some good coverage of the most basic stuff.

### DIFF
--- a/test/e2e/cypress/fixtures/healthy-27-node-SAP-cluster/0a055c90-4cb6-54ce-ac9c-ae3fedaf40d4_ha_cluster_discovery.json
+++ b/test/e2e/cypress/fixtures/healthy-27-node-SAP-cluster/0a055c90-4cb6-54ce-ac9c-ae3fedaf40d4_ha_cluster_discovery.json
@@ -100,7 +100,7 @@
             {
               "Id": "cib-bootstrap-options-cluster-name",
               "Name": "cluster-name",
-              "Value": "hana_cluster"
+              "Value": "hana_cluster_1"
             },
             {
               "Id": "cib-bootstrap-options-stonith-enabled",
@@ -426,7 +426,7 @@
         }
       ]
     },
-    "Name": "hana_cluster",
+    "Name": "hana_cluster_1",
     "Crmmon": {
       "Nodes": [
         {
--- a/test/e2e/cypress/integration/clusters_overview.js
+++ b/test/e2e/cypress/integration/clusters_overview.js
@@ -1,0 +1,306 @@
+import {
+  allClusterNames,
+  allClusterIds,
+  clusterIdByName,
+} from '../fixtures/clusters-overview/available_clusters';
+
+context('Clusters Overview', () => {
+  const availableClusters = allClusterNames();
+  const availableClustersId = allClusterIds();
+  before(() => {
+    cy.resetDatabase();
+    cy.loadScenario('healthy-27-node-SAP-cluster');
+    cy.loadChecksCatalog('checks-catalog/catalog.json');
+    cy.loadChecksResults(
+      'clusters-overview/checks_results_critical.json',
+      '04b8f8c21f9fd8991224478e8c4362f8'
+    );
+    cy.loadChecksResults(
+      'clusters-overview/checks_results_warning.json',
+      '4e905d706da85f5be14f85fa947c1e39'
+    );
+    cy.loadChecksResults(
+      'clusters-overview/checks_results_passing.json',
+      '9c832998801e28cd70ad77380e82a5c0'
+    );
+
+    cy.visit('/');
+    cy.navigateToItem('Clusters');
+    cy.url().should('include', '/clusters');
+  });
+
+  describe('Registered Clusters should be available in the overview', () => {
+    it('should show 9 of the 9 registered clusters with default pagination settings', () => {
+      cy.get('.tn-clustername').its('length').should('eq', 9);
+    });
+    it('should show 9 as total items in the pagination controls', () => {
+      cy.get('.pagination-count').should('contain', '9 items');
+    });
+    it('should have 1 pages', () => {
+      cy.get('.page-item').its('length').should('eq', 3); // We add +2 to the page count because of the first and last page
+    });
+    describe('Discovered clusternames are the expected ones', () => {
+      availableClusters.forEach((clusterName) => {
+        it(`should have a cluster named ${clusterName}`, () => {
+          cy.get('.tn-clustername').each(($link) => {
+            const displayedClusterName = $link.text().trim();
+            expect(availableClusters).to.include(displayedClusterName);
+          });
+        });
+      });
+    });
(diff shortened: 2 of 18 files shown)


test/e2e/cypress/integration/clusters_overview.js
```diff
@@ -0,0 +1,306 @@
...
+        });
+      });
+    });
+  });
+
+  describe('Health Detection', () => {
+    describe('Health Container shows the health overview of all Clusters', () => {
+      it('should show health status of the entire cluster of 9 hosts with partial pagination', () => {
+        cy.reloadList('clusters', 10);
+        cy.get('.health-container .health-passing').should('contain', 1);
```
arbulu89, line 58: Maybe we could check that warning and critical are 0 here. Otherwise, the test is a bit redundant

test/e2e/cypress/integration/clusters_overview.js
```diff
@@ -0,0 +1,306 @@
...
+          it(`should show ${
+            expectedClustersWithThisHealth || 'an empty list of'
+          } clusters when filtering by health '${health}'`, () => {
+            cy.intercept('GET', `/clusters?per_page=100&health=${health}`).as(
+              'filterByHealthStatus'
+            );
+            const selectedOption = `#bs-select-1-${index}`;
+            cy.get(selectedOption).click();
+            cy.wait('@filterByHealthStatus').then(() => {
+              expectedClustersWithThisHealth == 0 &&
```
arbulu89, line 138: Why this `== 0` and `> 0` comparisons?
It is not enough to check the `expectedClustersWithThisHealth` value?
In fact, this `expectedClustersWithThisHealth` will always be `1` , or the value we have in the map. We have control over it. I wouldn't do this strange checkings, it might create some false positives

arbulu89: In fact, you do so in the next `Filtering by SAP system` test. It looks cleaner and easier to understand

test/e2e/cypress/integration/clusters_overview.js
```diff
@@ -0,0 +1,306 @@
...
+        cy.intercept('GET', '/api/tags?resource_type=clusters').as(
+          'filterRefreshed'
+        );
+
+        // Clear crap
+        cy.get('h1').click();
+
+        // We click on the tag button to open the tag modal in the first cluster
+        // and submit the tag
+        cy.get(
```
arbulu89, line 248: Coul we have this on a loop?
The code is simply duplicated, right?

test/e2e/cypress/integration/clusters_overview.js
```diff
@@ -0,0 +1,306 @@
...
+        cy.get('.dropdown-item').contains(tag).click();
+        cy.wait('@filterByTags').then(() => {
+          cy.get('.tn-clustername').should('have.length', 2);
+        });
+
+        cy.get('.tn-filters > :nth-child(6) > .btn').click();
+
+        cy.intercept('DELETE', `${tag}`).as('firstTagRemoved');
+        cy.get(
+          `#cluster-${availableClustersId[0]} > .tn-cluster-tags > .tagify > .tagify__tag > .tagify__tag__removeBtn`
```
arbulu89, line 285: I think we have a `x` class for the removing element. We could use it, instead of this long class queue

web/templates/blocks/clusters_table.html.tmpl
```diff
@@ -15,17 +15,19 @@
...
             <tbody>
             {{- range . }}
-                <tr>
+                <tr id="cluster-{{ .ID }}" class="cluster-{{ .Name }}">
                     <td class="row-status">{{ template "health_icon" .Health }}</td>
                     <td>
                         {{- if .HasDuplicatedName }}
                             <i class="eos-icons eos-18 text-info" data-toggle="tooltip" data-original-title="This cluster has a duplicated name">info</i>
                         {{- end }}
+                        <span class="tn-clustername">
```
arbulu89, line 24: This `span` looks a bit suspicious. We have not used in any of the other links in tables.
If you need the class, why don't simply add it to the `a` element.

arbulu89 (changes requested):
Hey @rtorrero ,
Great!
I have noted many comments, which many of them are just typos and "optional" improvements

test/e2e/cypress/integration/clusters_overview.js
rtorrero: Don't ask me why, I tried several times using this method (the same method used in the hosts overview) but without any luck :/

test/e2e/cypress/fixtures/healthy-27-node-SAP-cluster/0a055c90-4cb6-54ce-ac9c-ae3fedaf40d4_ha_cluster_discovery.json
```diff
@@ -100,7 +100,7 @@
             {
               "Id": "cib-bootstrap-options-cluster-name",
               "Name": "cluster-name",
-              "Value": "hana_cluster"
+              "Value": "hana_cluster_1"
```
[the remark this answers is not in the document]
rtorrero: Mostly to avoid that all clusters have the "duplicated cluster" warning on.

test/e2e/cypress/integration/clusters_overview.js
rtorrero: I agree this test is a bit useless as it's testing a pagination value of 10 vs 100 where in any case, all of them are going to be shown since we have just 9 clusters. Improving it a bit to have it ready in case we want to add more data to the fixtures, but I think it should be already a good improvement as is for now.

web/templates/blocks/clusters_table.html.tmpl
rtorrero: Yep, it's ugly, it's there because of the `{{- if ne .ClusterType "Unknown" }}`
that in case of an unknown cluster type, it wouldn't add the `<a>` element to the DOM. I wanted to keep changes to the templates minimal as it's outside the scope of the PR but we can improve this in the future for sure

arbulu89 (approved):
Hey @rtorrero ,
Once the issues in the`feature` file are fixed, we can merge.

Check the next scenarios:
- `Filtering the Clusters Overview by Cluster name`
- `Filtering the Clusters Overview by Tags`

rtorrero:
Ok, the last minor comments have been addressing, time to merge!

PR #768, opened 2022-02-03
Adds feedback collector

nelsonkopliku:
Experiment proposal for feedback collection.
The **Report feedback** button allows the customer using trento to provide direct feedback.

What do you think? Should it be available only for Premium flavor? :thinking: 

image

image

### DIFF
--- a/web/frontend/stylesheets/override.scss
+++ b/web/frontend/stylesheets/override.scss
@@ -140,3 +140,22 @@ footer.footer-side-menu ul.footer-list {
 .eos-checkbox ~ label::after {
   text-indent: 0;
 }
+
+.feedback-tool-button {
+  background-color: #0c322c;
+  color: #fff;
+  position: fixed;
+  top: 50%;
+  right: -56px;
+  -webkit-transform: rotate(-90deg);
+  -moz-transform: rotate(-90deg);
+  -o-transform: rotate(-90deg);
+  -ms-transform: rotate(-90deg);
+  transform: rotate(-90deg);
+}
+
+.feedback-tool-button:hover,
+.feedback-tool-button:focus {
+  color: #fff;
+  background-color: #0c322c;
+}
--- a/web/templates/blocks/header.html.tmpl
+++ b/web/templates/blocks/header.html.tmpl
@@ -23,5 +23,8 @@
         <link href="/static/frontend/assets/stylesheets/tagify.css" rel="stylesheet" type="text/css" />
         {{ block "additional_scripts" . }}{{ end }}
         <link rel="stylesheet" type="text/css" href="/static/frontend/assets/stylesheets/override.css"/>
+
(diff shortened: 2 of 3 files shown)


stefanotorresi (changes requested):
My 2 cents:

- This is certainly in scope and would be a great addition! Although in our backlog this was not exactly planned for now. 😛 
- The styling _must_ be customized to make it consistent with the rest of the design system.
- SUSE has customer support system, and we have support teams who are responsible for handling user feedback. Do we really want to completely bypass that and put ourselves in the first line? Other products have a similar functionality which probably hooks to SUSE service desk instead of Jira directly.
- Ask the SUMA team if and how are they doing such a thing. I'd be surprised if they didn't have something similar.

stefanotorresi:
more feedback from Diego:
- we should make _very_ clear for the user that this feature must not be used for issuing enterprise support requests, just ideas and feedback for which there is no contract obligation like SLAs and such.

arbulu89:
@nelsonkopliku What is this actually doing? Or is it just the view itself?
And for more tactical questions:
- Do we have someone who will take care of the responses?
- The modal asks for `name/email`,etc. This would most probably require GPDR things, etc.

nelsonkopliku:
> * This is certainly in scope and would be a great addition! Although in our backlog this was not exactly planned for now. stuck_out_tongue
>
Fair enough :smile: it's just been part of the boards spike, and it was pretty cheap, and i am not going to put much more effort on it :sweat_smile: 

> * The styling _must_ be customized to make it consistent with the rest of the design system.
>
How would you like it? Is this better?

image

nelsonkopliku:
> What is this actually doing? Or is it just the view itself?
> 
it is a jira funcionality that allows you to embed feedback collection forms like this.
Whenever there's a submission of a feedback, it goes as an item in a dedicated backlog space that generates data.

> Do we have someone who will take care of the responses?
>
It does not necessarily need too frequent activity, yet regular even if not really frequent.
It'd be a way to collect feedbacks more directly from who is using Trento and try to understand the users.

> The modal asks for `name/email`,etc. This would most probably require GPDR things, etc.
> I would be careful with this implementation and only provide it once we have all the tactical/legal things answered
>
really good point. They are not required, yet I'd rather remove those, but I cannot actually find a way to do that :confused: 


It just came quite free when playing with jira backlog :sweat_smile:

stefanotorresi:
Ok, so, before we merge this, we need to acknowledge that this implementation has a problem we can't solve now: the styling and structure of the form cannot be fully customized. This means we will have this feature with a completely different color palette and design language.
OTOH, I don't want to stall the feature because of this; after all, it is a very simple thing which could give us a lot of value, especially at this early stage of development.
While we could still merge this as is, before doing that I'd like us to acknowledge that we will sooner or later need to decommission this feature in favour of something similar with a properly integrated UI/UX.

nelsonkopliku:
@stefanotorresi 
green again
image

I will create followup backlog item :wink:

dottorblaster:
:shipit:

stefanotorresi (approved):

PR #771, opened 2022-02-04
Spike on prometheus http service discovery

arbulu89:
**PoC PR. Not merge!! Only for code discussions**
This PoC aims to spike on the prometheus HTTP service discovery: 
https://prometheus.io/docs/prometheus/latest/http_sd/

In a nutshell, the HTTP service discovery allows us to modify the prometheus scraping targets (nodes where metrics are generated) based on a HTTP Get call (that in this PoC is attached to the Web api endpoint).

This way, we collect all the hosts and information (ssh address) and generate the targets content.

Once this is done, we would have all the metrics from the node_exporter in the prometheus.

The 2nd step is to check if the exporter is working properly or not. For that, I have used the `up` method, which basically checks if the exporter is up or not (returning 1 if it is up, and 0 otherwise). With this, I can display in the Web host details page the exporter state. We could query any other information in the same exact way.

In this example, the prometheus query is done by the web application itself.

Here a pic:
image

In order to use it, we would need to configure prometheus to use the http sd. Find this config as an example:

```
scrape_configs:
  # we use job_name to group exporters for each cluster
  - job_name: http_sd_hosts
    scrape_interval: 30s
    scrape_timeout: 30s
    metrics_path: /proxy
    params:
      module:
        - node
    http_sd_configs:
    - url: http://localhost:8080/api/prometheus_http_sd
```

This would scrape the targets discovered by that endpoint. The `module.node` is needed as I have used the prometheus [exporter exporter](https://github.com/QubitProducts/exporter_exporter) exporter to group all the different exporters in 1 endpoint (we could put there the cluster and sap exporters, so we only scrape in one port).

So in this case, the exporter exporter is configured like:
```
modules:
  node:
    method: http
    http:
       port: 9100
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,6 +14,8 @@ require (
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/mitchellh/mapstructure v1.4.3
 	github.com/pkg/errors v0.9.1
+	github.com/prometheus/client_golang v1.4.0
+	github.com/prometheus/common v0.9.1
 	github.com/shirou/gopsutil v3.21.11+incompatible
 	github.com/sirupsen/logrus v1.8.1
 	github.com/spf13/afero v1.8.0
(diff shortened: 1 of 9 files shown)


arbulu89:
Closing the draft PR as it has been validated. It will follow a better structured code

PR #780, opened 2022-02-08
Add prometheus grafana to helm chart

fabriziosestito:
This PR adds Prometheus and Grafana to the helm chart.
The chart includes:
- prometheus configuration that uses our trento web `http_sd` endpoint to grab the target/exporter list
- grafana uses the sidecar container configuration to scrape configmaps and get the datasource configuration
- the ingress now forwards grafana to the `/grafana` path so it can be embedded in an iframe
- grafana has anonymous users active for the requirement above
- helm chart dev is updated to forward prometheuse on port `9090` for local development

To get the grafana password:
`kubectl get secret <release-name>-grafana -o jsonpath="{.data.admin-password}" | base64 --decode ; echo`


packaging/helm/trento-server/templates/prometheus_configmap.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Release.Name }}-prometheus-configmap
+data:
+  prometheus.yml: |-
```
arbulu89, line 6: Most probably we will need to change this on the road. But I guess it is totally fine by now

arbulu89 (approved):
Great @fabriziosestito 
We will work on the changes of the prometheus configuration file as they come

nelsonkopliku (approved):

dottorblaster (approved):
LGTM

PR #783, opened 2022-02-09
Add exporter deps

rtorrero:
This PR just adds `node_exporter` as a dependency in both the trento RPM spec and the agent systemd unit to ensure it's started in the proper order. 

I also applied tok the opportunity to format the ci-cd.yaml using a formatter.

### DIFF
--- a/.github/workflows/ci-cd.yaml
+++ b/.github/workflows/ci-cd.yaml
@@ -3,9 +3,9 @@ concurrency: ci-${{ github.ref }}
 on:
   push:
     tags-ignore:
-      - '*'
+      - "*"
     branches:
-      - 'main'
+      - "main"
   pull_request:
   release:
     types: [published]
@@ -134,7 +134,6 @@ jobs:
         with:
           name: e2e-videos
           path: test/e2e/cypress/videos/
-
 
   build-and-push-container-images:
     runs-on: ubuntu-20.04
@@ -296,10 +295,10 @@ jobs:
         with:
           name: trento-binaries
           path: |
-                  build/trento-amd64.tgz
-                  build/trento-arm64.tgz
-                  build/trento-ppc64le.tgz
-                  build/trento-s390x.tgz
+            build/trento-amd64.tgz
+            build/trento-arm64.tgz
+            build/trento-ppc64le.tgz
+            build/trento-s390x.tgz
 
   release-rolling:
     needs: [test-binary, test-checks, test-e2e]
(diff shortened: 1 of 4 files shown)


stefanotorresi:
Indulge me in an attempt to spark a deep discussion: could this be the chance to split the `trento` package into `trento-web`, `trento-runner` and `trento-agent`, or would that be overkill?

arbulu89 (approved):
It is good enough for me by now.
This opens up again the discussion we had about "Is prometheus a hard requirement for trento? Does it work (at least partially) without it?".
Right now, we are making it a hard requirement. Just to be conscious about that (nothing wrong with it).

arbulu89:
> Indulge me in an attempt to spark a deep discussion: could this be the chance to split the `trento` package into `trento-web`, `trento-runner` and `trento-agent`, or would that be overkill?

That requires a lot of side work, specially in the packages submission side. I wouldn't start with it. If for some reason we want it to do this (I have my doubts if it is a good idea), we should plan it as a task that might take some important time

fabriziosestito (approved):

dottorblaster (approved):

.github/workflows/ci-cd.yaml
```diff
@@ -375,6 +380,7 @@ jobs:
           set -ex
           for target_host in ${TRENTO_AGENT_HOSTS//,/ }
           do
+            ssh "$TRENTO_USER@$target_host" "zypper in -y golang-github-prometheus-node_exporter"
```
arbulu89, line 383: By the way @rtorrero ,
Here we would need to start the prometheus systemd daemon, right?
`systemctl start prometheus-node_exporter`

rtorrero: Not really, it will be automatically started with the `sudo systemctl enable --now trento-agent.service` that we have on line :385.

Notice that as you said; this will make node_exporter a hard dependency for trento, causing trento-agent.service to fail when node_exporter doesn't get started.

rtorrero:
> Indulge me in an attempt to spark a deep discussion: could this be the chance to split the `trento` package into `trento-web`, `trento-runner` and `trento-agent`, or would that be overkill?

I agree with Xabi here, it's going to cause some changes in how we submit packages. I'd rather tackle that one in a near future when exploring the new GitHub <-> OBS integration features that have been recently added (not sure if it's going to work for us, but worth exploring)

.github/workflows/ci-cd.yaml
arbulu89: Ok, thanks @rtorrero ,
I thought that the `tgz` installation wasn't installing the service file. But yes, it does hehe
Perfect them!

PR #784, opened 2022-02-10
Allows Grafana dashboards to be embedded

nelsonkopliku:
In order to be able to embed grafana iframes, we need to tell grafana to allow that.


fabriziosestito:
Thanks i was struggling with this

dottorblaster (approved):

rtorrero (approved):
LGTM

PR #791, opened 2022-02-14
Change trento path in the Dockerfile

fabriziosestito:
This PR changes the path of the trento binary in the containers to the same used for premium `/usr/bin/path` so we can use the same dump script for both the flavors.


dottorblaster (approved):
Oh well, fair enough

arbulu89 (approved):

nelsonkopliku (approved):

PR #793, opened 2022-02-14
Do not make assumptions about the shape of the payload of checks catalog

dottorblaster:
This prevents the checks selection from panicking in the frontend when the catalog is empty.


arbulu89 (approved):
Nice! Error handling is always needed.
In the other hand, this won't fix the real issue, right? The one that for some reason the checks are not loaded

dottorblaster:
@arbulu89 exactly! I have to debug a running deployment for that, let's see what I discover

nelsonkopliku (approved):

PR #796, opened 2022-02-15
Attach system replication status badge on secondary node

nelsonkopliku:
A feedback from the customers says: The system replication is a characteristic of secondary nodes.

And here's the PR addressing it.

image

### DIFF
--- a/test/e2e/cypress/fixtures/sap-systems-overview/available_sap_systems.js
+++ b/test/e2e/cypress/fixtures/sap-systems-overview/available_sap_systems.js
@@ -1,3 +1,11 @@
+const HANASystemReplicationModes = {
+  Primary: "HANA Primary",
+  Secondary: "HANA Secondary"
+}
+
+export const isHanaSecondary = (instance) => instance.systemReplication == HANASystemReplicationModes.Secondary
+export const isHanaPrimary = (instance) => instance.systemReplication == HANASystemReplicationModes.Primary
+
 export const availableSAPSystems = [
   {
     sid: 'NWD',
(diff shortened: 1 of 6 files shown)


fabriziosestito:
Great! Maybe we could leave the sap systems test in place and just check for 200 plus couple of strings without regex, wdyt?

nelsonkopliku:
> Great! Maybe we could leave the sap systems test in place and just check for 200 plus couple of strings without regex, wdyt?
>

Ok to me :wink:

dottorblaster (approved):
LGTM

arbulu89 (changes requested):
Hey @nelsonkopliku ,
I'm not sure if this is the correct solution. 
The system replication from the secondary should come from the secondary node data.

Here some examples (all of this information should be available in the raw data, in the cibadmin part):
Primary node:
```
<nvpair id="status-1-master-rsc_SAPHana_PRD_HDB00" name="master-rsc_SAPHana_PRD_HDB00" value="150"/>
          <nvpair id="status-1-hana_prd_version" name="hana_prd_version" value="2.00.050.00.1592305219"/>
          <nvpair id="status-1-hana_prd_clone_state" name="hana_prd_clone_state" value="PROMOTED"/>
          <nvpair id="status-1-hana_prd_sync_state" name="hana_prd_sync_state" value="PRIM"/>
          <nvpair id="status-1-hana_prd_roles" name="hana_prd_roles" value="4:P:master1:master:worker:master"/>
```

Secondary node:
```
<nvpair id="status-2-hana_prd_clone_state" name="hana_prd_clone_state" value="UNDEFINED"/>
          <nvpair id="status-2-hana_prd_version" name="hana_prd_version" value="2.00.050.00.1592305219"/>
          <nvpair id="status-2-hana_prd_roles" name="hana_prd_roles" value="1:S:master1::worker:"/>
          <nvpair id="status-2-master-rsc_SAPHana_PRD_HDB00" name="master-rsc_SAPHana_PRD_HDB00" value="0"/>
          <nvpair id="status-2-hana_prd_sync_state" name="hana_prd_sync_state" value="SFAIL"/>

```

I think, here, what we need to do is use the `hana_prd_sync_state` state value from the secondary node. All the information should be there, and not in the primary node.

I hope, I was able to explain it (but most probably I couldn't). It is a tricky thing, and I'm not an expert in the cluster neither, so we might need to seek help.

PR #797, opened 2022-02-15
Do not add bitnami charts repo from the installer if it's not needed

rtorrero:
This PR allows to skip the addition of the bitnami helm chart repo during the installation; useful for the `server-installer` subpackage in `trento-premium`; where we are using the TGZd `trento-premium-server` chart that already includes dependencies (such as bitnami's postgresql chart)

To be clear: Even though the repo was being added, it wasn't used as the script already skipped the download in such a case. In any case, it's cleaner like this so it's worth adding.

### DIFF
--- a/install-server.sh
+++ b/install-server.sh
@@ -192,6 +192,10 @@ install_helm() {
 }
 
 update_helm_dependencies() {
+    local download_chart=${DOWNLOAD_CHART:-true}
+    if [[ "$download_chart" != true ]]; then
+        return
+    fi
     echo "Updating Helm dependencies..."
     helm repo add bitnami https://charts.bitnami.com/bitnami >/dev/null
     helm repo update >/dev/null


fabriziosestito (approved):
amazing lgtm

arbulu89 (approved):
Thanks @rtorrero

dottorblaster (approved):
:heart_eyes_cat:

install-server.sh
```diff
@@ -192,9 +192,12 @@ install_helm() {
...
 update_helm_dependencies() {
-    echo "Updating Helm dependencies..."
-    helm repo add bitnami https://charts.bitnami.com/bitnami >/dev/null
-    helm repo update >/dev/null
+    local download_chart=${DOWNLOAD_CHART:-true}
+    if [[ "$download_chart" == true ]]; then
+        echo "Updating Helm dependencies..."
+        helm repo add bitnami https://charts.bitnami.com/bitnami >/dev/null
+        helm repo update >/dev/null
+    fi
```
stefanotorresi, line 200: nitpick: consider reversing the if condition and returning early to keep the happy path outside the if block.

nelsonkopliku (approved):

ajaeger (approved):

ajaeger (approved):

PR #802, opened 2022-02-16
Add Grafana initialization

fabriziosestito:
This PR adds Grafana initialization and dashboard boostrap to the running environment.
Since an admin password is needed to generate a token to consume the dashboard API, a secret is generated through a template function when the chart is installed the first time, and then passed to the deployment to be consumed by the `trento-web` container.

### DIFF
--- a/cmd/web/web.go
+++ b/cmd/web/web.go
@@ -46,6 +46,10 @@ func addServeCmd(webCmd *cobra.Command) {
 	var key string
 	var ca string
 
+	var grafanaURL string
+	var grafanaUser string
+	var grafanaPassword string
+
 	serveCmd := &cobra.Command{
 		Use:   "serve",
 		Short: "Starts the web application",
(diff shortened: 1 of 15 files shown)


dottorblaster (approved):

cmd/web/web.go
```diff
@@ -61,6 +65,10 @@ func addServeCmd(webCmd *cobra.Command) {
 	serveCmd.Flags().StringVar(&key, "key", "", "mTLS server key")
 	serveCmd.Flags().StringVar(&ca, "ca", "", "mTLS Certificate Authority")
 
+	serveCmd.Flags().StringVar(&grafanaURL, "grafana-url", "http://localhost:3000", "Grafana URL")
+	serveCmd.Flags().StringVar(&grafanaUser, "grafana-user", "admin", "Grafan user")
```
arbulu89, line 69: Typo `Grafan`

nelsonkopliku (approved):
LGTM.

We need to update env variables in the pipeline at a certain point I guess.
Next PRs

web/app.go
```diff
@@ -225,6 +227,8 @@ func NewAppWithDeps(config *Config, deps Dependencies) (*App, error) {
 }
 
 func (a *App) Start(ctx context.Context) error {
+	grafana.InitGrafana(ctx, a.config.GrafanaConfig)
```
arbulu89, line 230: Nothing wrong with this, but, shouldn't we initialize grafana like other dependencies in `DefaultDependencies`?

internal/grafana/grafana.go
```diff
@@ -0,0 +1,120 @@
...
+	return retry.Do(
+		func() error {
+			log.Info("Initializing Grafana Dashboards")
+			token, err := createToken(config)
+			if err != nil {
+				log.Error("Failed to create Grafana token")
+				return err
+			}
+
+			dashboardsURL := fmt.Sprintf("%s/%s", config.URL, "api/dashboards/db")
```
arbulu89, line 39: I guess that in the future we can change this easily to post more dashboards

arbulu89 (approved):
Great great job!
Added some comments (one of them a typo), but green light from my side

web/app.go
fabriziosestito: I wanted to pass the ctx to the grafana initialization, we might refactor the init db as well because otherwise the retry blocks any `SIG*`

PR #807, opened 2022-02-17
Sap system health computation

arbulu89:
Implement the SAP systems and HANA database health computation. The health is used to display the green/yellow/red icons in the SAP systems list, and in the future in the new home dashboard.

The health is computed in the next way:
- If 1 of the instances in the SAP system is in RED - System is RED
- If 1 of the instances in the SAP system is in YELLOW, and there is not RED - System is YELLOW
- If 1 of the instances in the SAP system is GRAY, and the previous conditions don't match - System is GRAY
- GREEN otherwise

The attached HANA databases health on the SAP systems is used as well to compute the system health

image


arbulu89:
> @arbulu89 didn't read the code, but what happens if the agent cannot collect the info. Like, the system worked before, but suddenly we cannot collect the data anymore? I assume it should raise Red on the system too.

This is out of scope. We have not defined what would happen in that scenario, when we don't have updates, and actually it is not something easy to do. By now, we agreed with @abravosuse to only use the current sapcontrol output to compute the health.

But I guess in the future we can start thinking about what to do when we don't have updates, like using the health of the hosts belonging to this system as additional info, and if we don't have an heartbeat them, set the system health as red/warning.

diegoakechi:
@arbulu89 all good... @abravosuse one more item for backlog ;)

fabriziosestito (approved):

dottorblaster (approved):
:+1: LGTM

nelsonkopliku (approved):

PR #808, opened 2022-02-17
Embed cpu and memory dashboards

nelsonkopliku:
image


arbulu89 (approved):
Thank you @nelsonkopliku ,
I think it is good enough by now. I think on the road, it would be nice to have a `GrafanService`, so we don't pass single variables over the functions.

On the selected graphs, the UX discussion needs still to happen. When I proposed to use tabs, it was thinking that eventually we will add a lot of more graphs (dozens), so having there together with the details might not be ideal. But maybe we don't need more graphs, I don't know

PR #809, opened 2022-02-17
Homepage UI component

dottorblaster:
^ :-))))


stefanotorresi:
screenshot please! 😸

dottorblaster:
@stefanotorresi yes I wanted to see if tests passed first :-D

dottorblaster:
2022-02-22-105842_1003x415_scrot

First attempt, if we don't like it we can refine it later :smile:

arbulu89 (commented):
Thanks @dottorblaster ,
Some initial comments before jumping to the code.
- I think we miss a 1st column like `SAP systems`, after the `SID`. This would consume the `sapsystems_health` value in the API.
- I think `Pacemaker cluster` should be changed to `Pacemaker clusters` (in plural). As we are showing there all the involved clusters (even though by now it is one hehe)
- Code related. I think the values in the database and cluster must be flipped (current 1st column is for database and the 2nd for the clusters, the data is inverted now)

dottorblaster:
I think you're right @arbulu89, I skipped the `sapsystems_health` field as it wasn't in the initial response but I'll add it as soon as possible, thank you!

dottorblaster:
Much better now:

image

abravosuse:
Guys, may I suggest a couple of changes in display/layout?

1. Column "SAP Systems": rename it as "SAP instances". It's the status of the SAP instances that we are collecting under this column.
2. Swipe places of the "Pacemaker cluster" and "Database" columns. It's a matter of association. The database layer is close to the SAP application (instances) layer. The pacemaker cluster is close to the OS  (hosts) layer.

dottorblaster:
image


@abravosuse this way? :grimacing:

abravosuse:
Perfect @dottorblaster . Thanks!

test/e2e/cypress/integration/home.js
```diff
@@ -3,39 +3,9 @@ context('Homepage', () => {
...
-    });
-    it('should provide correct link to the Architecture document', () => {
-      cy.get('#tn-architecture-link').should(
-        'have.attr',
-        'href',
-        'https://github.com/trento-project/trento/blob/main/docs/trento-architecture.md'
-      );
+  describe('The homepage has the global health component visible', () => {
+    it('should display the global health chart', () => {
+      cy.get('#homepage-component').contains('At a glance');
```
arbulu89, line 8: I think we could improve this. At least, with the currently discovered facts, check is we have 3 rows with everything on green. It is the less costly thing we can have, but I find it valuable. I think you can get it less than a half hour hehe

web/frontend/components/HealthSummary/HealthSummary.js
```diff
@@ -0,0 +1,129 @@
+import React from 'react';
+
+const getHealthIcon = (health) => {
+  switch (health) {
+    case 'critical':
+      return <i className="eos-icons health-summary-icon text-danger">error</i>;
+    case 'warning':
+      return (
+        <i className="eos-icons health-summary-icon text-warning">warning</i>
```
arbulu89, line 9: Why in the hell `prettier` has formatted so strangely hehe
Is it due the difference of line length?

web/frontend/components/HealthSummary/HealthSummary.js
```diff
@@ -0,0 +1,129 @@
...
+
+const any = (predicate, label) =>
+  Object.keys(predicate).reduce((accumulator, key) => {
+    if (accumulator) {
+      return true;
+    }
+    return predicate[key] === label;
+  }, false);
+
+const getCounters = (data) =>
```
arbulu89, line 34: I think here we should consider to count the `unknowns`. Otherwise, if some summary entry has unknonws and passes, we will compute this pass, which is not entirely sure. I wouldn't count them as pass at least.
But it is true that we don't have any box for them. @abravosuse What should we do here?
- Leave as it is. So if we have some unknown at worst, as pass
- Add other box for the unknown
- Not put other box, but not count neither

web/frontend/components/HealthSummary/HealthSummary.js
```diff
@@ -0,0 +1,129 @@
...
+            <i className="eos-icons-outlined eos-18 alert-icon">error</i>
+            <div className="alert-body">Critical</div>
+            <span className="badge badge-secondary">{counters.critical}</span>
+          </div>
+        </div>
+
+        <div className="health-summary">
+          <div className="health-summary-header">
+            <span className="health-summary-id health-summary-cell">SID</span>
+            <span className="health-summary-cell">SAP Systems</span>
```
arbulu89, line 88: Did you push the commit where you changed `SAP Systems` by `SAP instances`?
And the order change.
I think that's missing

web/templates/home.html.tmpl
```diff
@@ -1,71 +1,5 @@
...
-               href='https://github.com/trento-project/trento/blob/main/docs/scope.md'>scope</a>,
-            <a id='tn-readme-link' target='blank'
-               href='https://github.com/trento-project/trento/blob/main/README.md'>readme</a>
-            or
-            <a id='tn-architecture-link' target='blank'
-               href='https://github.com/trento-project/trento/blob/main/docs/trento-architecture.md'>architecture</a>
-            documents.
-        </div>
-    </section>
+  <div id="homepage-component"></div>
```
arbulu89, line 3: Question for @abravosuse : Do you think we should move the current home content to somewhere else?
Like the about page or some additional page. Just asking.
It doesn't need to done in this PR anyway

arbulu89 (changes requested):
Hey @dottorblaster ,
I did a more thorough review this time. 
Some comments added. The unique blocker is that I think you forgot to push some commit. At least the latest image you pasted doesn't match with the code.
We can discuss about everything else.
But it looks good to me.

web/frontend/components/HealthSummary/HealthSummary.js
dottorblaster: Yes it is! it looks strange the first times you see it, but it's fine :smile:

nelsonkopliku (commented):
Whoa! Great!

@arbulu89 already raised good points, mainly on the current homepage content and the e2e test.

Apart from those points, we're good I guess :wink:

web/frontend/components/HealthSummary/HealthSummary.js
abravosuse: @arbulu89 for this iteration let's go with the third option: not put other box, but no count neither (I don't want an unknown to be counted as a green). But looking down the road this is how I think it should be:

1. distinguish between stopped and unknown as they are different things (given that grey is the sapcontrol color associated to a stopped status, I suggest we keep grey for stopped and choose another one for unknown. OR, stopped is circle grey and unknown is triangle grey).
2. add as many boxes as possible statuses
3. stablish a clear hegemony among statuses:
I think we all agree that:
- all green > green
- one yellow, rest green > yellow
- one red, rest green/yellow > red
But how stopped or unknown weight in the overall status is something we are going to have to determine component by component. What I mean is that, for example, it's not the same to have an ABAP instance stopped than having the diagnostic agent stopped.
All this to be refined.

web/templates/home.html.tmpl
abravosuse: Yes, I think that eventually we will have to move it somewhere else because the home page should all be about the status of your environment. But I'd rather keep the About page as it it. That way the version info is clearly displayed and there is no confusion about it. Maybe another page under Settings called "Intro" or "First steps". To be refined as well.

web/frontend/stylesheets/stylesheets.scss
```diff
@@ -236,3 +236,36 @@ td i.critical {
...
+}
+
+.health-summary-icon {
+  font-size: 24px;
+  width: 24px;
+}
+
+.health-summary-id {
+  text-align: left;
+}
```
fabriziosestito, line 271: not gonna miss this 😇

dottorblaster:
@arbulu89 if you want you can check again, I wanted to page you privately but the major outage is preventing me to do so... :sweat_smile:

stefanotorresi (approved):
LGTM

web/frontend/components/HealthSummary/HealthSummary.js
arbulu89: @dottorblaster I don't know if you read this thread. I think it is missing to be implemented.
We would need to check before the passing, if there is any `unknown`. So simply adding the next should work:

```
      if (any(element, 'unknown')) {
        return { ...accumulator, unknown: accumulator.unknown + 1 };
      }
```
At least, we won't have a new passing counter added in this case (unknowns won't be showed in the frontend, but at least neither false passes)

web/frontend/components/HealthSummary/HealthSummary.js
```diff
@@ -0,0 +1,129 @@
...
+              return (
+                <div className="health-summary-row" key={id}>
+                  <span className="health-summary-id health-summary-cell">
+                    {sid}
+                  </span>
+                  <span className="health-summary-cell">
+                    {getHealthIcon(sapsystem_health)}
+                  </span>
+                  <span className="health-summary-cell">
+                    {getHealthIcon(clusters_health)}
```
arbulu89, line 111: This is still wrong hehe
I think the 2nd must be `database_health` and the 3rd `clusters_health`

arbulu89 (approved):
Hi @dottorblaster ,
Setting the PR as green (I don't think we need more reviews), but there are 2 things that must be fixed. Have a look on my latest 2 comments

web/frontend/components/HealthSummary/HealthSummary.js
dottorblaster: @arbulu89 thanks for the tip, actually I read the thread but I didn't get that detail :sweat_smile:

nelsonkopliku (approved):

PR #816, opened 2022-02-18
Add health summary api endpoint

fabriziosestito:
This PR adds the health summary API endpoint

The Health summary computation is wrapped in a service.
It also adds the possibility to filter by ID in the `GetAll` methods of cluster and hosts services.


web/sap_systems_api.go
```diff
@@ -0,0 +1,38 @@
...
+	"net/http"
+
+	"github.com/gin-gonic/gin"
+	"github.com/trento-project/trento/web/services"
+)
+
+type HealthSummary struct {
+	ID              string `json:"id"`
+	SID             string `json:"sid"`
+	SAPSystemHealth string `json:"sap_system_health"`
```
nelsonkopliku, line 13: This is ignored, right?

nelsonkopliku (approved):
Great! :rocket:

nelsonkopliku (commented):
When CI goes green :smile:

arbulu89 (changes requested):
Hey @fabriziosestito ,
I think there is something wrong with the health computation. Comment below.
Could you check?

web/models/health_summary.go
```diff
@@ -0,0 +1,18 @@
...
+	HealthSummaryHealthUnknown  = "unknown"
+)
+
+type HealthSummary []SAPSystemHealthSummary
+
+type SAPSystemHealthSummary struct {
+	ID             string `json:"id"`
+	SID            string `json:"sid"`
+	ClustersHealth string `json:"clusters_health"`
+	DatabaseHealth string `json:"databases_health"`
```
arbulu89, line 16: Maybe `DatabasesHealth` better (last `s` missing). Or remove the `s` from `databases_health` as we will only have 1 database per system

web/sap_systems_api.go
```diff
@@ -0,0 +1,27 @@
...
+
+	"github.com/gin-gonic/gin"
+	"github.com/trento-project/trento/web/services"
+)
+
+// ApiSAPSystemsHealthSummaryHandler godoc
+// @Summary Retrieve SAP Systems Health Summary
+// @Accept json
+// @Produce json
+// @Success 200 {object} HealthSummaryResponse
```
arbulu89, line 14: Where is this `HealthSummaryResponse` defined? i cannot find it anywhere...

web/services/health_summary_service.go
```diff
@@ -0,0 +1,128 @@
...
+	case models.SAPSystemHealthWarning:
+		return models.HealthSummaryHealthWarning
+	case models.SAPSystemHealthCritical:
+		return models.HealthSummaryHealthCritical
+	default:
+		return models.HealthSummaryHealthUnknown
+	}
+}
+
+func computeAggregatedClustersHealth(clusters []*models.Cluster) string {
```
arbulu89, line 80: As commented offline, hopefully we can remove all of this duplicated code at some point hehe

web/models/health_summary.go
fabriziosestito: i blame copilot, fixed, thanks!

web/sap_systems_api.go
fabriziosestito: updated with the right struct, thanks!

PR #817, opened 2022-02-18
Implement cluster heatlh computation projection

arbulu89:
Hey folks! This is the improvement for the cluster health computation. It is based on projections in the discovery side and the checks results event. Hopefully it makes the work.
I have written some E2E tests and all goes green (they are not uploaded yet).

Some more details:
The partial healths fields is a map which stores all the partial healths of a resource, cluster in our case (but easy extendable for other resources like sap systems, hosts, etc). So the partial health is updated in the different events (now, when a new cluster discovery is done and the cluster checks results are executed). The 2 healths are stored, and the overall health computed.

I have a doubt with the new table though (`health_states`). I thinking if we should truncate the values of this table on web service boot up. This would have 2 benefits:
- If for some reason we need the change the partial healths labels (`hana_sr_health` and `config_checks` by now), we would get rid off them, and we wouldn't have zombie entries.
- We wouldn't have "old" health states. Maybe it is safe to restore the health to nothing on boot up, to not show old values, as it might be misleading.

### DIFF
--- a/test/e2e/cypress/fixtures/.photofinish.toml
+++ b/test/e2e/cypress/fixtures/.photofinish.toml
@@ -57,3 +57,19 @@ files = ["./sap-systems-overview/1b0e9297-97dd-55d6-9874-8efde4d84c90_sap_system
 [sap-systems-overview-hana-RED]
 
 files = ["./sap-systems-overview/13e8c25c-3180-5a9a-95c8-51ec38e50cfc_sap_system_discovery_RED.json"]
+
+[cluster-4-SOK]
+
+files = ["./clusters-overview/13e8c25c-3180-5a9a-95c8-51ec38e50cfc_ha_cluster_discovery_4_SOK.json"]
+
+[cluster-1-SOK]
+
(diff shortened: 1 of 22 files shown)


fabriziosestito (approved):
LGTM, I think it is the best trade-off with the current architecture.
Not sure about truncating the table at boot i think we could merge this and then have a discussione on it.

nelsonkopliku (approved):
LGTM as well.
Generally speaking I'd strive to find a clearer naming for things like `HANASrHealth4` :smile: 
but it is good enough here.

Not sure also about truncating health on boot, but let's discuss on that.

arbulu89:
> LGTM as well. Generally speaking I'd strive to find a clearer naming for things like `HANASrHealth4` smile but it is good enough here.
> 
> Not sure also about truncating health on boot, but let's discuss on that.

Update at: https://github.com/trento-project/trento/pull/817/commits/1f668a8354918764584f667af76598114a8e1662

PR #823, opened 2022-02-21
Fix health summary api

fabriziosestito:
This PR adds SAPSystemHealth to the health summary API and fixes the health summary service so that it uses only the HANA clusters to compute the cluster aggregate health

### DIFF
--- a/web/services/health_summary_service.go
+++ b/web/services/health_summary_service.go
@@ -34,17 +34,12 @@ func (s *healthSummaryService) GetHealthSummary() (models.HealthSummary, error) 
 	}
 
 	for _, sapSystem := range sapSystems {
-		var clusterIDs []string
 		var hostIDs []string
+		var clusterIDs []string
 
 		for _, instance := range sapSystem.GetAllInstances() {
-			clusterIDs = append(clusterIDs, instance.ClusterID)
 			hostIDs = append(hostIDs, instance.HostID)
-		}
-
-		clusters, err := s.clustersService.GetAll(&ClustersFilter{ID: clusterIDs}, nil)
-		if err != nil {
-			return nil, err
+			clusterIDs = append(clusterIDs, instance.ClusterID)
 		}
 
 		hosts, err := s.hostsService.GetAll(&HostsFilter{ID: hostIDs}, nil)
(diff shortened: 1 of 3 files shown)


nelsonkopliku (approved):

web/services/health_summary_service.go
```diff
@@ -34,38 +34,49 @@ func (s *healthSummaryService) GetHealthSummary() (models.HealthSummary, error)
...
 		}
 
-		hosts, err := s.hostsService.GetAll(&HostsFilter{ID: hostIDs}, nil)
-		if err != nil {
-			return nil, err
+		var hanaClusters []*models.Cluster
+		if sapSystem.AttachedDatabase != nil {
+			var hanaClusterIDs []string
+			for _, instance := range sapSystem.AttachedDatabase.GetAllInstances() {
+				hanaClusterIDs = append(hanaClusterIDs, instance.ClusterID)
```
arbulu89, line 52: I guess that it won't affect in anything, but here we will be adding the same cluster id 2 times

arbulu89 (approved):
Thanks @fabriziosestito ,
I thought that we would be just filtering the `clustersService.GetAll` with the HANAScaleUp cluster type. So, once we had the netweaver or scale out healths computed, we could just remove that condition. **You don't need to change the current logic though, it is still good**

PD: For future work, we would need to apply some sort of logic for the SAP systems that don't have any cluster at all (maybe not so common in our dev environments, but some user will have it). In this case, we would be returning a `passing` cluster summary health, which is wrong, and we should return `unknown`. The same would happen if the `AttachedDatabase` is nil, where you did the error handling

It would be as simple as returning `unknown` in `computeAggregatedClustersHealth` if he received variable length is 0.

web/services/health_summary_service.go
fabriziosestito: yes it doesnt matter to the filter so i intentionally ignored the duplication

nelsonkopliku (approved):

web/services/health_summary_service.go
```diff
@@ -34,38 +34,46 @@ func (s *healthSummaryService) GetHealthSummary() (models.HealthSummary, error)
...
-		clusters, err := s.clustersService.GetAll(&ClustersFilter{ID: clusterIDs}, nil)
+		hosts, err := s.hostsService.GetAll(&HostsFilter{ID: hostIDs}, nil)
 		if err != nil {
 			return nil, err
 		}
 
-		hosts, err := s.hostsService.GetAll(&HostsFilter{ID: hostIDs}, nil)
+		clusters, err := s.clustersService.GetAll(&ClustersFilter{
+			ID:          clusterIDs,
+			ClusterType: []string{models.SAPSystemTypeDatabase},
```
arbulu89, line 52: I think this won't work. Here, we need to filter by `models.ClusterTypeHANAScaleUp`

arbulu89 (changes requested):
I think now it is better!
Yet another small change request though XD

arbulu89 (approved):
Here we go!!!!!

PR #824, opened 2022-02-22
Compute attached database health

arbulu89:
Compute the `AttachedDatabase` health. Otherwise, the summary doesn't show the real health, as it uses a non computed value.

### DIFF
--- a/web/services/sap_systems.go
+++ b/web/services/sap_systems.go
@@ -197,7 +197,7 @@ func (s *sapSystemsService) getAllByType(sapSystemType string, tagResourceType s
 	}
 
 	sapSystemList := instances.ToModel()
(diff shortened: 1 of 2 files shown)


web/services/sap_systems.go
```diff
@@ -197,7 +197,7 @@ func (s *sapSystemsService) getAllByType(sapSystemType string, tagResourceType s
 	}
 
 	sapSystemList := instances.ToModel()
-	err = s.ernichSAPSystemList(sapSystemList)
+	err = s.enrichSAPSystemList(sapSystemList)
```
nelsonkopliku, line 200: `ernich`  :smile: eagle eye :eagle:

nelsonkopliku (approved):

PR #825, opened 2022-02-22
adapting values for AWS infrastructure, not all checks are covered by…

pirat013:
… the vars file and they need to be adapted in addition
* initial vars file for AWS setups


runner/ansible/vars/aws/10-default.yml
```diff
@@ -0,0 +1,27 @@
+---
+
+expected_core:
+  "1.1.1": "30000"
+  "1.1.2": "36000"
+  "1.1.3": "" # unkown value, default may apply
```
arbulu89, line 6: Typo here `unkown` -> `unknown`.
Anyway, what means `default may apply`?
If we leave this blank, what will happen is that the check will use an empty value in the comparison, which will produce a failure in the checks result most probably. I don't think this is what we want, right?
I see 2 options. We put any value (what we have in Azure for example), or at least we remove the `default may apply` comment, which would confuse someone who reads this

runner/ansible/vars/aws/10-default.yml
```diff
@@ -0,0 +1,27 @@
...
+  "1.3.1": ""  # doesn't apply for AWS
+  "1.3.2": ""  # doesn't apply for AWS
+  "1.3.4": ""  # doesn't apply for AWS
+  "1.3.5": ""  # doesn't apply for AWS
+  "1.3.6": ""  # doesn't apply for AWS
+  "2.2.1": "SLES_SAP"
+  "2.2.2": "15.1"
+  "2.2.3": "1.1.18"
+  "2.2.3.exclude": "2.0.3+20200511.2b248d828"
+  "2.2.4": "2.4.5" # unkown value, default may apply
```
arbulu89, line 23: I don't see why these last 3 are unknown. I would expect using the same package versions used in Azure.
By the way, we miss the variable for `2.2.5.exclude`, which would make the check fail

arbulu89 (changes requested):
Hi @pirat013 ,
Thanks some comments added. I think we need to something with them

stefanotorresi:
implemented in https://github.com/trento-project/runner/pull/34

PR #826, opened 2022-02-22
adapting values for GCP infrastructure, not all checks are covered by…

pirat013:
… the vars file and they need to be adapted in addition


runner/ansible/vars/gcp/10-default.yml
```diff
@@ -0,0 +1,27 @@
+---
+
+expected_core:
+  "1.1.1": "20000"
+  "1.1.2": "" # unkown value, default may apply
```
arbulu89, line 5: Typo `unkown` -> `unknown`.
Like in https://github.com/trento-project/trento/pull/825 I would remove the `default may apply` comment or add some real value (no empty)

runner/ansible/vars/gcp/10-default.yml
```diff
@@ -0,0 +1,27 @@
...
+  "1.1.3": "20"
+  "1.1.4": "60"
+  "1.1.5": "10"
+  "1.1.6": "udpu"
+  "1.1.7": "2"
+  "1.1.8": "1"
+  "1.2.1": "true"
+  "1.2.2.fence_gcpstonith": "300" # new variable
```
arbulu89, line 13: Not harmful, but this variable won't do anything by now. I would be for removing it, and put it once we have the test updated (or a new one added most probably), so the new PR will make more sense

runner/ansible/vars/gcp/10-default.yml
```diff
@@ -0,0 +1,27 @@
...
+  "1.3.2": "" # doesn't apply to gcp
+  "1.3.4": "" # doesn't apply to gcp
+  "1.3.5": "" # doesn't apply to gcp
+  "1.3.6": "" # doesn't apply to gcp
+  "2.2.1": "SLES_SAP"
+  "2.2.2": "15.1"
+  "2.2.3": "2.0.1"
+  "2.2.3.exclude": "2.0.3+20200511.2b248d828"
+  "2.2.4": "2.4.5" # unkown value, default may apply
+  "2.2.5": "1.4.0" # unkown value, default may apply
```
arbulu89, line 24: Value `"2.2.5.exclude"` for missing

arbulu89 (changes requested):

pirat013:
@arbulu89  Some checks were not successful
Is there something I can fix?

runner/ansible/vars/gcp/10-default.yml
pirat013: Fixed

runner/ansible/vars/gcp/10-default.yml
pirat013: Done

runner/ansible/vars/gcp/10-default.yml
pirat013: Added

stefanotorresi:
implemented in https://github.com/trento-project/runner/pull/34

PR #829, opened 2022-02-22
Enable Grafana persistence

fabriziosestito:
This PR enables the Grafana persistence value in the helm chart, so that dashboards are persisted after rolling a pod

### DIFF
--- a/hack/helm/trento-dev/Chart.lock
+++ b/hack/helm/trento-dev/Chart.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: trento-server
   repository: file://../../../packaging/helm/trento-server
-  version: 0.3.8
-digest: sha256:296438dbc64a55fea561d037f5d6904a2599c88ff18ca2846bbccb2426bdfc6c
-generated: "2022-02-16T12:08:25.452419738+01:00"
+  version: 0.4.1
+digest: sha256:8db94b864c3e7eb9c0e1cb73762ae7dc1a586c388a3d59e27524e2801e36030e
+generated: "2022-02-22T12:29:53.450241777+01:00"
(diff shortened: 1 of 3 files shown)


hack/helm/trento-dev/Chart.lock
```diff
@@ -1,6 +1,6 @@
 dependencies:
 - name: trento-server
   repository: file://../../../packaging/helm/trento-server
-  version: 0.3.8
-digest: sha256:296438dbc64a55fea561d037f5d6904a2599c88ff18ca2846bbccb2426bdfc6c
-generated: "2022-02-16T12:08:25.452419738+01:00"
+  version: 0.4.0
```
stefanotorresi, line 4: hmm I don't think this is right, since you just bumped the version to 0.4.1. Maybe forgot to update the dev chart?

stefanotorresi (changes requested):

hack/helm/trento-dev/Chart.lock
dottorblaster: Versions tend to misalign, I bumped all to the same version number some time ago but it's totally fine to have different versions for charts :/

nelsonkopliku (approved):

stefanotorresi (approved):

packaging/helm/trento-server/Chart.yaml
```diff
@@ -6,7 +6,7 @@ description: The trento server chart contains all the components necessary to ru
 type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates
-version: 0.4.0
+version: 0.4.1
```
arbulu89, line 9: @fabriziosestito Didn't have the time to do the review, but we always forget to update the top comments here XD, to put the same version. Otherwise, IBS won't update the chart package.

PR #830, opened 2022-02-22
Bump helm chart build tag version

fabriziosestito:
^


arbulu89 (approved):
Hoping that merging with a broken CI doesn't break anything more hehe

PR #849, opened 2022-03-02
Exclude diagnostics service sap systems

arbulu89:
Implement the SAP diagnostics agent (DAA usually) detection. When this instance is discovered, the projection is skipped, as we don't want to handle this instance type by now. We will see what to do in the future.

The diagnostics agent instance id is computed using the `/etc/machine-id` as this sap instance is host specific, it is not a distributed deployment. In this case, the agent id and system id will match. Let me know if you have some better suggestion.

The instance type detection has been changed to use the `Features` field with regexp, as this field gives exactly the type (some types are part of the application layer).

Besides that, I have added some test.

FYI: @abravosuse

PD: I have included a small change in the templating, to unify the SAP instances table headers usage. It is here: https://github.com/trento-project/trento/pull/849/commits/8c3b92253c37ac7675dcf82848a89cf06ab5b770
I could remove this commit in any case, but well, it doesn't almost affect. I found it ugly to see when 2 sap systems are discovered, but we won't have this situation often as we won't show diagnostics agents as SAP systems anymore

### DIFF
--- a/agent/discovery/mocks/discovered_sap_system_mock.go
+++ b/agent/discovery/mocks/discovered_sap_system_mock.go
@@ -37,3 +37,18 @@ func NewDiscoveredSAPSystemApplicationMock() sapsystem.SAPSystemsList {
 
 	return s
 }
+
+func NewDiscoveredSAPSystemDiagnosticsMock() sapsystem.SAPSystemsList {
+	var s sapsystem.SAPSystemsList
+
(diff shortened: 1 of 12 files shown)


nelsonkopliku (approved):
:+1:

dottorblaster (approved):
LGTM

PR #854, opened 2022-03-07
Associate attached database properly when the database name is resolved

arbulu89:
If the database name is not an IP address directly, and it must be resolved, this cannot be done in the server side, as the resolution most probably is different.
In order to fix it, the Database IP address must be discovered and sent from the agent.

This PR includes this auto-detection and adds a new field, `DBAddress` in the payload.
Later on, the data projection is slightly changed to include this new field and use to associate App and database.


nelsonkopliku (approved):
LGTM

dottorblaster (approved):
LGTM

PR #855, opened 2022-03-08
Add persistence to postgresql deployment on helm

arbulu89:
Add persistence to postgresql database, as by now, all the data is lost in machine reboot.

The implementation adds a permanent volume and permanent volume claim, together with some small changes in the values usage.

Unfortunately, I have not found the way to use the globals mountPath in the postgresql values (the value is duplicated) as the bitnami chart doesn't run template rendering

### DIFF
--- a/packaging/helm/trento-server/Chart.yaml
+++ b/packaging/helm/trento-server/Chart.yaml
@@ -1,12 +1,12 @@
-#!BuildTag: trento/trento-server:0.4.3
-#!BuildTag: trento/trento-server:0.4.3-build%RELEASE%
+#!BuildTag: trento/trento-server:0.4.4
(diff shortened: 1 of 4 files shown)


fabriziosestito:
This is already enabled by default by the bitnami helm chart, it actually works on my machine :D

image

arbulu89:
> This is already enabled by default by the bitnami helm chart, it actually works on my machine :D
> 
> image

We already commented in other thread, but I think this pvc don't create permanent mounts, so when the machine hosting k3s is rebooted, the data is lost.
Here some links on what bitnami says to enable persistence layer:
https://github.com/bitnami/charts/tree/master/bitnami/postgresql/#persistence

arbulu89:
Closing. The issue looks something else when custom db image is used

PR #858, opened 2022-03-10
Add /usr/sbin to the PATH for the execution

arbulu89:
Most of `crm` based tools are installed in `/usr/sbin` and it looks like not everyone has this directory in their PATH (or sudoers specification).
This change simply adds this directory to the execution PATH.

To reproduce the issue, change the next line in one of the hosts `/etc/sudoers` file:
```
Defaults secure_path="/usr/sbin:/usr/bin:/sbin:/bin"
```
And remove the `/usb/sbin` from there

### DIFF
--- a/runner/ansible/check.yml
+++ b/runner/ansible/check.yml
@@ -28,10 +28,14 @@
       delegate_to: localhost
 
     # Do not change the name. It is use in the trento callback call
-    - name: run_checks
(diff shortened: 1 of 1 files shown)


dottorblaster (approved):
LGTM

abravosuse (approved):

PR #864, opened 2022-03-14
Add helm chart license

rtorrero:
Added the license on the Chart.yaml required by submission guidelines.

BTW: do we need to explicitly state the license in the sub charts?


arbulu89:
@rtorrero You need to upgrade the version of this chart (and the entries in the top of the file, in the #Build, entries)

PR #867, opened 2022-03-15
Allow adjustable polling rate in install-agent.sh

rtorrero:
This PR:
 - Adds a parameter in install-agent.sh to set the discovery interval for the agents
 - Makes use of that parameter from the `ci-cd.yaml` to set it to a new value more appropriate for our demo environment.
 
Notes:
 - The environment secret `TRENTO_DISCOVERY_INTERVAL` has already been added with a value of `90`.
 - I didn't want to adjust a new default value in the code as it might impact customers / partners that are already testing and would like to keep the 10s default.

### DIFF
--- a/.github/workflows/ci-cd.yaml
+++ b/.github/workflows/ci-cd.yaml
@@ -418,7 +418,7 @@ jobs:
       TRENTO_USER: ${{ secrets.TRENTO_USER }}
       TRENTO_SERVER_HOST: ${{ secrets.TRENTO_SERVER_HOST }}
       TRENTO_REPO_OWNER: ${{ github.repository_owner }}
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-cd.yaml
```diff
@@ -418,6 +418,7 @@ jobs:
       TRENTO_USER: ${{ secrets.TRENTO_USER }}
       TRENTO_SERVER_HOST: ${{ secrets.TRENTO_SERVER_HOST }}
       TRENTO_REPO_OWNER: ${{ github.repository_owner }}
+      INTERVAL: ${{ secrets.TRENTO_DISCOVERY_INTERVAL }}
```
arbulu89, line 421: Could we rename this to `TRENTO_DISCOVERY_INTERVAL`?

PR #876, opened 2022-03-25
Split subscription discovery to a 900 sec interval

rtorrero:
This PR's main goal is to reduce the possible load we are hitting our SCC service with. It attempts this by:
 - Splitting the subscription discovery from the rest of the discoveries so that we can set a separate interval
 - Increasing the default interval time to a tick per each 15 minutes.
 
In addition to this, we have added a hidden parameter in cobra to allow quickly adjusting this time interval. Feedback if this hidden param is a good idea or not is appreciated!

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -18,18 +18,20 @@ import (
 const trentoAgentCheckId = "trentoAgent"
 
 type Agent struct {
-	config          *Config
-	collectorClient collector.Client
-	discoveries     []discovery.Discovery
-	ctx             context.Context
-	ctxCancel       context.CancelFunc
+	config                *Config
+	collectorClient       collector.Client
+	discoveries           []discovery.Discovery
+	subscriptionDiscovery discovery.Discovery
+	ctx                   context.Context
+	ctxCancel             context.CancelFunc
 }
 
 type Config struct {
-	InstanceName    string
-	SSHAddress      string
-	DiscoveryPeriod time.Duration
-	CollectorConfig *collector.Config
+	InstanceName                string
+	SSHAddress                  string
+	DiscoveryPeriod             time.Duration
+	SubscriptionDiscoveryPeriod time.Duration
+	CollectorConfig             *collector.Config
 }
 
 // NewAgent returns a new instance of Agent with the given configuration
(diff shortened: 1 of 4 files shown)


abravosuse (commented):
I see the 900s (15m) discovery period for subscription info. For me it is good. Maybe update the PR title to reflect this number?

rtorrero:
> I see the 900s (15m) discovery period for subscription info. For me it is good. Maybe update the PR title to reflect this number?

woops! you are right @abravosuse, sorry!

arbulu89 (commented):
Maybe I would drive for something more generic, like having the option to specify the discovery time for each individual discovery. And we would spawn a new go routine for each of them, with their interval. Maybe we wouldn't need to supply the option to change all of them to the user, but at least we would have already this flexibility. We could still use the current `discovery-period` value to populate all the discovery times by now, or even create flags for all of them.

I don't know. Having the sles subscription out there looks pretty ugly, at least for my eyes hehe

Finally, I hope nobody uses the new flag to put something like 10 seconds... because we would still be in trouble again... XD

If everyone agrees with the current solution, I'm fine. It should do what we want at the end

dottorblaster:
Same comment as @arbulu89, but for the current iteration is fine for me. @fabriziosestito @nelsonkopliku what do you think?

fabriziosestito:
> Same comment as @arbulu89, but for the current iteration is fine for me. @fabriziosestito @nelsonkopliku what do you think?

I also agree it in making it configurable for all the loops. What I don't really dig about the proposed approach is that we have a different (duplicated) loop for the subscription. I would rather have a `start ticker` function where we can pass the discovery time, so we can dry up the code as well.

rtorrero:
> Following up on @arbulu89 idea, maybe we could set minimum values for certain individual discoveries? For instance, for the subscription info discovery..

That sounds like a good idea for me; add some sanity checks preventing values that don't make much sense.

> > Same comment as @arbulu89, but for the current iteration is fine for me. @fabriziosestito @nelsonkopliku what do you think?
> 
> I also agree it in making it configurable for all the loops. What I don't really dig about the proposed approach is that we have a different (duplicated) loop for the subscription. I would rather have a `start ticker` function where we can pass the discovery time, so we can dry up the code as well.

This sounds good to me too, should we merge this meanwhile or go directly with the better approach?

PR #881, opened 2022-03-28
Split discovery intervals

rtorrero:
This PR superseedes #876 and in addition to just splitting the subscription discovery, it splits all the discoveries into separately configurable variables:
```
ClusterDiscoveryPeriod      time.Duration
SAPSystemDiscoveryPeriod    time.Duration
CloudDiscoveryPeriod        time.Duration
HostDiscoveryPeriod         time.Duration
SubscriptionDiscoveryPeriod time.Duration
```
Some notes:
 - The subscription discovery interval is kept as a hidden parameter
 - All others are visible in the `trento` CLI when using `--help`
 - We add the interval as part of the `BaseDiscovery` struct in the `interval` field and access them through the `GetInterval()` method
 
 @abravosuse min/max limits on the intervals have not been added as have a few questions:
  - Do we add the limits to all discoveries or just the subscription one?
  - What do you think about the default values? (10 for all execept subscription which is 900)

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -24,41 +24,33 @@ type Agent struct {
 	ctxCancel       context.CancelFunc
 }
 
-type DiscoveryPeriodConfig struct {
-	Cluster      time.Duration
-	SAPSystem    time.Duration
-	Cloud        time.Duration
-	Host         time.Duration
-	Subscription time.Duration
-}
-
 type Config struct {
-	InstanceName           string
-	SSHAddress             string
-	DiscoveryPeriodsConfig *DiscoveryPeriodConfig
-	CollectorConfig        *collector.Config
+	InstanceName      string
+	DiscoveriesConfig *discovery.DiscoveriesConfig
 }
 
 // NewAgent returns a new instance of Agent with the given configuration
 func NewAgent(config *Config) (*Agent, error) {
-	collectorClient, err := collector.NewCollectorClient(config.CollectorConfig)
+	collectorClient, err := collector.NewCollectorClient(config.DiscoveriesConfig.CollectorConfig)
 	if err != nil {
 		return nil, errors.Wrap(err, "could not create a collector client")
 	}
 
 	ctx, ctxCancel := context.WithCancel(context.Background())
+
+	var discoveries discovery.DiscoveryList
+	discoveries, err = discoveries.AddDiscovery(discovery.NewClusterDiscovery, collectorClient, *config.DiscoveriesConfig)
+	discoveries, err = discoveries.AddDiscovery(discovery.NewSAPSystemsDiscovery, collectorClient, *config.DiscoveriesConfig)
+	discoveries, err = discoveries.AddDiscovery(discovery.NewCloudDiscovery, collectorClient, *config.DiscoveriesConfig)
+	discoveries, err = discoveries.AddDiscovery(discovery.NewSubscriptionDiscovery, collectorClient, *config.DiscoveriesConfig)
+	discoveries, err = discoveries.AddDiscovery(discovery.NewHostDiscovery, collectorClient, *config.DiscoveriesConfig)
+
 	agent := &Agent{
 		config:          config,
 		collectorClient: collectorClient,
 		ctx:             ctx,
 		ctxCancel:       ctxCancel,
-		discoveries: []discovery.Discovery{
-			discovery.NewClusterDiscovery(collectorClient, config.DiscoveryPeriodsConfig.Cluster),
-			discovery.NewSAPSystemsDiscovery(collectorClient, config.DiscoveryPeriodsConfig.SAPSystem),
-			discovery.NewCloudDiscovery(collectorClient, config.DiscoveryPeriodsConfig.Cloud),
-			discovery.NewHostDiscovery(config.SSHAddress, collectorClient, config.DiscoveryPeriodsConfig.Host),
-			discovery.NewSubscriptionDiscovery(collectorClient, config.DiscoveryPeriodsConfig.Subscription),
-		},
+		discoveries:     discoveries,
 	}
 	return agent, nil
 }
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -46,10 +46,7 @@ func NewAgent(config *Config) (*Agent, error) {
 	var discoveries discovery.DiscoveryList
 
 	for _, initializer := range discoveryInitializers {
-		discoveries, err = discoveries.AddDiscovery(initializer, collectorClient, *config.DiscoveriesConfig)
-		if err != nil {
-			return nil, errors.Wrap(err, "could not create a discovery")
-		}
+		discoveries = discoveries.AddDiscovery(initializer, collectorClient, *config.DiscoveriesConfig)
 	}
 
 	ctx, ctxCancel := context.WithCancel(context.Background())
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -36,14 +36,23 @@ func NewAgent(config *Config) (*Agent, error) {
 		return nil, errors.Wrap(err, "could not create a collector client")
 	}
 
-	ctx, ctxCancel := context.WithCancel(context.Background())
-
+	discoveryInitializers := []discovery.DiscoveryInitializer{
+		discovery.NewClusterDiscovery,
+		discovery.NewSAPSystemsDiscovery,
+		discovery.NewCloudDiscovery,
+		discovery.NewSubscriptionDiscovery,
+		discovery.NewHostDiscovery,
+	}
 	var discoveries discovery.DiscoveryList
-	discoveries, err = discoveries.AddDiscovery(discovery.NewClusterDiscovery, collectorClient, *config.DiscoveriesConfig)
-	discoveries, err = discoveries.AddDiscovery(discovery.NewSAPSystemsDiscovery, collectorClient, *config.DiscoveriesConfig)
-	discoveries, err = discoveries.AddDiscovery(discovery.NewCloudDiscovery, collectorClient, *config.DiscoveriesConfig)
-	discoveries, err = discoveries.AddDiscovery(discovery.NewSubscriptionDiscovery, collectorClient, *config.DiscoveriesConfig)
-	discoveries, err = discoveries.AddDiscovery(discovery.NewHostDiscovery, collectorClient, *config.DiscoveriesConfig)
+
+	for _, initializer := range discoveryInitializers {
+		discoveries, err = discoveries.AddDiscovery(initializer, collectorClient, *config.DiscoveriesConfig)
+		if err != nil {
+			return nil, errors.Wrap(err, "could not create a discovery")
+		}
+	}
+
+	ctx, ctxCancel := context.WithCancel(context.Background())
 
 	agent := &Agent{
 		config:          config,
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -24,15 +24,19 @@ type Agent struct {
 	ctxCancel       context.CancelFunc
 }
 
+type DiscoveryPeriodConfig struct {
+	Cluster      time.Duration
+	SAPSystem    time.Duration
+	Cloud        time.Duration
+	Host         time.Duration
+	Subscription time.Duration
+}
+
 type Config struct {
-	InstanceName                string
-	SSHAddress                  string
-	ClusterDiscoveryPeriod      time.Duration
-	SAPSystemDiscoveryPeriod    time.Duration
-	CloudDiscoveryPeriod        time.Duration
-	HostDiscoveryPeriod         time.Duration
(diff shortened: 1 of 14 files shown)


agent/agent.go
```diff
@@ -26,10 +25,14 @@ type Agent struct {
 }
 
 type Config struct {
-	InstanceName    string
-	SSHAddress      string
-	DiscoveryPeriod time.Duration
-	CollectorConfig *collector.Config
+	InstanceName                string
+	SSHAddress                  string
+	ClusterDiscoveryPeriod      time.Duration
```
nelsonkopliku, line 30: I'd encapsulate the specific `ClusterDiscoveryPeriod`, `SAPSystemDiscoveryPeriod` and so forth under a `DiscoveryPeriodConfig` or something similar.

agent/discovery/host.go
```diff
@@ -22,18 +23,22 @@ type HostDiscovery struct {
...
-	d.discovery = NewDiscovery(collectorClient)
+	d.discovery = NewDiscovery(collectorClient, interval)
 	return d
 }
 
 func (h HostDiscovery) GetId() string {
 	return h.id
 }
 
+func (d HostDiscovery) GetInterval() time.Duration {
```
nelsonkopliku, line 38: Is this needed? Can we use the one in `BaseDiscovery`? :thinking:

agent/discovery/sapsystem.go
```diff
@@ -15,17 +16,21 @@ type SAPSystemsDiscovery struct {
...
-	r.discovery = NewDiscovery(collectorClient)
+	r.discovery = NewDiscovery(collectorClient, interval)
 	return r
 }
 
 func (d SAPSystemsDiscovery) GetId() string {
 	return d.id
 }
 
+func (d SAPSystemsDiscovery) GetInterval() time.Duration {
```
nelsonkopliku, line 30: Same as before: Is this needed? Can we use the one in BaseDiscovery? thinking

abravosuse:
@rtorrero sorry for the late reply:

- Do we add the limits to all discoveries or just the subscription one?
I think all discoveries should have, at least, a min limit. No sot sure about the max limit: users should be able to adjust them to whatever value comfortable with. If a user is OK having their cluster info updated every 10m, who are we to say no?

- What do you think about the default values? (10 for all except subscription which is 900)
Good.

agent/discovery/host.go
rtorrero: I don't think it is, as of now there is no reason to not use the one in BaseDiscovery, I have no strong opinion here; it's just a matter of taste as the attribute is really stored in the BaseDiscovery

arbulu89 (commented):
Hi @rtorrero,
The code looks good in overall. I tend to agree with @nelsonkopliku that we could reduce a bit of duplication in the `GetInterval` methods just storing the interval in the primary structs themselves.
Anyway, it will do exactly the same hehe

PD: I put a comment about the variable name used in the `GetInterval` methods, as it doesn't match in all the entries

cmd/agent/agent.go
```diff
@@ -47,7 +52,13 @@ func NewAgentCmd() *cobra.Command {
 
 	startCmd.Flags().StringVar(&sshAddress, "ssh-address", "", "The address to which the trento-agent should be reachable for ssh connection by the runner for check execution.")
 
-	startCmd.Flags().IntVarP(&discoveryPeriod, "discovery-period", "", 10, "Discovery mechanism loop period in seconds")
+	startCmd.Flags().IntVarP(&clusterDiscoveryPeriod, "cluster-discovery-period", "", 10, "Cluster discovery mechanism loop period in seconds")
```
nelsonkopliku, line 55: What hsppens if we get a non integer value in one of these options? For instance the string `whatever`.
And what happens if we get a `0` or `-1`?

:sweat_smile:

nelsonkopliku: In the first case viper/cobra should scream I hope. For the `0` or `-1` cases not sure :thinking:

agent/discovery/host.go
rtorrero: @arbulu89 not sure, it feels a bit wrong to have exactly the same attribute (`interval`) repeated in the `HostDiscovery`, `ClusterDiscovery`,.... but I see your point about the `BaseDiscovery` not being that useful. I'd address that in a separate PR (I'll create an issue if you agree with this) and just move the `GetInterval()` function there for now. WDYT?

cmd/agent/agent.go
rtorrero: For -1: 
```
panic: non-positive interval for NewTicker

goroutine 6 [running]:
time.NewTicker(0xc0000bbca0)
	time/tick.go:24 +0x10f
github.com/trento-project/trento/internal.Repeat({0xef6be0, 0x14}, 0xc0004a5f00, 0xc0001880e0, {0x1a9b518, 0xc0000b8380})
	github.com/trento-project/trento/internal/utils.go:106 +0x7a
github.com/trento-project/trento/agent.(*Agent).startDiscoverTicker(0xc0001c2190, {0x1a92d90, 0xc0001c2230})
	github.com/trento-project/trento/agent/agent.go:108 +0x93
github.com/trento-project/trento/agent.(*Agent).Start.func1(0xc0000343d0, {0x1a92d90, 0xc0001c2230})
	github.com/trento-project/trento/agent/agent.go:75 +0xf1
created by github.com/trento-project/trento/agent.(*Agent).Start
	github.com/trento-project/trento/agent/agent.go:72 +0x209

```

We could add some validation so it doesn't reach the `NewTicker` call and we catch it in the cobra argument parsing. We need it anyway to implement the min values that @abravosuse suggested.

nelsonkopliku: Ok then, have a lot of fun :smile:

arbulu89 (commented):
@rtorrero Other extra comment. Now that we have different flags, we need to update the `agent.yml` configuration file and maybe the agent install script (I'm not sure about this last one)

cmd/agent/agent.go
```diff
@@ -47,7 +52,13 @@ func NewAgentCmd() *cobra.Command {
...
+	startCmd.Flags().IntVarP(&sapSystemDiscoveryPeriod, "sapsystem-discovery-period", "", 10, "SAP systems discovery mechanism loop period in seconds")
+	startCmd.Flags().IntVarP(&cloudDiscoveryPeriod, "cloud-discovery-period", "", 10, "Cloud discovery mechanism loop period in seconds")
+	startCmd.Flags().IntVarP(&hostDiscoveryPeriod, "host-discovery-period", "", 10, "Host discovery mechanism loop period in seconds")
+	startCmd.Flags().IntVarP(&subscriptionDiscoveryPeriod, "subscription-discovery-period", "", 900, "Subscription discovery mechanism loop period in seconds")
+
+	startCmd.Flags().MarkHidden("subscription-discovery-period")
```
arbulu89, line 61: Why is this flag hidden if we have validation later?

agent/discovery/host.go
rtorrero: I ended up improving this a bit. Now it makes a bit more sense as we don't have the BaseDiscovery struct anymore but the intervals are stored in each Discovery that implements the Discovery interface, but of course the GetInterval() is still there. I think it has sense to be there so that at some point we can e.g. query the effective interval values to display them somewhere.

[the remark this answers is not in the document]
rtorrero: I'd use `d` in all discoveries as we have e.g. `ClusterDiscovery` and `CloudDiscovery` which start both with `c`, wdyt?

agent/agent.go
```diff
@@ -26,32 +25,32 @@ type Agent struct {
...
-	collectorClient, err := collector.NewCollectorClient(config.CollectorConfig)
+	collectorClient, err := collector.NewCollectorClient(config.DiscoveriesConfig.CollectorConfig)
 	if err != nil {
 		return nil, errors.Wrap(err, "could not create a collector client")
 	}
 
 	ctx, ctxCancel := context.WithCancel(context.Background())
+
+	var discoveries discovery.DiscoveryList
+	discoveries, err = discoveries.AddDiscovery(discovery.NewClusterDiscovery, collectorClient, *config.DiscoveriesConfig)
```
arbulu89, line 42: We don't do anything here with the errors, right? So the validation is not really used for anything

cmd/agent/config.go
```diff
@@ -44,17 +45,31 @@ func LoadConfig() (*agent.Config, error) {
...
+		CollectorHost: viper.GetString("collector-host"),
+		CollectorPort: viper.GetInt("collector-port"),
+		EnablemTLS:    enablemTLS,
+		Cert:          cert,
+		Key:           key,
+		CA:            ca,
+	}
+
+	discoveryPeriodsConfig := &discovery.DiscoveriesPeriodConfig{
+		Cluster:      time.Duration(viper.GetInt("cluster-discovery-period")) * time.Second,
```
arbulu89, line 58: Here maybe we could use something like `GetDuration`: https://github.com/spf13/pflag/blob/master/duration.go#L47
More examples: https://golang.hotexamples.com/examples/github.com.ogier.pflag/-/DurationVar/golang-durationvar-function-examples.html
Maybe with this, we can get rid of further validation. I didn't check, but doing the validation in the way is done now looks a bit ugly hehe

arbulu89 (commented):
Hey @rtorrero ,

I'm approving the PR, even though there are thing that I don't personally like, but reckon that the PR is talking too long, so we can get the fix and revisit again if needed.

- I think there must be the way to the flags validation in cobra/viper itself. It is true that it wouldn't be good if the configuration was created somewhere else, but I don't think this will happen
- I think the old configuration was "better". I would simply change there the `DiscoveryPeriod` by the new `DiscoveriesPeriodConfig`
- I don't really like the new `AddDiscovery` thingy. It looks cumbersome. The old one looked better. I understand that you added it to make the validation though (anyway, the errors are not used now)

So, my personal suggestions are (**you don't really need to implement them**)
- Try to do the validation in the cmd part
- Return back to the original `Config` replacing the individual period by the periods new config struct
- Put the discoveries as before (they wouldn't return errors anymore), and pass the internval in the initialization, getting it from the config)
- The `host` attribute is not used in the discoveries except in the `HostDiscovery` I think. I would remove it in the other cases
- I don't really like the new `AddDiscovery` thingy. I find it pretty cumbersome. The old

arbulu89 (approved):

abravosuse:
> The changes that we discussed should be there. Regarding the default and minimal values: I've set all defaults to 10 seconds, except the discovery which is set at 900s. The minimal values are 1 for all except for the discovery where we have a min allowed value of 500s.
> 
> @abravosuse if you want to change these let me know

I'm fine with these values @rtorrero.

cmd/agent/agent.go
rtorrero: Yes, giving it another thought, the fact that it's allows us to put a lower min value in the validation

test/fixtures/config/agent.yaml
```diff
@@ -1,5 +1,8 @@
 ssh-address: some-ssh-address
-discovery-period: 10
```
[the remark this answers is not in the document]
arbulu89: Ok, yes, the `s` is needed, so maybe we would need to change the `agent.yaml` file, and I don't know maybe the `agent-install` script (as we create the configuration file there)

agent/discovery/subscription.go
```diff
@@ -2,37 +2,49 @@ package discovery
...
+	"os"
+	"time"
 
 	log "github.com/sirupsen/logrus"
 	"github.com/trento-project/trento/agent/discovery/collector"
 	"github.com/trento-project/trento/internal/subscription"
 )
 
 const SubscriptionDiscoveryId string = "subscription_discovery"
+const SubscriptionDiscoveryMinPeriod time.Duration = 20 * time.Second
```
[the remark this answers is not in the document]
rtorrero: 900 is the default, this is the minimal value. It was initially set to 500s but after you wisely said; if the flag is hidden then it doesn't make much sense to have a such a hard limit.

agent/agent.go
```diff
@@ -26,32 +25,38 @@ type Agent struct {
...
+		discovery.NewClusterDiscovery,
+		discovery.NewSAPSystemsDiscovery,
+		discovery.NewCloudDiscovery,
+		discovery.NewSubscriptionDiscovery,
+		discovery.NewHostDiscovery,
+	}
+	var discoveries discovery.DiscoveryList
+
+	for _, initializer := range discoveryInitializers {
+		discoveries = discoveries.AddDiscovery(initializer, collectorClient, *config.DiscoveriesConfig)
```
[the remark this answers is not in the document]
rtorrero: I think you are confusing `CollectorConfig` with `CollectorClient`, right?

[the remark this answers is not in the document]
arbulu89: Sorry, yes, you are right

PR #900, opened 2022-04-05
Add missing units in the agent installer config template

rtorrero:
This is a small leftover from https://github.com/trento-project/trento/pull/881 as the installer template misses the now required units (otherwise it defaults to nanoseconds which do not pass the validation)


arbulu89 (changes requested):
We miss them in the configuration examples too:
https://github.com/trento-project/trento/blob/main/packaging/config/agent.yaml

arbulu89 (approved):

PR #911, opened 2022-04-08
check 1.1.9 (DA114A): to scan node-directives for both rings

fmherschel:
Instead of counting the string "interface"  the test needs to scan the corosync.conf file for the directices node { ... }. There must be exact 2 such node-directives for the current se case and each of the entries needs exact two ring-addresses to be configured. 
The perl-snipset outputs a "2" per found node-directive (if exactly 2 rings are found). So the correct result following the best practice is "22" (both name-directives have exact two ring-address definitions).

### DIFF
--- a/runner/ansible/roles/checks/1.1.9/tasks/main.yml
+++ b/runner/ansible/roles/checks/1.1.9/tasks/main.yml
@@ -2,8 +2,31 @@
 
 - name: "{{ name }}.check"
   shell: |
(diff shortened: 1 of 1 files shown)


stefanotorresi:
note to self: needs forward-port into [runner](https://github.com/trento-project/runner)

pirat013:
Looks good for me. I vote to merge this PR.

stefanotorresi:
implemented in https://github.com/trento-project/runner/pull/38

